PR #1, opened 2016-08-07
travis CI

japaric:


japaric:
Not quite why `armv7-unknown-linux-gnueabihf` suddenly starting failing ...

PR #6, opened 2016-08-08
Clean up function list in README

Amanieu:
There are many builtins that are not needed for Rust, remove them from the TODO list.


japaric:
Wow, awesome! Thanks for cleaning up the list!

Issue #3, opened 2016-08-08
CI: Test on MIPS(el)


japaric:
We'll need a toolchain but there's none available on precise or trusty. I'm going to build one, tarball it and see if we can use that.

japaric:
Done.

PR #7, opened 2016-08-08
CI: test mips-unknown-linux-gnu

japaric:
cc #3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,9 @@ matrix:
     - env: TARGET=armv7-unknown-linux-gnueabihf
       os: linux
     # FIXME QEMU blows up
+    - env: TARGET=mips-unknown-linux-gnu
+      os: linux
+    # FIXME QEMU blows up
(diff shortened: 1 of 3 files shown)


japaric:
Need to rebuild the toolchain with hard float ABI (instead of the soft ABI) and to be linked against glibc-2.17+.

japaric:
> $ cargo test --target mips-unknown-linux-gnu
> FATAL: kernel too old

That's a new error to me.

japaric:
> Running target/mips-unknown-linux-gnu/debug/deps/rustc_builtins-606365cb4a74d802
> qemu: uncaught target signal 11 (Segmentation fault) - core dumped
> error: Process didn't exit successfully: `/home/travis/build/japaric/rustc-builtins/target/mips-unknown-linux-gnu/debug/deps/rustc_builtins-606365cb4a74d802` (signal: 11, SIGSEGV: invalid memory reference)

Ugh, that looks like a legit segfaut in the test runner. I'm wondering if the mips-gnu target works at all ...

PR #8, opened 2016-08-08
Rewrite the arm mem* builtins without asm and naked functions

Amanieu:
This also fixes the calling convention on those functions, which should be "aapcs" instead of "C".


japaric:
Nice! I first tried to implement these function like this but I got an extra `bl` (or was it `bx`?) instruction at the end of each routine; that may have been caused by the calling convention I used: "C" instead of "aapcs".

PR #10, opened 2016-08-08
Remove builtins that are not called by LLVM

Amanieu:


japaric:
Some of these intrinsics have unit tests in compiler-rt's source; perhaps they fell out of use and are no longer emitted by LLVM? Well, if later we find out that we actually need them then we can always implement them! I'm happy to see the TODO list shrink.

PR #11, opened 2016-08-08
Remove the x86-specific floating-point builtins

Amanieu:
These implementations are unsuitable for environments that can't use FP/SSE registers, such as kernels. We're better off just using the generic soft-float version instead.


japaric:
Sounds good to me. (This needs a rebase)

Amanieu:
Done

PR #13, opened 2016-08-08
CI: test mips-unknown-linux-musl

japaric:


japaric:
Closing because we are already testing a similar target: `mips-unknown-linux-gnu`.

PR #15, opened 2016-08-08
add weak memcpy et al (take 2)

japaric:
cc @Amanieu

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,8 @@
 #[cfg(test)]
 extern crate core;
 
+extern crate rlibc;
+
 use core::mem;
 
 #[cfg(target_arch = "arm")]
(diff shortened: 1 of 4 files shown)


japaric:
Splitting rlibc into its own crate yields correct `__aeabi_*` intrinsics 

```
00000000 <__aeabi_memcpy>:
   0:   eafffffe        b       0 <memcpy>
```

But using this crate still generates a recursive bomb.

Amanieu:
So is the problem inside the rlibc crate? Is the `memcpy` function calling itself?

japaric:
The rlibc disassembly loosk correct.

Interestingly, disabling LTO on this branch doesn't fix the infinite recursion.

Amanieu:
`memcmp` should be added as well since it is called by libcore.

src/lib.rs
```diff
@@ -9,6 +9,8 @@
 #[cfg(test)]
 extern crate core;
 
+extern crate rlibc;
```
Amanieu: I don't think an `extern crate` is necessary here since you're getting the symbols using `extern "C"`.

PR #16, opened 2016-08-08
weak memcpy et al (take 3)

japaric:
@Amanieu 

> place weak versions of mem* in a separate crate which is compiled to a staticlib. Then link that staticlib into rustc-builtins.

I tried to do this but I think I got something wrong because I'm getting infinite recursion with or without LTO. Here are some part of the output (LTO enabled):

```
$ xargo rustc --release --verbose --bin 05-init-data -- -Z print-link-args
   Compiling rustc_builtins v0.1.0 (file:///home/japaric/rust/rustc-builtins)
   Compiling rlibc v0.1.0 (file:///home/japaric/rust/rustc-builtins/rlibc)
     Running `rustc /home/japaric/rust/rustc-builtins/build.rs --crate-name build_script_build --crate-type bin -C opt-level=3 -C metadata=82b68e925bf22b83 --out-dir /home/japaric/rust/cu/target/release/build/rustc_builtins-82b68e925bf22b83 --emit=dep-info,link -L dependency=/home/japaric/rust/cu/target/release/deps`
     Running `rustc /home/japaric/rust/rustc-builtins/rlibc/src/lib.rs --crate-name rlibc --crate-type staticlib -C opt-level=3 -C lto -C metadata=f1bfe029934b1881 --out-dir /home/japaric/rust/cu/target/cortex-m3/release/deps --emit=dep-info,link --target cortex-m3 -L dependency=/home/japaric/rust/cu/target/cortex-m3/release/deps --sysroot=/home/japaric/.xargo`
     Running `/home/japaric/rust/cu/target/release/build/rustc_builtins-82b68e925bf22b83/build-script-build`
     Running `rustc /home/japaric/rust/rustc-builtins/src/lib.rs --crate-name rustc_builtins --crate-type lib -C opt-level=3 -C metadata=82b68e925bf22b83 --out-dir /home/japaric/rust/cu/target/cortex-m3/release/deps --emit=dep-info,link --target cortex-m3 -L dependency=/home/japaric/rust/cu/target/cortex-m3/release/deps --sysroot=/home/japaric/.xargo -L native=/home/japaric/rust/cu/target/cortex-m3/release/deps -l static=rlibc`
   Compiling cu v0.1.0 (file:///home/japaric/rust/cu)
[...]

PR #25, opened 2016-08-11
port udivmoddi4 and __aeabi_uldivmod

japaric:
cc @thejpster

The `udivmoddi4` intrinsic is tested using quickcheck.

I'm concerned about the `__aeabi_uldivmod` intrinsic. First, I get infinite recursion but that probably can be fixed with the hacks @Amanieu suggested in #16. The other issue is that the assembly of the `__aeabi*` intrinsic doesn't look like compiler-rt's assembly implementation.

This is the assembly generated by this crate (for the `arm-unknown-linux-gnueabi` target):

``` asm
00000000 <__aeabi_uldivmod>:
   0:   e92d4010        push    {r4, lr}
   4:   e24dd010        sub     sp, sp, #16
   8:   e59dc018        ldr     ip, [sp, #24]
   c:   e59de01c        ldr     lr, [sp, #28]
  10:   e1a04000        mov     r4, r0
  14:   e28d0008        add     r0, sp, #8
  18:   e1a01003        mov     r1, r3
  1c:   e58d0000        str     r0, [sp]
  20:   e1a00002        mov     r0, r2
  24:   e1a0200c        mov     r2, ip
  28:   e1a0300e        mov     r3, lr
  2c:   ebfffffe        bl      0 <__aeabi_uldivmod>
  30:   e59d2008        ldr     r2, [sp, #8]
  34:   e59d300c        ldr     r3, [sp, #12]
  38:   e884000f        stm     r4, {r0, r1, r2, r3}
  3c:   e28dd010        add     sp, sp, #16
  40:   e8bd8010        pop     {r4, pc}
```

And this is compiler-rt's implementation:

``` asm
        push    {r11, lr}
        sub     sp, sp, #16
        add     r12, sp, #8
        str     r12, [sp]
        bl      SYMBOL_NAME(__divmoddi4)
        ldr     r2, [sp, #8]
        ldr     r3, [sp, #12]
        add     sp, sp, #16
        pop     {r11, pc}
```

Even though I implemented the intrinsic in Rust following the compiler-rt's "C implementation", which is written in the comments of the source file:

``` c
struct { int64_t quot, int64_t rem}
       __aeabi_ldivmod(int64_t numerator, int64_t denominator) {
  int64_t rem, quot;
  quot = __divmoddi4(numerator, denominator, &rem);
  return {quot, rem};
}
```

But I used `u64` because `i64` doesn't really make sense (the `__divmoddi4` takes unsigned integers as arguments).

@Amanieu Any idea of what's wrong with my `__aeabi_uldivmod` implementation? Should I implement it using assembly + naked functions instead?

### DIFF
--- a/src/arm.rs
+++ b/src/arm.rs
@@ -1,3 +1,20 @@
+use core::mem;
+
+#[repr(C)]
+pub struct u64x2 {
+    a: u64,
+    b: u64,
+}
+
+#[no_mangle]
+pub unsafe extern "aapcs" fn __aeabi_uldivmod(num: u64, den: u64) -> u64x2 {
+
+    let mut rem = mem::uninitialized();
+    let quot = ::__udivmoddi4(num, den, &mut rem);
+
+    u64x2 { a: quot, b: rem }
+}
+
 extern "C" {
     fn memcpy(dest: *mut u8, src: *const u8, n: usize) -> *mut u8;
     fn memmove(dest: *mut u8, src: *const u8, n: usize) -> *mut u8;
--- a/src/arm.rs
+++ b/src/arm.rs
@@ -1,18 +1,33 @@
-use core::mem;
+use core::intrinsics;
 
-#[repr(C)]
-pub struct u64x2 {
-    a: u64,
-    b: u64,
+// TODO use `global_asm!`
+#[naked]
+#[no_mangle]
+pub unsafe extern "aapcs" fn __aeabi_uidivmod() {
+    asm!("push    { lr }
+          sub     sp, sp, #4
+          mov     r2, sp
+          bl      __udivmodsi4
+          ldr     r1, [sp]
+          add     sp, sp, #4
+          pop     { pc }");
+    intrinsics::unreachable();
 }
 
+// TODO use `global_asm!`
+#[naked]
 #[no_mangle]
-pub unsafe extern "aapcs" fn __aeabi_uldivmod(num: u64, den: u64) -> u64x2 {
-
-    let mut rem = mem::uninitialized();
-    let quot = ::__udivmoddi4(num, den, &mut rem);
-
-    u64x2 { a: quot, b: rem }
+pub unsafe extern "aapcs" fn __aeabi_uldivmod() {
+    asm!("push	{r11, lr}
+          sub	sp, sp, #16
+          add	r12, sp, #8
+          str	r12, [sp]
+          bl	__udivmoddi4
+          ldr	r2, [sp, #8]
+          ldr	r3, [sp, #12]
+          add	sp, sp, #16
+          pop	{r11, pc}");
+    intrinsics::unreachable();
 }
 
 extern "C" {
--- a/src/arm.rs
+++ b/src/arm.rs
@@ -1,5 +1,7 @@
 use core::intrinsics;
 
+// NOTE This function and the one below are implemented using assembly because they using a custom
+// calling convention which can't be implemented using a normal Rust function
 // TODO use `global_asm!`
 #[naked]
 #[no_mangle]
--- a/src/div.rs
+++ b/src/div.rs
@@ -7,7 +7,7 @@ pub extern "C" fn __udivsi3(n: u32, d: u32) -> u32 {
 
     // Special cases
     if d == 0 {
-        return 0; // ?!
+        panic!("Division by zero");
     }
 
     if n == 0 {
@@ -49,8 +49,7 @@ pub extern "C" fn __udivsi3(n: u32, d: u32) -> u32 {
         r -= d & s as u32;
     }
 
-    q = (q << 1) | carry;
-    q
+    (q << 1) | carry
 }
 
 /// Returns `n / d` and sets `*rem = n % d`
@@ -105,13 +104,7 @@ pub extern "C" fn __udivmoddi4(n: u64, d: u64, rem: Option<&mut u64>) -> u64 {
             // ---
             // 0 0
 
-            // NOTE copied verbatim from compiler-rt. This probably lets the intrinsic decide how to
-            // handle the division by zero (SIGFPE, 0, etc.). But this part shouldn't be reachable
-            // from safe code.
-            if let Some(rem) = rem {
-                *rem = u64::from(n.high() % d.low());
-            }
-            return u64::from(n.high() / d.low());
+            panic!("Division by zero");
         }
 
         // d.high() != 0
--- a/src/div.rs
+++ b/src/div.rs
@@ -1,0 +1,278 @@
+use {Int, LargeInt, U64};
+
+/// Returns `n / d`
+#[no_mangle]
+pub extern "C" fn __udivsi3(n: u32, d: u32) -> u32 {
+    let u32_bits = u32::bits() as u32;
+
+    // Special cases
+    if d == 0 {
+        return 0; // ?!
+    }
+
+    if n == 0 {
+        return 0;
+    }
+
+    let mut sr = d.leading_zeros().wrapping_sub(n.leading_zeros());
+
+    // d > n
+    if sr > u32_bits - 1 {
+        return 0;
+    }
+
+    // d == 1
+    if sr == u32_bits - 1 {
+        return n;
+    }
+
+    sr = sr + 1;
+
+    // 1 <= sr <= u32_bits - 1
+    let mut q = n << (u32_bits - sr);
+    let mut r = n >> sr;
+
+    let mut carry = 0;
+    for _ in 0..sr {
+        // r:q = ((r:q) << 1) | carry
+        r = (r << 1) | (q >> (u32_bits - 1));
+        q = (q << 1) | carry;
+
+        // carry = 0;
+        // if r > d {
+        //     r -= d;
+        //     carry = 1;
+        // }
+
+        let s = (d.wrapping_sub(r).wrapping_sub(1)) as i32 >> (u32_bits - 1);
+        carry = (s & 1) as u32;
+        r -= d & s as u32;
+    }
+
+    q = (q << 1) | carry;
+    q
+}
+
+/// Returns `n / d` and sets `*rem = n % d`
+#[no_mangle]
+pub extern "C" fn __udivmodsi4(a: u32, b: u32, rem: Option<&mut u32>) -> u32 {
+    let d = __udivsi3(a, b);
+    if let Some(rem) = rem {
+        *rem = a - (d * b);
+    }
+    return d;
+}
+
+/// Returns `n / d` and sets `*rem = n % d`
+#[no_mangle]
+pub extern "C" fn __udivmoddi4(n: u64, d: u64, rem: Option<&mut u64>) -> u64 {
+    let u32_bits = u32::bits() as u32;
+    let u64_bits = u64::bits() as u32;
+
+    // NOTE X is unknown, K != 0
+    if n.high() == 0 {
+        if d.high() == 0 {
+            // 0 X
+            // ---
+            // 0 X
+
+            if let Some(rem) = rem {
+                *rem = u64::from(n.low() % d.low());
+            }
+            return u64::from(n.low() / d.low());
+        } else
+        // d.high() != 0
+        {
+            // 0 X
+            // ---
+            // K X
+
+            if let Some(rem) = rem {
+                *rem = u64::from(n.low());
+            }
+            return 0;
+        };
+    }
+
+    let mut sr;
+    let mut q = U64 { low: 0, high: 0 };
+    let mut r = U64 { low: 0, high: 0 };
+
+    // n.high() != 0
+    if d.low() == 0 {
+        if d.high() == 0 {
+            // K X
+            // ---
+            // 0 0
+
+            // NOTE copied verbatim from compiler-rt. This probably lets the intrinsic decide how to
+            // handle the division by zero (SIGFPE, 0, etc.). But this part shouldn't be reachable
+            // from safe code.
+            if let Some(rem) = rem {
+                *rem = u64::from(n.high() % d.low());
+            }
+            return u64::from(n.high() / d.low());
+        }
+
+        // d.high() != 0
+        if n.low() == 0 {
+            // K 0
+            // ---
+            // K 0
+
+            if let Some(rem) = rem {
+                *rem = U64 {
+                    low: 0,
+                    high: n.high() % d.high(),
+                }[..];
+            }
+            return u64::from(n.high() / d.high());
+        }
+
+        // n.low() != 0
+        // K K
+        // ---
+        // K 0
+
+        if d.high().is_power_of_two() {
+            if let Some(rem) = rem {
+                *rem = U64 {
+                    low: n.low(),
+                    high: n.high() & (d.high() - 1),
+                }[..];
+            }
+
+            return u64::from(n.high() >> d.high().trailing_zeros());
+        }
+
+        sr = d.high().leading_zeros().wrapping_sub(n.high().leading_zeros());
+
+        // D > N
+        if sr > u32_bits - 2 {
+            if let Some(rem) = rem {
+                *rem = n;
+            }
+            return 0;
+        }
+
+        sr = sr + 1;
+
+        // 1 <= sr <= u32_bits - 1
+        // q = n << (u64_bits - sr);
+        q.low = 0;
+        q.high = n.low() << (u32_bits - sr);
+        // r = n >> sr
+        r.high = n.high() >> sr;
+        r.low = (n.high() << (u32_bits - sr)) | (n.low() >> sr);
+    } else
+    // d.low() != 0
+    {
+        if d.high() == 0 {
+            // K X
+            // ---
+            // 0 K
+            if d.low().is_power_of_two() {
+                if let Some(rem) = rem {
+                    *rem = u64::from(n.low() & (d.low() - 1));
+                }
+
+                if d.low() == 1 {
+                    return n;
+                } else {
+                    let sr = d.low().trailing_zeros();
+                    return U64 {
+                        low: (n.high() << (u32_bits - sr)) | (n.low() >> sr),
+                        high: n.high() >> sr,
+                    }[..];
+                };
+            }
+
+            sr = 1 + u32_bits + d.low().leading_zeros() - n.high().leading_zeros();
+
+            // 2 <= sr <= u64_bits - 1
+            // q = n << (u64_bits - sr)
+            // r = n >> sr;
+            if sr == u32_bits {
+                q.low = 0;
+                q.high = n.low();
+                r.high = 0;
+                r.low = n.high();
+            } else if sr < u32_bits
+            // 2 <= sr <= u32_bits - 1
+            {
+                q.low = 0;
+                q.high = n.low() << (u32_bits - sr);
+                r.high = n.high() >> sr;
+                r.low = (n.high() << (u32_bits - sr)) | (n.low() >> sr);
+            } else
+            // u32_bits + 1 <= sr <= u64_bits - 1
+            {
+                q.low = n.low() << (u64_bits - sr);
+                q.high = (n.high() << (u64_bits - sr)) | (n.low() >> (sr - u32_bits));
+                r.high = 0;
+                r.low = n.high() >> (sr - u32_bits);
+            }
+
+        } else
+        // d.high() != 0
+        {
+            // K X
+            // ---
+            // K K
+
+            sr = d.high().leading_zeros().wrapping_sub(n.high().leading_zeros());
+
+            // D > N
+            if sr > u32_bits - 1 {
+                if let Some(rem) = rem {
+                    *rem = n;
+                    return 0;
+                }
+            }
+
+            sr += 1;
+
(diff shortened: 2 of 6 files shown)


Amanieu:
It seems that `__aeabi_uldivmod` uses a custom calling convention, so you'll have to implement it as a naked function. Also note that your 64-bit division function makes use of 32-bit division, so you'll have to implement that as well.

Amanieu:
You might want to rebase your implementation on top of #26 and make use of the traits.

src/arm.rs
```diff
@@ -1,3 +1,20 @@
+use core::mem;
+
+#[repr(C)]
+pub struct u64x2 {
+    a: u64,
+    b: u64,
+}
+
+#[no_mangle]
+pub unsafe extern "aapcs" fn __aeabi_uldivmod(num: u64, den: u64) -> u64x2 {
```
Amanieu: I don't think unsafe is necessary here. We only needed it for the mem* builtins.

src/lib.rs
```diff
@@ -38,3 +41,257 @@ absv_i2!(__absvsi2: i32);
 absv_i2!(__absvdi2: i64);
 // TODO(rust-lang/35118)?
 // absv_i2!(__absvti2, i128);
+
+#[no_mangle]
+pub extern "C" fn __udivmoddi4(a: u64, b: u64, rem: *mut u64) -> u64 {
```
Amanieu: Use `Option<&mut u64>` instead.

Amanieu: Also, I think this function should be moved into a separate file that contains all division-related functions.

japaric:
> It seems that __aeabi_uldivmod uses a custom calling convention, so you'll have to implement it as a naked function.

Alright, thanks for the input and done.

> Also note that your 64-bit division function makes use of 32-bit division, so you'll have to implement that as well.

Just ported that.

japaric:
> You might want to rebase your implementation on top of #26 and make use of the traits.

Nice. I'll get to it tomorrow.

src/lib.rs
```diff
@@ -295,3 +295,64 @@ pub extern "C" fn __udivmoddi4(a: u64, b: u64, rem: *mut u64) -> u64 {
     }
     q.u64()
 }
+
+#[no_mangle]
+pub extern "C" fn __udivmodsi4(a: u32, b: u32, rem: *mut u32) -> u32 {
```
Amanieu: Use `Option<&mut u32>`

src/lib.rs
```diff
@@ -295,3 +295,64 @@ pub extern "C" fn __udivmoddi4(a: u64, b: u64, rem: *mut u64) -> u64 {
...
+
+        // carry = 0;
+        // if r > d {
+        //     r -= d;
+        //     carry = 1;
+        // }
+
+        let s = (d.wrapping_sub(r).wrapping_sub(1)) as i32 >> (u32_bits - 1);
+        carry = (s & 1) as u32;
+        r -= d & s as u32;
```
Amanieu: Should this be a `wrapping_sub` as well?

src/lib.rs
```diff
@@ -38,3 +41,318 @@ absv_i2!(__absvsi2: i32);
...
+
+        // carry = 0
+        // if r.u64() >= d.u64() {
+        //     *r.all() -= d.u64();
+        //     carry = 1;
+        // }
+
+        let s = (d.u64().wrapping_sub(r.u64()).wrapping_sub(1)) as i64 >> (u64_bits - 1);
+        carry = (s & 1) as u32;
+        *r.all() -= d.u64() & s as u64;
```
Amanieu: `wrapping_sub`?

src/lib.rs
```diff
@@ -38,3 +41,318 @@ absv_i2!(__absvsi2: i32);
...
+    } else
+    // d.low != 0
+    {
+        if d.high == 0 {
+            // K X
+            // ---
+            // 0 K
+            if d.low.is_power_of_two() {
+                if let Some(rem) = unsafe { rem.as_mut() } {
+                    *rem = u64::from(n.low & (d.low - 1));
```
Amanieu: `wrapping_sub`?

src/lib.rs
```diff
@@ -38,3 +41,318 @@ absv_i2!(__absvsi2: i32);
...
+        // n.low != 0
+        // K K
+        // ---
+        // K 0
+
+        if d.high.is_power_of_two() {
+            if let Some(rem) = unsafe { rem.as_mut() } {
+                *rem = words {
+                        low: n.low,
+                        high: n.high & (d.high - 1),
```
Amanieu: `wrapping_sub`?

japaric:
> First, I get infinite recursion but that probably can be fixed with the hacks

So, even though `objdump`ing this crate shows infinite recursion, when I use this crate in other project to produce executables the infinite recursion disappears e.g.:

``` asm
0800013c <__aeabi_uidivmod>:
 800013c:       b500            push    {lr}
 800013e:       b081            sub     sp, #4
 8000140:       466a            mov     r2, sp
 8000142:       f000 f92b       bl      800039c <__udivmodsi4>
 8000146:       9900            ldr     r1, [sp, #0]
 8000148:       b001            add     sp, #4
 800014a:       bd00            pop     {pc}

0800014c <__aeabi_uldivmod>:
 800014c:       e92d 4800       stmdb   sp!, {fp, lr}
 8000150:       b084            sub     sp, #16
 8000152:       f10d 0c08       add.w   ip, sp, #8
 8000156:       f8cd c000       str.w   ip, [sp]
 800015a:       f000 f805       bl      8000168 <__udivmoddi4>
 800015e:       9a02            ldr     r2, [sp, #8]
 8000160:       9b03            ldr     r3, [sp, #12]
 8000162:       b004            add     sp, #16
 8000164:       e8bd 8800       ldmia.w sp!, {fp, pc}

08000168 <__udivmoddi4>:
 8000168:       e92d 4ff0       stmdb   sp!, {r4, r5, r6, r7, r8, r9, sl, fp, lr}
(...)
```

That's with LTO. Without LTO, I get pretty much the same.

src/lib.rs
japaric: I've only used `wrapping_sub` where the test suite was pointing out "overflow subtraction" panics. But, I guess it makes sense to use wrapping operations in these other operations as well.

Amanieu:
Yes, that's to be expected. Objdumping object files sometimes has issues with symbol resolution. Objdumping an executable should always show the correct symbols that are getting linked.

src/arm.rs
```diff
@@ -1,3 +1,35 @@
+use core::intrinsics;
+
```
Amanieu: A comment explaining why an asm implementation was necessary (non-standard calling convention) would be nice here.

src/lib.rs
```diff
@@ -38,3 +41,318 @@ absv_i2!(__absvsi2: i32);
...
+
+    let u32_bits = (mem::size_of::<u32>() * U8_BITS) as u32;
+    let u64_bits = (mem::size_of::<u64>() * U8_BITS) as u32;
+
+    let n = words::from(a);
+    let d = words::from(b);
+
+    // NOTE X is unknown, K != 0
+    if n.high == 0 {
+        return if d.high == 0 {
```
Amanieu: That looks very strange. Maybe you should put the return on the actual values like the C code does.

src/lib.rs
```diff
@@ -38,3 +41,318 @@ absv_i2!(__absvsi2: i32);
...
+    let mut r = words { low: 0, high: 0 };
+
+    // n.high != 0
+    if d.low == 0 {
+        if d.high == 0 {
+            // K X
+            // ---
+            // 0 0
+
+            // NOTE copied verbatim from compiler-rt, but does division by zero even make sense?
```
Amanieu: I think the idea here is to let the 32-bit division function decide how to handle division by zero (SIGFPE, return 0, etc). In this end it doesn't matter what you do since the user has already performed UB if you reach this point.

src/lib.rs
Amanieu: On second thought this isn't necessary here since `d.low` is a power of 2.

src/lib.rs
Amanieu: On second thought this isn't necessary here since `d.high` is a power of 2.

src/lib.rs
```diff
@@ -38,3 +41,318 @@ absv_i2!(__absvsi2: i32);
...
+    // *.r.all() = n.u64() >> sr
+    // 1 <= sr <= u64_bits - 1
+    let mut carry = 0;
+
+    for _ in 0..sr {
+        // r:q = ((r:q) << 1) | carry
+        r.high = (r.high << 1) | (r.low >> (u32_bits - 1));
+        r.low = (r.low << 1) | (q.high >> (u32_bits - 1));
+        q.high = (q.high << 1) | (q.low >> (u32_bits - 1));
+        q.low = (q.low << 1) | carry;
```
Amanieu: You can do a normal 64-bit shift here, no need to mess with the carry bits:
```rust
r = (r << 1) | (q >> 63);
q = (q << 1) | carry;
```

src/lib.rs
Amanieu: Also probably not necessary here since `r >= d`.

src/lib.rs
Amanieu: Also probably not necessary here since `r >= d`.

japaric:
@Amanieu I think I addressed all your comments. Let me know if I missed anything.

src/div.rs
```diff
@@ -0,0 +1,278 @@
...
+            // ---
+            // 0 0
+
+            // NOTE copied verbatim from compiler-rt. This probably lets the intrinsic decide how to
+            // handle the division by zero (SIGFPE, 0, etc.). But this part shouldn't be reachable
+            // from safe code.
+            if let Some(rem) = rem {
+                *rem = u64::from(n.high() % d.low());
+            }
+            return u64::from(n.high() / d.low());
```
Amanieu: Actually I think we could just go with `panic!("Division by zero")` here.

src/div.rs
```diff
@@ -0,0 +1,278 @@
+use {Int, LargeInt, U64};
+
+/// Returns `n / d`
+#[no_mangle]
+pub extern "C" fn __udivsi3(n: u32, d: u32) -> u32 {
+    let u32_bits = u32::bits() as u32;
+
+    // Special cases
+    if d == 0 {
+        return 0; // ?!
```
Amanieu: `panic!("Division by zero")`

src/div.rs
```diff
@@ -0,0 +1,278 @@
...
+        //     carry = 1;
+        // }
+
+        let s = (d.wrapping_sub(r).wrapping_sub(1)) as i32 >> (u32_bits - 1);
+        carry = (s & 1) as u32;
+        r -= d & s as u32;
+    }
+
+    q = (q << 1) | carry;
+    q
```
Amanieu: Just return `(q << 1) | carry` directly

Amanieu:
I think that covers everything, it should be good to go after those changes.

japaric:
Merging (test failures (ARM) are known to be flaky)

Thanks @Amanieu for the review!

@thejpster I hope you can give this crate another try! Let us know if you need other intrinsics!

thejpster:
Humm. I remembered to set the no-compiler-rt boolean in the target file back to false, but something else seems to have gone wrong.

```
$ ./build.sh 
/home/jonathan/.cargo/bin/xargo
   Compiling sysroot for lm4f120
   Compiling core v0.0.0 (file:///home/jonathan/.xargo/src/libcore)
     Running `rustc /home/jonathan/.xargo/src/libcore/lib.rs --crate-name core --crate-type lib -C opt-level=3 --out-dir /tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps --emit=dep-info,link --target lm4f120 -L dependency=/tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps`
   Compiling alloc v0.0.0 (file:///home/jonathan/.xargo/src/liballoc)
   Compiling rand v0.0.0 (file:///home/jonathan/.xargo/src/librand)
   Compiling rustc_unicode v0.0.0 (file:///home/jonathan/.xargo/src/librustc_unicode)
     Running `rustc /home/jonathan/.xargo/src/liballoc/lib.rs --crate-name alloc --crate-type lib -C opt-level=3 --out-dir /tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps --emit=dep-info,link --target lm4f120 -L dependency=/tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps --extern core=/tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps/libcore.rlib`
     Running `rustc /home/jonathan/.xargo/src/librand/lib.rs --crate-name rand --crate-type lib -C opt-level=3 --out-dir /tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps --emit=dep-info,link --target lm4f120 -L dependency=/tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps --extern core=/tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps/libcore.rlib`
     Running `rustc /home/jonathan/.xargo/src/librustc_unicode/lib.rs --crate-name rustc_unicode --crate-type lib -C opt-level=3 --out-dir /tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps --emit=dep-info,link --target lm4f120 -L dependency=/tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps --extern core=/tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps/libcore.rlib`
   Compiling collections v0.0.0 (file:///home/jonathan/.xargo/src/libcollections)
     Running `rustc /home/jonathan/.xargo/src/libcollections/lib.rs --crate-name collections --crate-type lib -C opt-level=3 --out-dir /tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps --emit=dep-info,link --target lm4f120 -L dependency=/tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps --extern core=/tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps/libcore.rlib --extern alloc=/tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps/liballoc.rlib --extern rustc_unicode=/tmp/xargo.d54QJuI0qosA/target/lm4f120/release/deps/librustc_unicode.rlib`
    Finished release [optimized] target(s) in 131.48 secs
       Fresh spin v0.4.3
       Fresh rlibc v1.0.0
       Fresh rustc_builtins v0.1.0 (https://github.com/japaric/rustc-builtins.git#d546f187)
   Compiling bare-metal-arm-rust v0.3.0 (file:///home/jonathan/Documents/programming/rust/bare-metal-arm-rust)
       Fresh lazy_static v0.2.1
       Fresh linked_list_allocator v0.2.2
     Running `/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/debug/build/bare-metal-arm-rust-4ad68ea8a08ec1a3/build-script-build`
       Fresh primer v0.2.0 (file:///home/jonathan/Documents/programming/rust/bare-metal-arm-rust/libs/primer)
     Running `rustc src/main.rs --crate-name bare_metal_arm_rust --crate-type bin -C panic=abort -g --out-dir /home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug --emit=dep-info,link --target lm4f120 -C ar=arm-none-eabi-ar -C linker=arm-none-eabi-gcc -L dependency=/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/deps --extern primer=/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/deps/libprimer.rlib --sysroot=/home/jonathan/.xargo -L native=/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/build/bare-metal-arm-rust-4ad68ea8a08ec1a3/out`
error: linking with `arm-none-eabi-gcc` failed: exit code: 1 
note: "arm-none-eabi-gcc" "-mcpu=cortex-m4" "-mthumb" "-Tlm4f120.ld" "-L" "/home/jonathan/.xargo/lib/rustlib/lm4f120/lib" "/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/bare_metal_arm_rust.0.o" "-o" "/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/bare_metal_arm_rust" "-Wl,--gc-sections" "-nodefaultlibs" "-L" "/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/deps" "-L" "/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/build/bare-metal-arm-rust-4ad68ea8a08ec1a3/out" "-L" "/home/jonathan/.xargo/lib/rustlib/lm4f120/lib" "-Wl,-Bstatic" "-Wl,-Bdynamic" "/home/jonathan/.xargo/lib/rustlib/lm4f120/lib/libcollections.rlib" "/home/jonathan/.xargo/lib/rustlib/lm4f120/lib/librustc_unicode.rlib" "/home/jonathan/.xargo/lib/rustlib/lm4f120/lib/liballoc.rlib" "/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/deps/libprimer.rlib" "/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/deps/liblinked_list_allocator-e85086c046371af8.rlib" "/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/deps/liblazy_static-359f5533c970cd71.rlib" "/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/deps/libspin-267d812cb8c611d8.rlib" "/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/deps/librlibc-975886925ea7e4d8.rlib" "/home/jonathan/.xargo/lib/rustlib/lm4f120/lib/libcore.rlib" "-nostdlib" "-nostartfiles" 
note: /home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/deps/libprimer.rlib(primer.0.o): In function `primer::lm4f120h5qr::systick::run_time_us':
/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/libs/primer/src/lm4f120h5qr/systick.rs:171: undefined reference to `__aeabi_uldivmod'
collect2: error: ld returned 1 exit status
```

thejpster:
https://github.com/thejpster/bare-metal-arm-rust/commit/3c901c2e4422f64bd3eaafebcdee63af5f174e3c

Looks good to me. I reported a failure earlier, but I'd forgotten to `extern create rustc_builtins`!

Amanieu:
I don't think you need `extern crate`, you should only need to add the dependency to `Cargo.toml`.

thejpster:
Should no_compiler_rt be true or false?

Amanieu:
true

thejpster:
If I don't `extern crate` I get:

```
note: /home/jonathan/Documents/programming/rust/bare-metal-arm-rust/target/lm4f120/debug/deps/libprimer.rlib(primer.0.o): In function `primer::lm4f120h5qr::systick::run_time_us':
/home/jonathan/Documents/programming/rust/bare-metal-arm-rust/libs/primer/src/lm4f120h5qr/systick.rs:171: undefined reference to `__aeabi_uldivmod'
collect2: error: ld returned 1 exit status
```

Amanieu:
Oh, I guess it is required then. My bad.

japaric:
@thejpster 

> Looks good to me. I reported a failure earlier, but I'd forgotten to extern create rustc_builtins!

Nice! Experienced any significant change in binary size after moving to rustc_builtins?

Amanieu:
I would expect it to be slightly slower since we aren't using the asm implementation of division from compiler-rt and instead use a rust implementation.

thejpster:
Hard to say as if I build in release mode, I hit a trap on startup. The
Debug binary is obviously quite big.

If I get it going I'll switch back to compiler-rt.rs temporarily and let
you know.

japaric:
I did some measurements on this snippet:

``` rust
#[no_mangle]
pub extern "C" fn start() -> ! {
    unsafe {
        let x = ptr::read_volatile(0x0 as *const u64);
        let y = ptr::read_volatile(0x0 as *const u64);
        let z = x / y;
        ptr::write_volatile(0x0 as *mut u64, z);

        loop {}
    }
}
```

compiler-rt.rs (intrinsics in C/asm)

```
   text    data     bss     dec     hex
    976       0       0     976     3d0
```

This crate (intrinsics in Rust)

```
   text    data     bss     dec     hex
   1046       0       0    1046     416
```

Build using Cargo's release profile with LTO enabled for ARM Cortex-M3.

And for reference, this program:

``` rust
#[no_mangle]
pub extern "C" fn start() -> ! {
    loop {}
}
```

has this size:

```
   text    data     bss     dec     hex
    294       0       0     294     126
```

70 bytes bigger doesn't sound that bad. If as @Amanieu says there are other optimizations we could do here then perhaps that can bring down the size on par with compiler-rt's version maybe even make it smaller(!).

PR #30, opened 2016-08-13
Various changes

Amanieu:
- Cleaned up div implementation and added the rest of the unsigned division builtins
- Added integer multiply builtins
- Added shift builtins
- Removed absv
- Disabled `no_mangle` for tests so we can compare with the native version
- Optimized the ARM asm a bit


japaric:
Awesome! Looks good to me, but there's an issue around tests:

Quickcheck is biased in its generation of integer values; it only generates values in a small range (`[-100, 100]`) (cf burntsushi/quickcheck#119), so these tests won't hit important branches that e.g. deal with overflow. This is why I was generating a `u64` from 2 `u32`s in the `udiv*` tests. I'm going to land this and we can address the test coverage later in a more general way.

On another note, I re-measured [my `u64 % u64` program](https://github.com/japaric/rustc-builtins/pull/25#issuecomment-239296799) and this PR brings the binary size down to 946 bytes, which is 30 bytes smaller than the `compiler-rt.rs` version :+1:.

PR #34, opened 2016-08-13
Fix inline assembly to work on ARMv6-M

### LINKED ISSUE #33: Can't compile for `thumbv6-none-eabi`
Because some of our assembly implementation use thumb2 instructions that aren't available for it.

```
$ xargo build --target thumbv6m-none-eabi
   Compiling rustc_builtins v0.1.0 (file:///home/japaric/tmp/rustc-builtins)
error: <inline asm>:5:11: error: instruction requires: thumb2
          ldr r1, [sp], #4
          ^

  --> src/arm.rs:8:5
   |
8  |     asm!("push {lr}
   |     ^

error: <inline asm>:2:11: error: instruction requires: thumb2
          sub r12, sp, #12
          ^

  --> src/arm.rs:20:5
   |
20 |     asm!("push {lr}
   |     ^

error: <inline asm>:3:11: error: instruction requires: arm-mode
          str r12, [sp, #-20]!
          ^

  --> src/arm.rs:20:5
   |
20 |     asm!("push {lr}
   |     ^

error: <inline asm>:5:11: error: instruction requires: arm-mode
          ldrd r2, r3, [sp, #8]
          ^

  --> src/arm.rs:20:5
   |
20 |     asm!("push {lr}
   |     ^

error: aborting due to 4 previous errors

error: Could not compile `rustc_builtins`.

To learn more, run the command again with --verbose.
error: `cargo` process didn't exit successfully
```

Possible solutions:

- Change the assembly implementations to not use thumb2 instructions.
- Add a special implementation just for this target. The problem is that, AFAIK, there's no `cfg` argument that handles this conditional compilation. We'll likely need some way to `cfg` on the arch component of `llvm-target`.

cc @Amanieu

Amanieu:
Fixes #33

### DIFF
--- a/src/arm.rs
+++ b/src/arm.rs
@@ -9,7 +9,8 @@ pub unsafe fn __aeabi_uidivmod() {
           sub sp, sp, #4
           mov r2, sp
           bl __udivmodsi4
-          ldr r1, [sp], #4
+          ldr r1, [sp]
+          add sp, sp, #4
           pop {pc}");
     intrinsics::unreachable();
 }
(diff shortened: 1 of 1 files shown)


japaric:
Thanks!

PR #36, opened 2016-08-13
quickcheck: better generation of input arguments

### LINKED ISSUE #31: Quickcheck and test coverage
Quickcheck generates integer values in a small range around zero (`[-100, 100]`) (cf burntsushi/quickcheck#119). These values are not enough to tests all the branches of some intrinsics implementation. We need better control over the generated values; here's are some possible solutions of the top of my head:

- Synthesize integer values from two halves. That's is use `(u32, u32)` as quickcheck argument and then use `u64::from_parts(some_u32, another_u32)` in the actual test.
- Create a newtype, e.g. `U64(u64)`, and implement the `Arbitrary` trait for it, but have the implementation generate values in the `[MIN, MAX]` range.

japaric:
closes #31 

cc @Amanieu tests for muldi and udivdi3 are now failing due to overflows. I've yet to investigate.

### DIFF
--- a/src/qc.rs
+++ b/src/qc.rs
@@ -1,0 +1,123 @@
+//  When testing functions, QuickCheck (QC) uses small values for integer (`u*`/`i*`) arguments
+// (~ `[-100, 100]`), but these values don't stress all the code paths in our intrinsics. Here we
+// create newtypes over the primitive integer types with the goal of having full control over the
+// random values that will be used to test our intrinsics.
+
+use std::boxed::Box;
+use std::fmt;
+
+use quickcheck::{Arbitrary, Gen};
+
+use LargeInt;
+
+// Generates values in the full range of the integer type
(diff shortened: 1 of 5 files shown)


src/qc.rs
```diff
@@ -0,0 +1,123 @@
...
+
+        impl fmt::Debug for $TY {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                fmt::Debug::fmt(&self.0, f)
+            }
+        }
+    }
+}
+
+arbitrary!(I32: i32);
```
japaric: I wasn't sure if these should also use the `arbitrary_large` generation method. Does any intrinsic have special handling of cases where the lower/higher half-word is zero?

Amanieu:
Ah yes, nice catch in `muldi`, it looks fine to me.

PR #39, opened 2016-08-16
Use link_args to alias arm builtins

mattico:
This is a bit of a hack, but it seems to me like the best way we have to create function aliases right now. It's up to you whether this is better than just having more function stubs.

Note that I haven't actually tested this yet as I don't have access to my arm toolchain.

### DIFF
--- a/src/arm.rs
+++ b/src/arm.rs
@@ -37,6 +37,7 @@ extern "C" {
 }
 
 // Create aliases for the *4 and *8 variants
+#[cfg(not(test))]
 #[link_args = "-Wl,--defsym=__aeabi_memcpy4=__aeabi_memcpy -Wl,--defsym=__aeabi_memcpy8=__aeabi_memcpy"]
 extern {}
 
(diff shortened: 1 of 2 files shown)


Amanieu:
`arm-linux-gnueabi-gcc: error: unrecognized command line option '--defsym=__aeabi_memcpy8=__aeabi_memcpy'`

Try using `-Wl,--defsym=__aeabi_memcpy8=__aeabi_memcpy` instead so that gcc passes the arguments to the linker.

japaric:
I couldn't get this to work locally. When linking an executable (from [cu](https://github.com/japaric/cu)), these --defsym arguments didn't appear in the linker arguments (as seen from -Z print-link-args) and I got "an undefined reference to __aeabi_memcpy4" error.

But if we can get this to work, then it seems fine to add to me. I don't know if these arguments are gcc specific but rustc only knows how to gcc as a linker, and no other linker, anyway so it doesn't matter. This change also seems easy to revert if link_args gets removed and removal shouldn't be visible by downstream users.

Amanieu:
You need to mark these with `#[cfg(not(test))]`.

mattico:
Thanks for helping me code-by-proxy @Amanieu. I'm finally back on my real computer so I can test this myself, but I expect there aren't any changes left to make after that last one :P

mattico:
@japaric I'll see if I can actually get this working locally now.

japaric:
This is blocked by rust-lang/rust#36176. `link_args` are not propagated from rlibs to binaries.

japaric:
re: `link_args`. Alex [said]:

[said]: https://github.com/rust-lang/rust/issues/36176#issuecomment-243954914

> This "feature" has basically always been slated for deprecation/removal.

So we can't really on `link_args` to do this :-(. Closing ...

Regardless, thanks for the PR @mattico!

PR #43, opened 2016-08-17
Add soft float addition builtins

mattico:
TODO:
- [x] ~~Move `rep_clz` to `int` module?~~
- [x] Use integer associated type for Float trait again
- [x] Module-ify Int/Float functions
- [x] Use wrapping_* ops where appropriate (everywhere?)
- [x] Finish tests
- [x] Fix `from_repr(1) + from_repr(1) == INFINITY`

### DIFF
--- a/src/float/add.rs
+++ b/src/float/add.rs
@@ -1,0 +1,326 @@
+use core::num::Wrapping;
+use float::Float;
+
+macro_rules! add {
+    ($intrinsic:ident: $ty:ty) => {
+        /// Returns `a + b`
+        #[allow(unused_parens)]
+        #[cfg_attr(not(test), no_mangle)]
+        pub extern fn $intrinsic(a: $ty, b: $ty) -> $ty {
+            let one = Wrapping(1 as <$ty as Float>::Int);
+            let zero = Wrapping(0 as <$ty as Float>::Int);
+
+            let bits =             Wrapping(<$ty>::bits() as <$ty as Float>::Int);
+            let significand_bits = Wrapping(<$ty>::significand_bits() as <$ty as Float>::Int);
+            let exponent_bits =    bits - significand_bits - one;
+            let max_exponent =     (one << exponent_bits.0 as usize) - one;
+
+            let implicit_bit =     one << significand_bits.0 as usize;
+            let significand_mask = implicit_bit - one;
+            let sign_bit =         one << (significand_bits + exponent_bits).0 as usize;
+            let abs_mask =         sign_bit - one;
+            let exponent_mask =    abs_mask ^ significand_mask;
+            let inf_rep =          exponent_mask;
+            let quiet_bit =        implicit_bit >> 1;
+            let qnan_rep =         exponent_mask | quiet_bit;
+
+            let mut a_rep = Wrapping(a.repr());
+            let mut b_rep = Wrapping(b.repr());
+            let a_abs = a_rep & abs_mask;
+            let b_abs = b_rep & abs_mask;
+
+            // Detect if a or b is zero, infinity, or NaN.
+            if a_abs - one >= inf_rep - one ||
+                b_abs - one >= inf_rep - one {
+                // NaN + anything = qNaN
+                if a_abs > inf_rep {
+                    return (<$ty as Float>::from_repr((a_abs | quiet_bit).0));
+                }
+                // anything + NaN = qNaN
+                if b_abs > inf_rep {
+                    return (<$ty as Float>::from_repr((b_abs | quiet_bit).0));
+                }
+
+                if a_abs == inf_rep {
+                    // +/-infinity + -/+infinity = qNaN
+                    if (a.repr() ^ b.repr()) == sign_bit.0 {
+                        return (<$ty as Float>::from_repr(qnan_rep.0));
+                    } else {
+                        // +/-infinity + anything remaining = +/- infinity
+                        return a;
+                    }
+                }
+
+                // anything remaining + +/-infinity = +/-infinity
+                if b_abs == inf_rep {
+                    return b;
+                }
+
+                // zero + anything = anything
+                if a_abs.0 == 0 {
+                    // but we need to get the sign right for zero + zero
+                    if b_abs.0 == 0 {
+                        return (<$ty as Float>::from_repr(a.repr() & b.repr()));
+                    } else {
+                        return b;
+                    }
+                }
+
+                // anything + zero = anything
+                if b_abs.0 == 0 {
+                     return a;
+                }
+            }
+
+            // Swap a and b if necessary so that a has the larger absolute value.
+            if b_abs > a_abs {
+                let temp = a_rep;
+                a_rep = b_rep;
+                b_rep = temp;
+            }
+
+            // Extract the exponent and significand from the (possibly swapped) a and b.
+            let mut a_exponent = Wrapping((a_rep >> significand_bits.0 as usize & max_exponent).0 as i32);
+            let mut b_exponent = Wrapping((b_rep >> significand_bits.0 as usize & max_exponent).0 as i32);
+            let mut a_significand = a_rep & significand_mask;
+            let mut b_significand = b_rep & significand_mask;
+
+            // normalize any denormals, and adjust the exponent accordingly.
+            if a_exponent.0 == 0 {
+                let (exponent, significand) = <$ty>::normalize(a_significand.0);
+                a_exponent = Wrapping(exponent);
+                a_significand = Wrapping(significand); 
+            }
+            if b_exponent.0 == 0 {
+                let (exponent, significand) = <$ty>::normalize(b_significand.0);
+                b_exponent = Wrapping(exponent);
+                b_significand = Wrapping(significand); 
+            }
+
+            // The sign of the result is the sign of the larger operand, a.  If they
+            // have opposite signs, we are performing a subtraction; otherwise addition.
+            let result_sign = a_rep & sign_bit;
+            let subtraction = ((a_rep ^ b_rep) & sign_bit) != zero;
+
+            // Shift the significands to give us round, guard and sticky, and or in the
+            // implicit significand bit.  (If we fell through from the denormal path it
+            // was already set by normalize(), but setting it twice won't hurt
+            // anything.)
+            a_significand = (a_significand | implicit_bit) << 3;
+            b_significand = (b_significand | implicit_bit) << 3;
+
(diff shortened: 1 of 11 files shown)


Amanieu:
You can use `u32::leading_zeros`, it's already available from libcore.

src/float.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// Returns the leading zeroes of `a`
+    fn rep_clz(a: T) -> T;
+
+    /// Returns a normalized version of the significand
+    fn normalize(significand: &mut T) -> T {
+        let shift = Self::rep_clz(*significand) - Self::rep_clz(T::one() << Self::significand_bits());
+        *significand <<= shift as T;
+        T::one() - shift as T
+    }  
+}
```
Amanieu: I would suggest a simpler version of this trait:
```rust
trait Float {
type Int: Sized + Copy + Clone + Sub<Output=T> + ShlAssign<T> + Shl<T, Output=T> + One;
fn bits() -> u32; // Always use u32 for bit widths
fn significand_bits() -> u32; 
fn normalize(Self::Int) -> Self::Int;
}
```

Amanieu:
I think we should move all int related stuff into an int/ subdirectory and all the float related stuff into float/. The source tree will then look like this:
src/int/mod.rs
src/int/mul.rs
src/int/udiv.rs
src/float/mod.rs
src/float/add.rs

src/float.rs
```diff
@@ -0,0 +1,85 @@
...
+    fn bits() -> u32 {
+        32
+    }
+    fn significand_bits() -> u32 {
+        23
+    }
+    fn rep_clz(a: u32) -> u32 {
+        #[cfg(not(msvc))]
+        extern "C" {
+            fn __builtin_clz(a: u32) -> u32;
```
japaric: Where does this symbol come from? And is this even necessary? Couldn't this just use `u32::leading_zeros`? That method (and the u64 variant) already lower to an LLVM intrinsic: `llvm.ctlz.i*` in LLVM IR.

mattico:
@Amanieu I don't think you can use leading_zeroes when compiling with `no_builtin`. Either that or I'm having some strange error that doesn't allow me to call methods on primitives.

```
error: no method named `leading_zeroes` found for type `u64` in the current scope
```

I agree with your other comments.

Amanieu:
It's `leading_zeros` not `leading_zeroes`.

mattico:
@Amanieu *facepalm*

src/add.rs
```diff
@@ -0,0 +1,189 @@
...
+            let max_exponent =     ((1 << exponent_bits) - 1) as <$ty as Float>::Int;
+
+            let implicit_bit =     (1 << significand_bits) as <$ty as Float>::Int;
+            let significand_mask = (implicit_bit - 1) as <$ty as Float>::Int;
+            let sign_bit =         (1 << (significand_bits + exponent_bits)) as <$ty as Float>::Int;
+            let abs_mask =         (sign_bit - 1) as <$ty as Float>::Int;
+            let exponent_mask =    (abs_mask ^ significand_mask) as <$ty as Float>::Int;
+            let inf_rep =          (exponent_mask) as <$ty as Float>::Int;
+            let quiet_bit =        (implicit_bit >> 1) as <$ty as Float>::Int;
+            let qnan_rep =         (exponent_mask | quiet_bit) as <$ty as Float>::Int;
```
Amanieu: You can probably get rid of all the `as <$ty as Float>::Int` casts here. Type inference will figure it out.

Amanieu:
Can you move the transmutes into the `Float` trait? That way you keep all the unsafe code in one place.
```rust
fn repr(self) -> Self::Int;
fn from_repr(Self::Int) -> Self;
```

And generally anything that might be reused by the other soft-float implementations should go into that trait.

mattico:
@Amanieu that makes a lot of sense, I'll do that.

src/float/add.rs
```diff
@@ -0,0 +1,193 @@
...
+add!(__adddf3: f64);
+
+// FIXME: Implement these using aliases
+#[cfg(arm)]
+#[cfg_attr(not(test), no_mangle)]
+pub extern fn __aeabi_dadd(a: f64, b: f64) -> f64 {
+    __adddf3(a, b)
+}
+
+#[cfg(arm)]
```
Amanieu: `#[cfg(target_arch = "arm")]`

src/float/mod.rs
```diff
@@ -0,0 +1,67 @@
+use core::mem::transmute;
+use core::ops::*;
+
+pub mod add;
+
+/// Trait for some basic operations on floats
+pub trait Float: Sized {
+    /// A uint of the same with as the float
+    type Int: Sized + Copy + Clone + Sub<Output=Self::Int> +
```
Amanieu: You can just use plain `type Int;` here, like I did with the `LargeInt` trait.

Amanieu:
@mattico Can you add tests for this? See the integer builtins for examples.

@japaric We'll want to make sure the tests properly cover "exotic" floating-point values like infinity and NaN.

Amanieu:
For now tests with plain `f32` and `f64` should be enough.

mattico:
@Amanieu yeah, I've started writing the tests, using quickcheck and some regular ones for the special cases. Doing this has identified a few issues:
- Need to use overflowing_* ops to match the original C code (lots of panics)
- Something's wrong with the logic somewhere (`from_repr(1) + from_repr(1) == INFINITY`)

japaric:
> We'll want to make sure the tests properly cover "exotic" floating-point values like infinity and NaN.

Yeah, we can add `F32`/`F64` wrappers to handle this in a follow-up PR.

> I started writing the tests earlier, using quickcheck and some regular ones for the special cases.

This is good enough for this PR. :+1:

mattico:
Rebased on master, I'll cleanup the history once this works.

src/float/add.rs
```diff
@@ -0,0 +1,268 @@
...
+mod test {
+    use core::{f32, f64};
+    use qc::{U32, U64};
+    use float::Float;
+
+    // TODO: Add F32/F64 to qc so that they print the right values (at the very least)
+    quickcheck! {
+        fn addsf3(a: U32, b: U32) -> bool {
+            let (a, b) = (a.0, b.0);
+            let r = super::__addsf3(<f32 as Float>::from_repr(a), <f32 as Float>::from_repr(b));
```
japaric: I think this can simply be `f32::from_repr`. Same in the `f64` case.

src/float/add.rs
```diff
@@ -0,0 +1,268 @@
...
+    use core::{f32, f64};
+    use qc::{U32, U64};
+    use float::Float;
+
+    // TODO: Add F32/F64 to qc so that they print the right values (at the very least)
+    quickcheck! {
+        fn addsf3(a: U32, b: U32) -> bool {
+            let (a, b) = (a.0, b.0);
+            let r = super::__addsf3(f32::from_repr(a), f32::from_repr(b));
+            r == f32::from_repr(a) + f32::from_repr(b)
```
japaric: Hmm, is okay to use *equality* here in the case of hard float targets? For those targets, the `f32 + f32` operation will be lowered to a FPU instruction. Are these software implementations guaranteed to perform their operations with the exact same precision as FPUs in all scenarios? If not, these tests should use *approximate* equality for hard float targets and equality for soft float targets -- in the latter case we are comparing our software implementation to `libcompiler-rt.a`'s so the results should match exactly.

Thoughts @Amanieu?

japaric: > I think we should leave strict equality for now and revisit this later if we find that it causes problems.

Sounds good to me.

> However I think we should be comparing the raw bits of the float, which means transmuting it to an u32 before comparing it

Yeah, I agree raw bits comparison is better. NaN has multiple, different representations, right? (as in there are a bunch of NaNs among the possible 2^32 single precision float values) Even raw bits equality may fail in the NaN case.

mattico: @japaric yes, I thought about this a bit but came to the same conclusion as @Amanieu.

@Amanieu 
> However I think we should be comparing the raw bits of the float, which means transmuting it to an u32 before comparing it.

Yep, I did this for the explicit NaN tests, but neglected to do it for the qc ones.

mattico:
Another thing that might cause an issue is constant folding, which might give different results than actual runtime addition.

mattico:
I'm starting to think the use of `Wrapping()` types was a mistake. They make casts between integer types very verbose.

This last commit fixed the common qc test failure case of repr(1) + repr(1), so the tests occasionally pass, but there is still at least one case that fails sometimes: repr(1) + repr(<large number>).

src/float/add.rs
```diff
@@ -0,0 +1,326 @@
...
+    // doesn't care about calculating the 'correct' one, if both values are NaN
+    // the values are considered equivalent.
+
+    // TODO: Add F32/F64 to qc so that they print the right values (at the very least)
+    quickcheck! {
+        fn addsf3(a: U32, b: U32) -> bool {
+            let (a, b) = (f32::from_repr(a.0), f32::from_repr(b.0));
+            let x = super::__addsf3(a, b);
+            let y = a + b;
+            if !(x.is_nan() && y.is_nan()) {
```
japaric: I personally find this easier to read as:

``` rust
if x.is_nan() && y.is_nan() {
    true
} else {
    x.repr() == y.repr()
}
```

japaric: I also think this if-else block should be written as an utility function because it repeats in several places.

japaric: I that utility function should be used everywhere we need to test for float "equality" -- i.e. in all these tests.

src/float/add.rs
```diff
@@ -0,0 +1,326 @@
...
+
+                // anything + zero = anything
+                if b_abs.0 == 0 {
+                     return a;
+                }
+            }
+
+            // Swap a and b if necessary so that a has the larger absolute value.
+            if b_abs > a_abs {
+                let temp = a_rep;
```
japaric: It may be possible to use `mem::swap` here.

japaric:
@mattico Thanks for working on this! Looks great to me. I left some comments about nits. I'd personally like to see tests that involve `+0` and `-0` as well (unless they are already there and I missed them) but we can tackle those with the `F32`/`F64` wrappers.

mattico:
Thanks for the review! I'll keep these comments in mind when writing the
next one.

I do handle +-0 in the follow up PR, although I'm not sure about the method
I'm using there.

japaric:
I'm going to merge this as is. We can add the utility function in a follow up PR. Thanks again @mattico!

PR #44, opened 2016-08-18
Add signed division functions

Amanieu:


src/arm.rs
```diff
@@ -30,6 +30,43 @@ pub unsafe fn __aeabi_uldivmod() {
     intrinsics::unreachable();
 }
 
+#[naked]
+#[cfg_attr(not(test), no_mangle)]
+pub unsafe fn __aeabi_idivmod() {
```
japaric: Would it make sense to add a test for this assembly routine (and the one below)? Just to check its expected signature (`fn(u32, u32) -> (u32,u32)` in this case, I think) works as expected?

japaric:
I thought the aeabi intrinsics could have been tested by calling `let (q, r) = __aeabi_uldivmod(n, d)` in the test function (instead of the current `asm!` block) but I guess is not possible because of the custom calling convention these intrinsics use?

Amanieu:
Yea, you need asm to call them with the custom convention.

src/sdiv.rs
```diff
@@ -0,0 +1,121 @@
...
+            let s = a >> (<$ty>::bits() - 1);
+            let a = (a ^ s) - s;
+            let r = (a as $uty) % (b as $uty);
+            (r as $ty ^ s) - s
+        }
+    }
+}
+
+macro_rules! divmod {
+    ($intrinsic:ident, $div:ident: $ty:ty, $uty:ty) => {
```
japaric: $uty is not used in this macro?

src/sdiv.rs
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+}
+
+div!(__divsi3: i32, u32);
+div!(__divdi3: i64, u64);
+mod_!(__modsi3: i32, u32);
+mod_!(__moddi3: i64, u64);
+divmod!(__divmodsi4, __divsi3: i32, u32);
+divmod!(__divmoddi4, __divdi3: i64, u64);
```
japaric: The divmoddi4 intrinsic was marked as "never called by LLVM" in the README. Should it by marked as implemented in the README or not added at all?

src/arm.rs
```diff
@@ -30,6 +30,43 @@ pub unsafe fn __aeabi_uldivmod() {
...
+          bl __divmoddi4
+          ldr r2, [sp, #8]
+          ldr r3, [sp, #12]
+          add sp, sp, #16
+          pop {r4, pc}");
+    intrinsics::unreachable();
+}
+
+#[cfg_attr(not(test), no_mangle)]
+pub extern "C" fn __aeabi_uidiv(a: u32, b: u32) -> u32 {
```
japaric: This intrinsic is supposed to be an alias, right? Could you add a note about that?

japaric:
> Yea, you need asm to call them with the custom convention.

OK. Thanks for clarifying.

I left some comments about nits. r=me after those are addressed.

src/sdiv.rs
Amanieu: It's only called by the ARM builtins indirectly. I'll add it to the README anyways.

japaric:
Thanks @Amanieu!

PR #48, opened 2016-08-20
Add float quickcheck implementations (fixes #45)

mattico:
This builds on top of #43 to add Quickcheck helpers for floating point.

I generate a few different types of floats by constructing them from their component parts, using new functions defined on the Float trait. The Add implementation could be refactored to use these new functions for clarity.

PR #49, opened 2016-08-21
qc: bias towards generation of edge cases

japaric:
cc @mattico note the use of other `Rng` methods to control how often special ~~methods~~ cases get generated


mattico:
@japaric Neat! I missed a whole section of the quickcheck docs. I'll apply the same technique to #48.

PR #50, opened 2016-08-21
use utility function to compare the repr of floats

japaric:
follow up of #43

r? @mattico

### DIFF
--- a/src/float/add.rs
+++ b/src/float/add.rs
@@ -1,4 +1,6 @@
+use core::mem;
 use core::num::Wrapping;
+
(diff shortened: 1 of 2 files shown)


mattico:
LGTM, though there's likely to be a merge conflict between this and #48. I can rebase that one if you want to merge this first.

japaric:
I think we'll have to first sort out the soft float arm target test failure in #48 before it lands. Does merging this first sounds good to you?

mattico:
Yes that makes sense.

PR #53, opened 2016-08-21
Gist disassembly in CI (fixes #51)

### LINKED ISSUE #51: Upload CI disassembilies to Pastebin/Gist
The disassembly makes the CI logs really long, which makes the page load slowly, and could obscure errors which happen after them. This problem is only going to get worse as the library gets bigger. Uploading the disassembly to \<somewhere\> makes it fairly easy to have a disassembly available for every build, but makes the logs more manageable.

mattico:
Let's see if this works the first time ;)


japaric:
I think we can also use the [gist] crate. `objdump $FILE | gist -a -f "blah blah blah"`. It should be easier to install everywhere: just `cargo install gist`

[gist]: https://crates.io/crates/gist

mattico:
@japaric that sounds like a better option.

PR #54, opened 2016-08-22
CI: gist the disassemblies

### LINKED ISSUE #51: Upload CI disassembilies to Pastebin/Gist

japaric:
closes #53 
fixes #51 

cc @mattico 
*crosses fingers*

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,9 +21,8 @@ test_script:
 
 on_success:
   - cmd: |
-    CALL "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat"
-    dumpbin /disasm target/release/librustc_builtins.rlib |
-    gist -d "'%TARGET%/rustc-builtins.rlib' from commit '%APPVEYOR_REPO_COMMIT%' on branch '%APPVEYOR_REPO_branch%'"
+      CALL "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat"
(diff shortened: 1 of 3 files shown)


japaric:
[Linux](https://travis-ci.org/japaric/rustc-builtins/jobs/154306543#L356)

I can update the docker image to include the openssl header files. Maybe I can even pre-install the gist crate.

[OSX](https://travis-ci.org/japaric/rustc-builtins/jobs/154306548#L271)

There seems to be a legit error on the osx targets. "API error+"?

[Windows](https://ci.appveyor.com/project/japaric/rustc-builtins/build/1.0.205/job/ud7wig28p7y3bfn4#L99)

And I'm not sure how to install openssl on appveyor.

japaric:
> openssl

(I hope that someday a production grade ssl library written in Rust will appear so we don't have to rely on openssl anymore)

japaric:
> CALL "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat"
dumpbin /disasm target/release/librustc_builtins.rlib |
The syntax of the command is incorrect.
Command exited with code 255

Windows shell doesn't like pipes? or may we just need to remove the newline after the pipe.

japaric:
It passes everything now but OSX is not uploading a gist. I'm going to try something.

Issue #66, opened 2016-09-14
rust-lang/rust integration metabug


japaric:
- [x] Build missing intrinsics from compiler-rt (C) source code #63
- [x] Put weak memcpy et al symbols behind an opt-in Cargo feature #64
- [x] Test our implementations against libgcc #65 
- [x] Rename this repo to compiler-builtins
- [x] Transfer to rust-lang-nursery
- [x] compile functions with hidden visibility
- [x] i128 support - #133 
- [x] "C" ABI -> "aapcs" on ARM - #116 #141

cc @alexcrichton

est31:
Note, now that https://github.com/rust-lang/rust/pull/38482 is merged, i128 support is required as well.

Just using the C implementation is not enough as the C implementation only is enabled and works on non windows 64 bit platforms (that was the reason the PR had to implement the intrinsics in Rust in the first place).

See #133 that attempts to add support.

Also note that during the making of that PR I not just had linker issues with calls to panic, it also complained about calls to eh_personality, which in some cases survives while panic does not.

japaric:
Added i128 support to the list. After #133 lands, we would still need the famility of floatundif intrinsics, I believe.

Also added changing the ABI of (some?) intrinsics from C to aapcs on ARM. cc #116 

> compile functions with hidden visibility

And I believe this issue was fixed in rust-lang/rust proper? cc @alexcrichton

est31:
With #133 merged, the following float functions are required to be ported:

* `__fixsfti`
* `__fixdfti`
* `__fixunsdfti`
* `__fixunssfti`
* `__floattidf`
* `__floattisf`
* `__floatuntidf`
* `__floatuntisf`

They all need `extern "unadjusted"` instead of `extern "C"` on windows.

japaric:
>  "C" ABI -> "aapcs" on ARM - #116 #141

I think that powisf2 and powidf2 may need to be changed to AAPCS too. Honestly, I'm no longer sure what's the state of those intrinsics on LLVM 3.9 vs 4.0.

japaric:
OK. I believe that all the required intrinsics to land this in rust-lang/rust have been implemented. All that's left is: (a) check that the compiler-rt submodule in this repo matches the revision being used in rust-lang/rust and then (b) fight with bors and land this in rust-lang/rust.

Any volunteers? Here's what needs to be done for (b):

- Remove the `src/libcompiler_builtins` directory in rust-lang/rust
[...]

PR #74, opened 2016-09-26
add an opt-in cargo feature to build intrinsics from compiler-rt source

### LINKED ISSUE #63: Build missing intrinsics from compiler-rt (C) source code
- [ ] Add compiler-rt source code as a submodule. Use rust-lang/compiler-rt
- [ ] Add an opt-in Cargo feature for this. We'll only compile the intrinsics from C code if the feature is enabled.

japaric:
closes #63
cc #66 

This uses the build.rs version from rust-lang/rust#36512 minus the dependency on llvm-target being in `rustc --print cfg`

cc @alexcrichton I forgot to ask before but how do you feel about using the `rustc-cfg` crate, which parses the output of `rustc --print cfg` and exposes the target specification as a Rust `struct`, in the build.rs? Its functionality overlaps with what you proposed in rust-lang/rfcs#1721 i.e. CARGO_CFG_TARGET_OS, etc.

TODO stop compiling the C version of the intrinsics that we have already ported to Rust.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,8 +1,430 @@
+extern crate gcc;
+extern crate rustc_cfg;
+
+use std::collections::BTreeMap;
 use std::env;
+use std::path::Path;
+
+use rustc_cfg::Cfg;
+
+struct Sources {
+    // SYMBOL -> PATH TO SOURCE
+    map: BTreeMap<&'static str, &'static str>,
+}
+
+impl Sources {
+    fn new() -> Sources {
+        Sources { map: BTreeMap::new() }
+    }
+
+    fn extend(&mut self, sources: &[&'static str]) {
+        // NOTE Some intrinsics have both a generic implementation (e.g.
+        // `floatdidf.c`) and an arch optimized implementation
+        // (`x86_64/floatdidf.c`). In those cases, we keep the arch optimized
+        // implementation and discard the generic implementation. If we don't
+        // and keep both implementations, the linker will yell at us about
+        // duplicate symbols!
+        for &src in sources {
+            let symbol = Path::new(src).file_stem().unwrap().to_str().unwrap();
+            if src.contains("/") {
+                // Arch-optimized implementation (preferred)
+                self.map.insert(symbol, src);
+            } else {
+                // Generic implementation
+                if !self.map.contains_key(symbol) {
+                    self.map.insert(symbol, src);
+                }
+            }
+        }
+    }
+
+    fn remove(&mut self, symbols: &[&str]) {
+        for symbol in symbols {
+            self.map.remove(*symbol).unwrap();
+        }
+    }
+}
 
 fn main() {
-    if env::var("TARGET").unwrap().ends_with("gnueabihf") {
+    println!("cargo:rerun-if-changed=build.rs");
+
+    let target = env::var("TARGET").unwrap();
+    let Cfg { ref target_arch, ref target_os, ref target_env, ref target_vendor, .. } =
+        Cfg::new(&target).unwrap();
+    // NOTE we are going to assume that llvm-target, what determines our codegen option, matches the
+    // target triple. This is usually correct for our built-in targets but can break in presence of
+    // custom targets, which can have arbitrary names.
+    let llvm_target = target.split('-').collect::<Vec<_>>();
+    let target_vendor = target_vendor.as_ref().unwrap();
+
+    // Build missing intrinsics from compiler-rt C source code
+    if env::var_os("CARGO_FEATURE_C").is_some() {
+        let cfg = &mut gcc::Config::new();
+
+        if target_env == "msvc" {
+            // Don't pull in extra libraries on MSVC
+            cfg.flag("/Zl");
+
+            // Emulate C99 and C++11's __func__ for MSVC prior to 2013 CTP
+            cfg.define("__func__", Some("__FUNCTION__"));
+        } else {
+            // Turn off various features of gcc and such, mostly copying
+            // compiler-rt's build system already
+            cfg.flag("-fno-builtin");
+            cfg.flag("-fvisibility=hidden");
+            cfg.flag("-fomit-frame-pointer");
+            cfg.flag("-ffreestanding");
+        }
+
+        // NOTE Most of the ARM intrinsics are written in assembly. Tell gcc which arch we are going to
+        // target to make sure that the assembly implementations really work for the target. If the
+        // implementation is not valid for the arch, then gcc will error when compiling it.
+        if llvm_target[0].starts_with("thumb") {
+            cfg.flag("-mthumb");
+
+            if llvm_target.last() == Some(&"eabihf") {
+                cfg.flag("-mfloat-abi=hard");
+            }
+        }
+
+        if llvm_target[0] == "thumbv6m" {
+            cfg.flag("-march=armv6-m");
+        }
+
+        if llvm_target[0] == "thumbv7m" {
+            cfg.flag("-march=armv7-m");
+        }
+
+        if llvm_target[0] == "thumbv7em" {
+            cfg.flag("-march=armv7e-m");
+        }
+
+        let mut sources = Sources::new();
+        sources.extend(&["absvdi2.c",
+                         "absvsi2.c",
+                         "addvdi3.c",
+                         "addvsi3.c",
+                         "apple_versioning.c",
+                         "clear_cache.c",
+                         "clzdi2.c",
+                         "clzsi2.c",
+                         "cmpdi2.c",
+                         "comparedf2.c",
+                         "comparesf2.c",
+                         "ctzdi2.c",
+                         "ctzsi2.c",
+                         "divdc3.c",
+                         "divdf3.c",
+                         "divsc3.c",
+                         "divsf3.c",
+                         "divxc3.c",
+                         "extendsfdf2.c",
+                         "extendhfsf2.c",
+                         "ffsdi2.c",
+                         "fixdfdi.c",
+                         "fixdfsi.c",
+                         "fixsfdi.c",
+                         "fixsfsi.c",
+                         "fixunsdfdi.c",
+                         "fixunsdfsi.c",
+                         "fixunssfdi.c",
+                         "fixunssfsi.c",
+                         "fixunsxfdi.c",
+                         "fixunsxfsi.c",
+                         "fixxfdi.c",
+                         "floatdidf.c",
+                         "floatdisf.c",
+                         "floatdixf.c",
+                         "floatsidf.c",
+                         "floatsisf.c",
+                         "floatundidf.c",
+                         "floatundisf.c",
+                         "floatundixf.c",
+                         "floatunsidf.c",
+                         "floatunsisf.c",
+                         "int_util.c",
+                         "muldc3.c",
+                         "muldf3.c",
+                         "muloti4.c",
+                         "mulsc3.c",
+                         "mulsf3.c",
+                         "mulvdi3.c",
+                         "mulvsi3.c",
+                         "mulxc3.c",
+                         "negdf2.c",
+                         "negdi2.c",
+                         "negsf2.c",
+                         "negvdi2.c",
+                         "negvsi2.c",
+                         "paritydi2.c",
+                         "paritysi2.c",
+                         "popcountdi2.c",
+                         "popcountsi2.c",
+                         "powidf2.c",
+                         "powisf2.c",
+                         "powixf2.c",
+                         "subdf3.c",
+                         "subsf3.c",
+                         "subvdi3.c",
+                         "subvsi3.c",
+                         "truncdfhf2.c",
+                         "truncdfsf2.c",
+                         "truncsfhf2.c",
+                         "ucmpdi2.c"]);
+
+        if target_os != "ios" {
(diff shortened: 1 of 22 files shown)


japaric:
@homunkulus try

homunkulus:
:hourglass: Trying commit 310aaa9 with merge 310aaa9...

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/rustc-builtins/builds/162905081)

alexcrichton:
@japaric I'd personally prefer to keep the dependencies of the rust-lang/rust repo as slim as possible, and in that sense calling `rustc --print cfg` manually in theory shouldn't be too hard (including parsing). Otherwise using it to inform codegen seems good to me!

japaric:
> I'd personally prefer to keep the dependencies of the rust-lang/rust repo as slim as possible

does that include dev dependencies? This repo uses quickcheck for its tests.

japaric:
Also, `rustc-cfg` has zero dependencies :wink:.

japaric:
@homunkulus try

homunkulus:
:hourglass: Trying commit 82bf52c with merge 82bf52c...

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/rustc-builtins/builds/162968002)

japaric:
@alexcrichton having that intrinsics.rs test in this repository has helped me a lot in debugging issues with missing/undefined symbols. Thanks to it I just learned that the `thumbv6m-none-eabi` target doesn't support some floating point operations because the related compiler-rt intrinsics, which are written in assembly, can't be "compiled" for that target. (cf #75)

I'm actually wondering if it's actually necessary to also run the intrinsics.rs test on the rust-lang/rust repo. Seems like all intrinsics related changes would have to first go through this repository and the CI + intrinsics.rs in this repo should catch any problem as long as we are testing all tier 1 targets (I think I'm only missing android atm) and the intrinsics.rs file covers all the relevant operations (right now, it covers ~70 intrinsics).

japaric:
@homunkulus try

homunkulus:
:hourglass: Trying commit ee05eb4 with merge ee05eb4...

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/rustc-builtins/builds/162974332)

japaric:
@homunkulus try

OK. This should be ready to go now. Who wants to review? @Amanieu @mattico 

The build.rs addition is pretty much a copy paste of the one from rust-lang/rust#36512 with additional changes to support the thumb* targets. I should have done that part in two commits so the additional changes were more obvious, sorry; but all the additions should have a "thumb" string near to it :-).

homunkulus:
:hourglass: Trying commit 7d49d99 with merge 7d49d99...

mattico:
I'll take a look.

homunkulus:
:broken_heart: Test failed - [status-appveyor](https://ci.appveyor.com/project/japaric/rustc-builtins/build/1.0.253)

mattico:
Looks like we ran into the gist rate limit on the CI, should be pretty easy to fix.

japaric:
Oh right, I have to update the AppVeyor CI to test these changes. I'll probably have to use *ugh* powershell so I'm going to leave that for a follow-up PR.

ci/script.sh
```diff
@@ -23,9 +18,42 @@ inspect() {
...
+        cargo test --target $TARGET
+        cargo test --target $TARGET --release
+    fi
+}
+
+c_test() {
+    $CARGO clean
+
+    # verify we haven't drop any intrinsic/symbol
+    $CARGO build --features c --target $TARGET --bin intrinsics
```
mattico: Would it make sense to have `RUSTFLAGS="-C link-args='-Wl,-nostartfiles -Wl,-nodefaultlibs'"` so that it applies to all targets? I'm assuming that we don't need libc anywhere in there...

thumbv6m-none-eabi.json
```diff
@@ -1,10 +1,12 @@
 {
     "arch": "arm",
     "data-layout": "e-m:e-p:32:32-i64:64-v128:64:128-a:0:32-n32-S64",
+    "executables": true,
     "features": "+strict-align",
     "llvm-target": "thumbv6m-none-eabi",
     "max-atomic-width": 0,
     "os": "none",
+    "pre-link-args": ["-nostartfiles"],
```
mattico: If so, we wouldn't need these anymore.

ci/script.sh
japaric: Only -nostartfiles is needed on Linux. I think that would work on Linux, but idk about OSX and Windows. I'm using the libc crate because, on each platform, it links to different libraries needed to produce an executable. Without it the executable may not link at all.

mattico: Sounds good 👍. In any case it doesn't make much of a difference.

japaric:
@homunkulus try

This should be ready to go.

@alexcrichton Mind taking a look at the changes in the testing infra? Also, are you OK with adding rust-cfg here or would you prefer to just use $TARGET and `.contains()`?

japaric:
@homunkulus try

homunkulus:
:hourglass: Trying commit b5797dc with merge b5797dc...

japaric:
@homunkulus try

homunkulus:
:hourglass: Trying commit 62258b5 with merge 62258b5...

src/int/mul.rs
```diff
@@ -66,7 +66,9 @@ macro_rules! mulo {
     }
 }
 
+#[cfg(not(all(feature = "c", target_arch = "x86")))]
```
alexcrichton: What's up w/ these `#[cfg]`?

alexcrichton: Oh I see the comment below now.

build.rs
```diff
@@ -1,8 +1,430 @@
...
+    fn remove(&mut self, symbols: &[&str]) {
+        for symbol in symbols {
+            self.map.remove(*symbol).unwrap();
+        }
+    }
+}
 
 fn main() {
-    if env::var("TARGET").unwrap().ends_with("gnueabihf") {
+    println!("cargo:rerun-if-changed=build.rs");
```
alexcrichton: If this is compiling a bunch of C files then this'll probably also want to print a similar key for those files as well.

build.rs
```diff
@@ -1,8 +1,430 @@
...
+            cfg.flag("-fno-builtin");
+            cfg.flag("-fvisibility=hidden");
+            cfg.flag("-fomit-frame-pointer");
+            cfg.flag("-ffreestanding");
+        }
+
+        // NOTE Most of the ARM intrinsics are written in assembly. Tell gcc which arch we are going to
+        // target to make sure that the assembly implementations really work for the target. If the
+        // implementation is not valid for the arch, then gcc will error when compiling it.
+        if llvm_target[0].starts_with("thumb") {
```
alexcrichton: Could this logic be added to gcc-rs? (it'll be needed for any C code, right?)

ci/run.sh
```diff
@@ -0,0 +1,64 @@
...
+# Look out for duplicated symbols when we include the compiler-rt (C) implementation
+PREFIX=$(echo $1 | sed -e 's/unknown-//')
+case $1 in
+    armv7-*)
+        PREFIX=arm-linux-gnueabihf-
+        ;;
+    thumb*)
+        PREFIX=arm-none-eabi-
+        ;;
+    *-unknown-linux-gnu | *-apple-darwin)
```
alexcrichton: this'll catch mips/mipsel/powerpc, right? (is that intended?)

alexcrichton:
> Mind taking a look at the changes in the testing infra? 

Sure!

> Also, are you OK with adding rust-cfg here or would you prefer to just use $TARGET and .contains()?

I think we'll end up adding this no matter what, so seems good to me!

src/int/mul.rs
japaric: There's a comment about this in src/lib.rs. We have generic implementations of some intrinsics written in Rust but for some of those intrinsics, compiler-rt has arch-optimized implementations written in assembly. We want to use compiler-rt arch optimized impls where available; in those cases, build.rs will compile them and these cfgs remove our Rust impl from the library so we don't end with duplicate symbols.

alexcrichton: 👍

build.rs
japaric: If it's fine to add logic for targets that are not built into the compiler then I can send a PR.

ci/run.sh
japaric: Ah no. I meant to only pick up x86-ish targets. I'll fix it.

build.rs
japaric: gcc-rs doesn't generate those for me?

japaric:
@alexcrichton I've addressed all your comments I think.

japaric:
Minus the gcc-rs bit because that will take time to land and we can always remove the gcc flags for thumb* later on

build.rs
alexcrichton: Ah yeah that's true, but it's looking like we'll add these soon so I wouldn't mind adding them.

ci/run.sh
```diff
@@ -0,0 +1,64 @@
...
+        ;;
+esac
+
+# Verify that we haven't drop any intrinsic/symbol
+case $1 in
+    thumb*)
+        xargo build --features c --target $1 --bin intrinsics
+        ;;
+    *)
+        cargo build --features c --target $1 --bin intrinsics
```
alexcrichton: Perhaps this build could also be tested as well?

src/bin/intrinsics.rs
```diff
@@ -0,0 +1,399 @@
...
+
+// NOTE cfg(not(thumbv6m)) means that the operation is not supported on ARMv6-M at all. Not even
+// compiler-rt provides a C/assembly implementation.
+
+// Every function in this module maps will be lowered to an intrinsic by LLVM, if the platform
+// doesn't have native support for the operation used in the function. ARM has a naming convention
+// convention for its intrinsics that's different from other architectures; that's why some function
+// have an additional comment: the function name is the ARM name for the intrinsic and the comment
+// in the non-ARM name for the intrinsic.
+#[cfg(feature = "c")]
```
alexcrichton: Some of these are defined even when feature=c is disabled, right?

ci/run.sh
japaric: `cargo test --features c` would end testing the same thing I think. We'll just be testing less intrinsics on some targets. Also, this would require sprinkling more cfg over the test modules.

alexcrichton:
Looking good to me!

The `#[cfg(not(all(feature = "c", ..)))]` declarations look like they may get a bit unwieldy though in terms of management, but perhaps that can just be dealt with as it happens.

ci/run.sh
alexcrichton: Ah, carry on then!

src/bin/intrinsics.rs
japaric: Want to test those in the cfg(not(feature = "c")) main?

japaric:
> The #[cfg(not(all(feature = "c", ..)))] declarations look like they may get a bit unwieldy though in terms of management,

Let's get rid of the C code fast so we can soon get rid of all these cfgs and the C feature as well!

src/bin/intrinsics.rs
alexcrichton: In theory yeah, but it's fine to land this ahead of time.

alexcrichton:
> Let's get rid of the C code fast so we can soon get rid of all these cfgs and the C feature as well!

Agreed!

japaric:
@homunkulus r+

Alright, let's land this. Then I can try landing this in rust-lang/rust, which is probably going to be awful

homunkulus:
:pushpin: Commit 61484ac has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 61484ac with merge 61484ac...

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/rustc-builtins/builds/163895424)

japaric:
Did https://static.rust-lang.org/rustup.sh break? All the `language: rust` builds are failing.

alexcrichton:
I [think](https://github.com/rust-lang/rust/issues/36845)? But [claimed to be fixed](https://github.com/rust-lang/rust/issues/36845#issuecomment-250636034).

Maybe it's not though?

japaric:
@homunkulus retry

homunkulus:
:hourglass: Testing commit 61484ac with merge 61484ac...

japaric:
@homunkulus r+

homunkulus:
:pushpin: Commit cab88e6 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit cab88e6 with merge cab88e6...

homunkulus:
:sunny: Test successful - [status-appveyor](https://ci.appveyor.com/project/japaric/rustc-builtins/build/1.0.260), [status-travis](https://travis-ci.org/japaric/rustc-builtins/builds/164117381)
Approved by: japaric
Pushing cab88e6133b0db9c6663ffd8b2f65cb35e8a9dda to master...

Issue #72, opened 2016-09-22
Better test harness


mattico:
I'd like to have results like:

```
target: thumbv6m-none-eabi

Testing 123 intrinsics...
test muldi ... ok
test mulosi ... ok
test mulodi ... ok
test divdi3 ... ERROR
  input: (-1, INT_MAX)
  results:
    compiler-rt:       32768
    gcc_s:             32768
    compiler-builtins: 32767
test udivmoddi4 ... ERROR
  input: (25, 367)
  results:
    compiler-rt:       3
    gcc_s:             N/A
    compiler-builtins: 365
...
```

```
target: thumbv6m-none-eabi

Benchmarking 123 intrinsics...
bench muldi...
  compiler-rt:       16,386 ns/iter (+- 366)
  gcc_s:             14,375 ns/iter (+- 540)
  compiler-builtins: 17,690 ns/iter (+- 750)
bench mulosi...
...
```

PR #76, opened 2016-09-27
Prep for and expand test infrastructure

alexcrichton:
This commit implements a few key features:

* All tests are now executed against compiler-rt in addition to gcc_s. This is a reference implementation that will be guaranteed to exist so we can ensure that tests always run.
* All test infrastructure now lives on the master branch of this repository, allowing it to be version controlled just as the code is. This has the benefits of being able to easily add a new platform, modify an existing platform, or reproduce what a platform is doing locally.

I've gotten a few tests to pass locally, but I suspect this isn't *quite* ready for merging yet. I'll try to fixup anything in the morning, but figured I'd get some thoughts!


src/qc.rs
```diff
@@ -142,3 +142,123 @@ macro_rules! arbitrary_large {
...
+//            let mut other_result = 0;
+//            let result = f(a.0, &mut other_result);
+//            Some((result, other_result))
+//        }
+//    }
+//
+// If anything returns `None` then the test is discarded, otherwise the two
+// results are compared for equality and the test fails if this equality check
+// fails.
+macro_rules! check {
```
alexcrichton: This was pretty heinous to write, so I can try to document it a bit more if you'd like or tweak the interface. It was just the first thing I came up with!

tests/inspect.rs
```diff
@@ -0,0 +1,31 @@
...
+use std::path::Path;
+
+fn nm(path: &Path) -> Command {
+    let mut nm = Command::new("nm");
+    nm.arg("-g").arg("--defined-only").arg(path);
+    return nm
+}
+
+#[test]
+fn inspect() {
```
alexcrichton: I'm not sure whether it'd still be worth keeping this around, but curious for your thoughts!

ci/script.sh
```diff
@@ -1,80 +0,0 @@
...
-
-    set +e
-    $PREFIX$OBJDUMP -Cd target/**/release/*.rlib | gist_it
-    set -e
-
-    # Check presence/absence of weak symbols
-    if [[ $WEAK ]]; then
-        local symbols=( memcmp memcpy memmove memset )
-        for symbol in "${symbols[@]}"; do
-            $PREFIX$NM target/$TARGET/debug/deps/librlibc-*.rlib | grep -q "W $symbol"
```
alexcrichton: I haven't re-added this test, but I figure that it's something that should be tested in rlibc rather than here.

tests/inspect.rs
japaric: The `nm` output in this form is not too important so I'd be OK with dropping this test in this PR.

src/qc.rs
japaric: Criminally beatiful.

src/qc.rs
```diff
@@ -142,3 +142,123 @@ macro_rules! arbitrary_large {
...
+                fn $name() {
+                    fn my_check($($arg:$t),*) -> TestResult {
+                        let my_answer = super::$name(super::super::$name,
+                                                     $($arg),*);
+                        let gcc_s_fn = match ::gcc_s::get(stringify!($name)) {
+                            Some(f) => f,
+                            // If it's not in libgcc, or we couldn't find
+                            // libgcc, then just ignore this. We should have
+                            // tests through compiler-rt in any case
+                            None => return TestResult::passed(),
```
japaric: could the logic be changed to avoid calling quickcheck::quickcheck at all if the gcc_s symbol is not available?

ci/docker/powerpc64-unknown-linux-gnu/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu:16.04
+
+RUN apt-get update
+RUN apt-get install -y --no-install-recommends \
+        gcc libc6-dev qemu-user ca-certificates \
+        gcc-powerpc64-linux-gnu libc6-dev-ppc64-cross \
+        qemu-system-ppc
+
+ENV CARGO_TARGET_POWERPC64_UNKNOWN_LINUX_GNU_LINKER=powerpc64-linux-gnu-gcc \
+    CC=powerpc64-linux-gnu-gcc \
```
japaric: Could we use the suffixed version (CC_powerpc64_...) here? My understanding is that CC also overrides the host compiler and, although we aren't using it here, I think it's better to avoid surprises in the future.

Some with the other Dockerfiles.

.travis.yml
```diff
@@ -1,62 +1,59 @@
...
-      os: linux
-    - env: TARGET=thumbv7em-none-eabi
-      os: linux
-    - env: TARGET=thumbv7em-none-eabi WEAK=true
-      os: linux
-    - env: TARGET=thumbv7em-none-eabihf
-      os: linux
-    - env: TARGET=thumbv7em-none-eabihf WEAK=true
-      os: linux
+      install: cargo install xargo
```
japaric: Installing Xargo takes a while. May be worth to (a) set cache: cargo in .travis.yml or (b) use one of the binary releases available via GitHub releases if it slows down the build job considerable.

Also cargo install Xargo requires a few extra packages: libcurl-dev libssl-dev cmake and maybe other stuff. If this command fails, then you can try installing those dependencies.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "compiler-rt/compiler-rt-cdylib/compiler-rt"]
+	path = compiler-rt/compiler-rt-cdylib/compiler-rt
+	url = https://github.com/llvm-mirror/compiler-rt
```
alexcrichton: Ah yeah all our "patches" now are just makefile related I think. I have a feeling getting all tests working here will cause problems though!

ci/docker/powerpc64-unknown-linux-gnu/Dockerfile
alexcrichton: Sure!

src/qc.rs
mattico: This is pretty much what I had in mind for #72. From the looks of things I'm glad I didn't have to write it 😆.

@alexcrichton do you have any thoughts about benchmarks? I'm assuming the starting performance target is "not significantly worse", which shouldn't be too bad since most of the code is pretty direct translation. I still think some basic benchmarks would be a good sanity check just to ward off possible regressions. I don't know how big the overhead is for dynamic function lookup but we could do that for `rustc-builtins` as well to make it more 'fair', I suppose.

alexcrichton: @mattico that's a good idea! I suspect that this'd be relatively easy to tweak to run benchmarks as well.

alexcrichton:
Hm ok, down to a few failures:

* i686 Linux/OSX - This is https://github.com/rust-lang/rust/issues/36793
* mips-unknown-linux-gnu - SIGTRAP if binary is executed manually, but succeeds if binary is executed with `qemu-mips{,el}`. Not sure why
* powerpc64-unknown-linux-gnu - not sure what's going on here

The "use qemu to emulate binaries we exec" support seems pretty sketchy. It's *super* nice to have but I can't get it to work...

@japaric any ideas on mips/powerpc64?

japaric:
>  SIGTRAP if binary is executed manually, but succeeds if binary is executed with qemu-mips{,el}. Not sure why

Sounds like the wrong binfmt interpreter was installed. 

> powerpc64-unknown-linux-gnu - not sure what's going on here

I think it's the same issue: bad interpreter or perhaps, more likely, the interpreter was not properly registered.

Dou you see the same errors locally, and are you also using the multiarch/qemu-user-static trick locally?

Perhaps the [register.sh] script in multiarch/qemu-user-static has some errors? I've never seen binfmt related errors like these when letting the binfmt-support/qemu-user-static packages handle the installation of binfmt interpreters (cf smoke). :-/

[register.sh]: https://github.com/multiarch/qemu-user-static/blob/master/register/register.sh#L26

alexcrichton:
Success! I've [finally got everything working](https://travis-ci.org/japaric/rustc-builtins/builds/163609319). I don't really know how, but at this point I'm not gonna ask...

I've squashed everything into one commit, and restored only testing on the auto/try branch.

r? @japaric

alexcrichton:
Oh so there's a few nightly bugs the travis file also works around. The 32-bit cross-compilation will be fixed tonight but for now I rolled back those nightlies, and there's also a recent bug with ARM and float intrinsics (I think you found this?) that I fixed by rolling back a few nights as well. Fixes are in the pipeline on rust-lang/rust so we should be able to go back to the nightly channel in short order.

japaric:
> Success! I've finally got everything working.

An herculean task! Thank you @alexcrichton 

> I fixed by rolling back a few nights as well.
> we should be able to go back to the nightly channel in short order.

:+1: 

@homunkulus r+

homunkulus:
:pushpin: Commit 8e161a7 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 8e161a7 with merge 8e161a7...

homunkulus:
:broken_heart: Test failed - [status-appveyor](https://ci.appveyor.com/project/japaric/rustc-builtins/build/1.0.255)

alexcrichton:
Aww :(

Want me to remove the gist part from appveyor as well?

japaric:
> Want me to remove the gist part from appveyor as well?

*sigh* I'll remove it.

The tests passed so I'm going to merge manually.

PR #80, opened 2016-09-30
panic! -> abort

japaric:
closes #79

PR #94, opened 2016-10-06
Upstream

japaric:
Some changes required to make this repo compatible with the rust-lang/rust repo. Most changes don't cause problems except for the new explicit `core` dependency which has a relative path (`../libcore`) that breaks standalone builds of this crate :disappointed:.

cc @alexcrichton Even though I marked the `core` dependency as optional with the hope that Cargo won't even look at the path which doesn't exist, it still does.


alexcrichton:
Perhaps we could take the route of libc? That way we'd leave the structure as-is here, but another `Cargo.toml` would be checked in to the rust-lang/rust repo (e.g. we wouldn't even use this repo's Cargo.toml during a bootstrap)

This is the `libc_shim` directory -- https://github.com/rust-lang/rust/tree/master/src/rustc/libc_shim

homunkulus:
:umbrella: The latest upstream changes (presumably #80) made this pull request unmergeable. Please resolve the merge conflicts.

japaric:
Closing in favor of #97

PR #98, opened 2016-10-12
Correct the license to that of upstream compiler-rt

brson:
As this is a derivative work we must maintain the same license terms.

This changes the license to NCSA or MIT instead of Apache 2 or MIT. Since @japaric, @mattico, and @Amanieu have not contributed their work under NCSA they need to sign off on this PR before merging.

Alternately we can simply drop Apache 2 and continue with solely MIT, but that restricts any contributions back upstream, which are unlikely, but could theoretically happen. Better to be compatible if we can.


mattico:
👍 I meant to bring this up earlier.

I agree to relicense my contributions under the NCSA license with no other restrictions or conditions.

japaric:
I agree to relicense my contributions under the MIT+NCSA license with no other restrictions or conditions.

Amanieu:
I agree to relicense my contributions under the MIT+NCSA license with no other restrictions or conditions.

japaric:
@alexcrichton also needs to sign off

alexcrichton:
I agree to relicense my contributions under the MIT+NCSA license with no other restrictions or conditions.

japaric:
Thank you all!

PR #102, opened 2016-10-14
Update webhook url

alexcrichton:


alexcrichton:
😲

japaric:
The OSX builds took quite a while ...

alexcrichton:
https://www.traviscistatus.com/

>  Update - We are currently estimating that we will have processed the backlog for OS X builds on open source repositories in approximately 8 hours. Take note, however, that it might be faster than that since there are fewer builds at this time. We will continue to keep you informed about the situation and update the ETA accordingly. 

(travis problems with OSX today I think)

PR #103, opened 2016-10-14
Update CI configs

alexcrichton:
* Run on the master branch (to get previews)
* Pass --target on AppVeyor (oops)
* Use the current nightly for arm (I think that bug is fixed)


japaric:
> Pass --target on AppVeyor (oops)

We are not cross compiling though? Or rather the rustc host = %TARGET% so native compilation is doing the right thing. I agree that passing --target is more robust though.

@bors r+

alexcrichton:
Oh, right, carry on!

PR #105, opened 2016-10-15
test the mips64 targets

japaric:
These should be all the required changes I think but libloading can't be cross compiled for these new targets:

```
error: failed to run custom build command for `libloading v0.3.1`
process didn't exit successfully: `/target/debug/build/libloading-07147fe7fffab0b2/build-script-build`(exit code: 101)
--- stderr
thread 'main' panicked at 'could not get target info: TargetNotFound', ../src/libcore/result.rs:799
```

I've openend nagisa/rust_libloading#19 to track that. Wel'll have to wait until that's fixed.

cc @alexcrichton

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ matrix:
       os: osx
     - env: TARGET=i686-unknown-linux-gnu
     - env: TARGET=mips-unknown-linux-gnu
+    - env: TARGET=mips64-unknown-linux-gnuabi64
+    - env: TARGET=mips64el-unknown-linux-gnuabi64
     - env: TARGET=mipsel-unknown-linux-gnu
     - env: TARGET=powerpc-unknown-linux-gnu
     - env: TARGET=powerpc64-unknown-linux-gnu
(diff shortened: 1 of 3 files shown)


japaric:
More targets became available today but I don't think it makes much sense to test the arm-musl targets because we are already testing the arm-gnu ones and we'd be testing the same arch again.

It'd be nice to test the s390x target but QEMU crashes when running s390x binaries like it did with the ppc64le target and we can't apply the QEMU_CPU trick because there are no CPUs to choose from, in the case of qemu-s390x.

alexcrichton:
r+ from me!

japaric:
```
error[E0514]: found crate `std` compiled by an incompatible version of rustc

  |

  = help: please recompile that crate using this compiler (rustc 1.14.0-nightly (98a3502da 2016-10-15))

  = note: crate `std` path #1: /rust/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-6eb85298.rlib compiled by "rustc 1.14.0-nightly (6dc035ed9 2016-10-15)"

  = note: crate `std` path #2: /rust/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-6eb85298.so compiled by "rustc 1.14.0-nightly (6dc035ed9 2016-10-15)"

error: aborting due to previous error
```

I think this was just a rustup hiccup (it installed the wrong std?)

@bors retry

Issue #90, opened 2016-10-03
Floating point add gives different results than compiler-rt on arm-unknown-linux-gnueabi


mattico:
I'm still trying to figure out what was wrong with #48, and I'd like another pair of eyes on this because I have no idea what is going on.

I added an implementation of `arbitrary_float!` again, and [ran the tests](https://travis-ci.org/mattico/rustc-builtins/builds/164723455). Only [the arm-unknown-linux-gnueabi target](https://travis-ci.org/mattico/rustc-builtins/jobs/164723458#L811) failed. So I grab the arguments that made the test fail and [put them into a regular test](https://github.com/mattico/rustc-builtins/commit/1dd0832a952d2f80ef273cb819b106f1553bc431). On the next build, [the test passes](https://travis-ci.org/mattico/rustc-builtins/jobs/164727780#L784) because we're actually getting the [right answer](https://www.wolframalpha.com/input/?i=0.000000000000000000000000000000000000011540496-0.00000000000000000000000000000000000000359621) in this case. It's gcc_s and compiler-rt that have the wrong answer.

Either:
1. There's some subtle bug in the test infrastructure
2. gcc_s and libcompiler-rt both have the same bug but we don't despite the code being almost identical to compiler-rt
3. QEMU is broken but only in this specific way
4. The C compiler's optimizations cause rounding in very small numbers, but rustc does not
5. Something else?

mattico:
New theory: compiler-rt and gcc_s are both compiled with optimizations, we are not (during tests). The optimizer moves/combines some operations which leads to subtle rounding differences in very small numbers.

alexcrichton:
Oh that's actually a really good point that we should always compile compiler-rt with optimizations. Right now compiler-rt is conditionally optimized, but gcc_s is pulled in from the system so I think that's always optimize.

japaric:
> The C compiler's optimizations cause rounding in very small numbers, but rustc does not

Do the errors go away if you test with optimizations enabled (`cargo test --release`)?

Are these roundings, `-ffast-math` optimizations? Who's doing the rounding (less precision) C or Rust?

> we should always compile compiler-rt with optimizations.

This should be a simple change in the build.rs via gcc::Config (I expect)

But ... enabling optimizations shouldn't change the results of the compiler-rt intrinsics, right? That doesn't sound quite right. You should get the same output regardless of the optimization level.

mattico:
[...]

PR #106, opened 2016-10-15
Implement floatunsisf (WIP)

Phaiax:
PR for review and travis.
I will delete the comments if you say that the approach is correct.

I'll do fmul and f2uiz too, because my project is complaining^^

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -145,7 +145,6 @@ fn main() {
                          "floatundisf.c",
                          "floatundixf.c",
                          "floatunsidf.c",
-                         "floatunsisf.c",
                          "int_util.c",
                          "muldc3.c",
                          "muldf3.c",
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -62,7 +62,7 @@ case $TRAVIS_OS_NAME in
 esac
 
 # NOTE On i586, It's normal that the get_pc_thunk symbol appears several times so ignore it
-if [ $TRAVIS_OS_NAME = osx ]; then
+if [ "$TRAVIS_OS_NAME" = "osx" ]; then
     path=target/${1}/debug/libcompiler_builtins.rlib
 else
     path=/target/${1}/debug/libcompiler_builtins.rlib
--- a/compiler-rt/compiler-rt-cdylib/build.rs
+++ b/compiler-rt/compiler-rt-cdylib/build.rs
@@ -60,6 +60,7 @@ fn main() {
         "addsf3.c",
         "powidf2.c",
         "powisf2.c",
+        "floatunsisf.c",
     ]);
 
     for src in sources.files.iter() {
--- a/compiler-rt/compiler-rt-cdylib/src/lib.rs
+++ b/compiler-rt/compiler-rt-cdylib/src/lib.rs
@@ -24,6 +24,7 @@ extern {
     fn __adddf3();
     fn __powisf2();
     fn __powidf2();
+    fn __floatunsisf();
 }
 
 macro_rules! declare {
@@ -57,6 +58,7 @@ declare!(___addsf3, __addsf3);
 declare!(___adddf3, __adddf3);
 declare!(___powisf2, __powisf2);
 declare!(___powidf2, __powidf2);
+declare!(___floatunsisf, __floatunsisf);
 
 #[lang = "eh_personality"]
 fn eh_personality() {}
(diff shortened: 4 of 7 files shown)


Phaiax:
Another question:
compiler-rt defines an alias for many functions: `ARM_EABI_FNALIAS(ui2f, floatunsisf)`
[See #define](https://github.com/rust-lang/compiler-rt/blob/8598065bd965d9713bfafb6c1e766d63a7b17b89/lib/builtins/assembly.h#L149)
Because my project complained about the alias (`undefined reference to `__aeabi_ui2f'`) I guess the aliases are needed. 
```
#[inline(always)]
#[cfg_attr(not(test), no_mangle)]
pub extern "C" fn $alias(a: $ity) -> $fty {
    $intrinsic(a)
}
```
Like this? But I guess this will not work since LLVM expects a symbol. `asm!`?

Phaiax:
I managed to test locally and now I understand the testing system better :D

For the target `arm-unknown-linux-gnueabihf` I get confusing results. The compiler-builtins answer is correct but the other two seem confusing.

[Test resuls](https://gist.github.com/Phaiax/b5fca23829fc1e3ac1f5a3b2a092f833)
[libcompiler-rt disassembly](https://gist.github.com/Phaiax/f96a3c438310b7e0da2bef1a3c1f1d0f#file-libcompier-rt-arm-unknown-linux-gnueabihf-debug-disassembly-L2620)
[libcompiler_rt_cdylib.so dissassembly](https://gist.github.com/Phaiax/f32b8b353afa475ed5e5082cd9ae6c47#file-compiler_rt_cdylib-arm-unknown-linux-gnueabihf-s-L290)

--
Currently I guess there is something wrong with the calling convention. The result is arbitrary (but constant on repeated runs with the same source code)
```rust
    #[test]
    fn test_floatunsisf_libcompilerrt() {
        let compiler_rt_fn = ::compiler_rt::get("__floatunsisf");
        let compiler_rt_fn : extern fn(u32) -> f32 = unsafe { mem::transmute(compiler_rt_fn) };
        println!("1231515 {:?}", compiler_rt_fn);
        let ans = compiler_rt_fn(0);
        println!("{:?}", ans);
        assert!(ans.eq_repr(0.0f32));

    }
```
gives
```
---- float::conv::tests::test_floatunsisf_libcompilerrt stdout ----
	1231515 0xf650a5e9
0.0000000006592472
thread 'float::conv::tests::test_floatunsisf_libcompilerrt' panicked at 'assertion failed: ans.eq_repr(0.0f32)', src/float/conv.rs:90
```
And the same code without the `println!("123...` gives
```
---- float::conv::tests::test_floatunsisf_libcompilerrt stdout ----
	17867484000000000000000000
thread 'float::conv::tests::test_floatunsisf_libcompilerrt' panicked at 'assertion failed: ans.eq_repr(0.0f32)', src/float/conv.rs:91
```
`::compiler_rt::get("__aeabi_ui2f");` does not work because there is no symbol  `__aeabi_ui2f` in compiler_rt. (see second dissassembly). The results differ between release and debug mode but are still arbitrary.

I guess either the calling convention is wrong or the linker does something bad (it removes the __aeabi_ui2f symbol at least).

mattico:
@Phaiax those test failures seem spookily similar to #90

A calling convention problem seems plausible to me. I'll look into that.

mattico:
Also we haven't found a perfect way to do function aliases yet, so having a [wrapper function](https://github.com/rust-lang-nursery/compiler-builtins/blob/master/src/arm.rs#L61) is what we do for now.

japaric:
> For the target arm-unknown-linux-gnueabihf I get confusing results.

Ugh, this target keeps causing problems. Perhaps, we could conditionally *not* compile some intrinsics for this target until we figure out what the problem is. That way we can continue landing changes without being blocked by this issue. @alexcrichton what do you think?

Reminder for @mattico: please re-send PR #48 when you get the chance :smile:

mattico:
So on arm there are 3 different floating point ABIs:
- `soft`: Software floating point, passes results in integer registers (since there are no fp registers).
- `softfp`: Software *or* hardware floating point, but passes results in integer registers.
- `hard`: Hardware floating point, passes results in VFP registers.

It appears that libgcc is built for `softfp` if I understand it's build system (I don't). It seems possible that Rust is using `hard` and some random intermediate value left in a VFP register is used as the argument.

mattico:
I think all this PR needs is to be updated to use quickcheck and to disable tests on the failing targets. See:
https://github.com/mattico/compiler-builtins/blob/655f642d3f90d59ea008da9254566d4344ed8cba/src/float/add.rs#L192

If you're short on time I can update this PR once #111 lands.

japaric:
A more complete version of this PR landed in #147. Sorry that it took so long to get this done :-/.

PR #109, opened 2016-10-16
Fix a few clippy warnings

CryZe:


src/int/udiv.rs
```diff
@@ -191,49 +191,47 @@ pub extern "C" fn __udivmoddi4(n: u64, d: u64, rem: Option<&mut u64>) -> u64 {
...
-                    *rem = u64::from(n.low() & (d.low() - 1));
-                }
-
-                if d.low() == 1 {
-                    return n;
-                } else {
-                    let sr = d.low().trailing_zeros();
-                    return n >> sr;
-                };
+    } else if d.high() == 0 {
```
CryZe: This just collapses the `else { if {} else {} }` into `else if {} else {}`. Apparently GitHub is quite confused by this.

japaric:
@bors r+
Thank you, @CryZe!

PR #110, opened 2016-10-16
Add eh_unwind_resume lang item

CryZe:
This is necessary for targets where the `custom_unwind_resume` flag is set.


japaric:
The crate you are modifying is only used in tests. Are you testing against a custom target?

mattico:
This appears to be needed to run tests on Windows FWIW.

alexcrichton:
I believe this is no longer necessary since #166, so closing.

PR #111, opened 2016-10-16
Add float quickcheck

mattico:


src/float/add.rs
```diff
@@ -187,111 +187,28 @@ add!(__adddf3: f64);
...
         fn __addsf3(f: extern fn(f32, f32) -> f32,
-                    a: U32,
-                    b: U32)
-                    -> Option<FRepr<f32> > {
-            let (a, b) = (f32::from_repr(a.0), f32::from_repr(b.0));
-            Some(FRepr(f(a, b)))
+                    a: F32,
+                    b: F32)
+                    -> Option<F32> {
+            if option_env!("TARGET") != Some("arm-unknown-linux-gnueabi") {
```
japaric: This doesn't seem to be working; Travis is still failing this test.

Could you instead `#[cfg]` away the `addsf3` and `adddf3` symbols along with these too tests for the `arm-unknown-linux-gnueabi`? And please add a comment about the problem and point to the issue that's tracking it #90

src/qc.rs
```diff
@@ -143,6 +145,69 @@ macro_rules! arbitrary_large {
...
+        }
+
+        impl PartialEq for $TY {
+            fn eq(&self, other: &$TY) -> bool {
+                // NOTE(cfg) for some reason, on hard float targets, our implementation doesn't
+                // match the output of its gcc_s counterpart. Until we investigate further, we'll
+                // just avoid testing against gcc_s on those targets. Do note that our
+                // implementation matches the output of the FPU instruction on *hard* float targets
+                // and matches its gcc_s counterpart on *soft* float targets.
+                if cfg!(gnueabihf) {
```
japaric: Is this still necessary? There's now a `FRepr` struct wrapper for this logic.

mattico: This was meant to replace FRepr, but I forgot to do that when rebasing.

mattico:
I hope you don't have build failure emails turned on 😉.

japaric:
> I hope you don't have build failure emails turned on :wink:.

Oh, they are enabled but they aren't reaching my mailbox which probably means they are being sent to the rust-lang org mailbox :smile:.

src/float/add.rs
```diff
@@ -181,117 +181,30 @@ macro_rules! add {
     }
 }
 
+// NOTE(cfg) These are disabled for gnueabihf due to 
+// https://github.com/rust-lang-nursery/compiler-builtins/issues/90
+#[cfg(not(gnueabi))]
```
japaric: Right now, cfg(gnueabihf) works but cfg(gnueabi) doesn't. You'll have to update the build.rs to make Cargo pass this cfg when the target is arm-unknown-linux-gnueabi. See https://github.com/rust-lang-nursery/compiler-builtins/blob/master/build.rs#L428

mattico: 👍

mattico: So should the addsf3.c files be enabled for `arm-unknown-linux-gnueabi` since we're disabling the rust version?

japaric: yes

mattico:
Oops, I've been neglecting this. Will update.

japaric:
@bors r+

Thank you, @mattico

japaric:
@alexcrichton The tests seemed to pass (they are "green") but bors/homu didn't merge it.

alexcrichton:
@bors: retry force clean

Your guess is as good as mine...

japaric:
@alexcrichton Time out again. Maybe something got misconfigured (the webhook?).

alexcrichton:
Hm so recently I reclassified the appveyor repo to run under a different account, and I've been seeing these same timeout problems on the libc repo as well.

Unfortunately I have no idea why they're happening... Appveyor/Travis are green though, so want to merge?

japaric:
> Hm so recently I reclassified the appveyor repo to run under a different account, and I've been seeing these same timeout problems on the libc repo as well.

Perhaps you should try removing the existing appveyor webhooks (this repo > settings > webhooks) and re-adding the projects on appveyor.

> Appveyor/Travis are green though, so want to merge?

:+1:  from me.

alexcrichton:
Let's try that one more time...

@bors: retry

mattico:
Hm, looking at this now I think I should unify the handling of `*-gnueabi` and `*-gnueabihf` since (AFAIK) the error is the same in both cases. That is, for both targets:

- Continue to use the rust add implementation since it gives correct answers
- Disable tests for the add impl on `*-gnueabi*` since compiler-rt and gcc_s give incorrect answers sometimes (I still haven't figured out why).

If this PR gets merged before you see this, I'll just make a follow-up PR.

mattico:
This probably won't work but:
@bors r-

japaric:
I think bors is not listening to me either...

japaric:
but it seems bors got stuck again anyway so this won't get merged.

mattico:
Ok this is much better now. It's simpler, and now the powi tests will still run on gnueabi.

build.rs
```diff
@@ -424,8 +424,8 @@ fn main() {
     }
 
     // To filter away some flaky test (see src/float/add.rs for details)
-    if llvm_target.last() == Some(&"gnueabihf") {
-        println!("cargo:rustc-cfg=gnueabihf")
+    if llvm_target.last().unwrap().contains("gnueabi") {
+        println!("cargo:rustc-cfg=gnueabi")
```
japaric: Could you use two cfgs here, gnueabi and gnueabihf, instead of overloading the word "gnueabi" to mean both ABIs? I think it would be less error prone that way.

mattico: Hm. Good point. This also accidentally disables the tests on armv7-* which doesn't seem to have this issue.

build.rs
```diff
@@ -424,8 +424,9 @@ fn main() {
     }
 
     // To filter away some flaky test (see src/float/add.rs for details)
-    if llvm_target.last().unwrap().contains("gnueabi") {
-        println!("cargo:rustc-cfg=gnueabi")
+    if llvm_target[0].starts_with("arm") &&
+       llvm_target.last().unwrap().contains("gnueabi") {
+        println!("cargo:rustc-cfg=arm-linux")
```
mattico: @japaric is this better?

japaric:
@bors r+

Thanks, @mattico.

mattico:
:four_leaf_clover:

japaric:
Stuck again but I'm going to merge manually.

PR #115, opened 2016-10-31
Add atomic support for pre-ARMv6 on Linux

Amanieu:
This uses the [kernel user helpers](https://www.kernel.org/doc/Documentation/arm/kernel_user_helpers.txt) which are available starting from kernel 2.6.15. Since Rust currently requires 2.6.18 at a minimum, this should be fine in practice. I did not include support for 64-bit atomics since that functionality is only available in kernel 3.1.

This PR allows Rust to work on older ARM versions such as ARMv4 and ARMv5 with the full libstd.

### DIFF
--- a/src/arm_linux.rs
+++ b/src/arm_linux.rs
@@ -1,0 +1,170 @@
+use core::intrinsics;
+use core::mem;
+
+// Kernel-provided user-mode helper functions:
+// https://www.kernel.org/doc/Documentation/arm/kernel_user_helpers.txt
+unsafe fn __kuser_cmpxchg(oldval: u32, newval: u32, ptr: *mut u32) -> bool {
+    let out: u32;
+    // FIXME: we can't use BLX on ARMv4
+    asm!("blx ${0}"
+         : "={r0}" (out)
+         : "r" (0xffff0fc0u32)
+           "{r0}" (oldval),
+           "{r1}" (newval),
+           "{r2}" (ptr)
+         : "r3", "r12", "lr", "cc", "memory");
+    out == 0
+}
+unsafe fn __kuser_memory_barrier() {
+    // FIXME: we can't use BLX on ARMv4
+    asm!("blx ${0}"
+         :
+         : "r" (0xffff0fa0u32)
+         : "lr", "memory");
+}
+
+// Word-align a pointer
+fn align_ptr<T>(ptr: *mut T) -> *mut u32 {
+    // This gives us a mask of 0 when T == u32 since the pointer is already
+    // supposed to be aligned, which avoids any masking in that case.
+    let ptr_mask = 3 & (4 - mem::size_of::<T>());
+    (ptr as usize & !ptr_mask) as *mut u32
+}
+
+// Calculate the shift and mask of a value inside an aligned word
+fn get_shift_mask<T>(ptr: *mut T) -> (u32, u32) {
+    // Mask to get the low byte/halfword/word
+    let mask = match mem::size_of::<T>() {
+        1 => 0xff,
+        2 => 0xffff,
+        4 => 0xffffffff,
+        _ => unreachable!(),
+    };
+
+    // If we are on big-endian then we need to adjust the shift accordingly
+    let endian_adjust = if cfg!(target_endian = "little") {
+        0
+    } else {
+        4 - mem::size_of::<T>() as u32
+    };
+
+    // Shift to get the desired element in the word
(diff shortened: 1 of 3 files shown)


alexcrichton:
Perhaps this could be behind an off-by-default feature? For inclusion into rustc at least we'll want to turn off these intrinsics for now.

Amanieu:
I don't see why these intrinsics should be hidden behind a feature. They are never called by rustc on any of the currently available ARM targets.

alexcrichton:
The compiler likes to have very strict control over the symbols it exposes and provides, so if we don't need these symbols we wouldn't want them exported. *Because* they're not called is just motivation for not having them in the first place as part of the compiler build (e.g. an off-by-default feature)

Amanieu:
Actually `compiler-rt` already provides an implementation of these symbols, however that implementation only works on ARMv6. But if you insist, I will put these behind a feature flag.

Amanieu:
I moved the code behind the `arm_linux_atomic` feature.

atilag:
Hi team! Are there any further steps left before landing this patch?
Thanks!

nelsonjchen:
This pull request was brought up [here][1]. [The OP of this reddit had to revert to 1.13 to get things working on ARMv5TE since newer std requires atomic instructions or some similar service][2]. Would this patch allow compiling for ARMv5TE with the current std?

[1]: https://www.reddit.com/r/rust/comments/64j39d/crosscompiling_for_arm926ejs/
[2]: https://www.reddit.com/r/rust/comments/64j39d/crosscompiling_for_arm926ejs/dg4ck13/

mattico:
@nelsonjchen 
rustc still uses compiler-rt ([just renamed to compiler-builtins](https://github.com/rust-lang/rust/tree/master/src/libcompiler_builtins)) from LLVM. If you built rustc using this compiler-builtins instead, then I think it would work.

Amanieu:
@mattico Actually, the compiler-rt implementations of these intrinsics in compiler-rt uses ARMv6 instructions and won't work on ARMv4/ARMv5. The libgcc version does support ARMv4/ARMv5 though.

alexcrichton:
@Amanieu I think with discussion on https://github.com/rust-lang/rust/pull/44978 as of recently we're leaning towards merging this now? Would you be willing to rebase this on master?

Amanieu:
Rebased. Do we still want this behind the "arm_linux_atomic" feature? I think these should always be available (the symbols are already exported by libgcc/libcompiler-rt, so we aren't adding any new symbols).

alexcrichton:
Thanks! With the discussion on the linked PR I think we'd unconditionally provide these symbols on armv5 Linux, but nowhere else.

Amanieu:
Do I need to do anything in the PR or can I just leave it as a feature?

alexcrichton:
Yes let's have compilation or not be automatic, detected based on the target (no features required)

Amanieu:
Do we have any way to differentiate the different ARM targets? AFAIK they all return the same thing for `target_arch`/`target_os`/`target_env`.

alexcrichton:
We've got a build script!

Amanieu:
Done.

alexcrichton:
Looks great! I'll hold off on merging until https://github.com/rust-lang/rust/pull/44978 is decided on, but I think we're very likely to merging that. Thanks for the updates here!

src/arm_linux.rs
```diff
@@ -0,0 +1,170 @@
+use core::intrinsics;
+use core::mem;
+
+// Kernel-provided user-mode helper functions:
+// https://www.kernel.org/doc/Documentation/arm/kernel_user_helpers.txt
+unsafe fn __kuser_cmpxchg(oldval: u32, newval: u32, ptr: *mut u32) -> bool {
+    let out: u32;
+    // FIXME: we can't use BLX on ARMv4
+    asm!("blx ${0}"
```
parched: Just wondering, couldn't 0xffff0fc0u32 just be transmuted to a function pointer and then called instead of using inline assembly, then the compiler will take care of what branch instruction is used?

Amanieu: Using inline asm is actually more efficient in this case since we can precisely specify which registers are clobbered by the call. This allows the compiler to avoid unnecessary spills. The set of clobbered registers is guaranteed by the kernel ABI.

Now that you mention it though, for `__kuser_cmpxchg` the clobber list is pretty much identical to that of a normal function, so there probably isn't much benefit there...

PR #117, opened 2016-11-12
Fix apt 404s in Docker

mattico:
`apt update` and `apt install` should be in the same Docker RUN statement, otherwise `apt update` will be cached and `apt install` will 404 if a package no longer exists.

### DIFF
--- a/ci/docker/mips64-unknown-linux-gnuabi64/Dockerfile
+++ b/ci/docker/mips64-unknown-linux-gnuabi64/Dockerfile
@@ -7,7 +7,7 @@ RUN apt-get update && \
     libc6-dev \
     libc6-dev-mips64-cross \
     qemu-user-static \
-    apt-get install -y --no-install-recommends qemu-system-mips
+    qemu-system-mips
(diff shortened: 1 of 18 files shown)


japaric:
Ohh this makes sense. I have seen these 404 errors locally.

@bors r+

mattico:
Forgot a backslash, but the other two travis failures don't look relavent.

ci/docker/mips64-unknown-linux-gnuabi64/Dockerfile
```diff
@@ -1,13 +1,13 @@
...
+    apt-get install -y --no-install-recommends \
     ca-certificates \
     gcc \
     gcc-mips64-linux-gnuabi64 \
     libc6-dev \
     libc6-dev-mips64-cross \
-    qemu-user-static
-RUN apt-get install -y --no-install-recommends qemu-system-mips
+    qemu-user-static \
+    apt-get install -y --no-install-recommends qemu-system-mips
```
japaric: This needs a `&&` between the `qemu-user-static` and the `apt-get`.

japaric:
Tests passed. Merging.

PR #119, opened 2016-11-18
Define VISIBILITY_HIDDEN when compiling objects

alexcrichton:
Apparently compiler-rt passed this and we just forgot to. Fixes visibility of
some symbols on 32-bit Linux.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -78,6 +78,7 @@ fn main() {
             cfg.flag("-fvisibility=hidden");
             cfg.flag("-fomit-frame-pointer");
             cfg.flag("-ffreestanding");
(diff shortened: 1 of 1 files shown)


japaric:
Travis error:

>  Downloading https://static.rust-lang.org/dist/2016-11-18/rustc-nightly-src.tar.gz
> error: Couldn't fetch Rust source tarball

Was there no source release yesterday? (Xargo uses that tarball)

alexcrichton:
@bors: retry

Looks like today's nightlies have sources?

alexcrichton:
Still need to debug what's what with bors...

PR #120, opened 2016-11-23
CI: use Xargo 0.2.x

japaric:
this version uses the rust-src component that rustup installs instead
of fetching the source tarball from static.r-l.o, which sometimes went
wrong due to the fallible logic that Xargo 0.1.x used

r? @alexcrichton 

The issue is that we must now install rustup as well


japaric:
> which sometimes went wrong due to the fallible logic that Xargo 0.1.x used

This caused the error seen in #119

PR #124, opened 2016-12-10
by default, mark this crate as the #![compiler_builtins] crate

japaric:
as this is how most users will want to use this crate

r? @alexcrichton


alexcrichton:
Hm this isn't actually used as `#[cfg(feature = "rustbuild")]` in the source, is that intentional?

japaric:
Nah, it was wrong to begin with. I also need to update ci/run.sh to not use #![compiler_builtins] during tests. I'll update the PR once #125 lands.

PR #125, opened 2016-12-11
fix the test suite for the thumb targets

japaric:


japaric:
seems that the intrinsics test broke for the thumbv6m somewhere between 2016-11-22 and 2016-12-05. Bisecting ...

japaric:
Actually, it seems that rust-lang/rust#38281 broke the intrinsics test for all the targets.

japaric:
> seems that the intrinsics test broke for the thumbv6m somewhere between 2016-11-22 and 2016-12-05. 

I can no longer see this locally and travis isn't complaining so it may have been my mistake.

@bors r+

PR #128, opened 2016-12-18
add implementations of memcpy et al

### LINKED ISSUE #126: linker error when the "weak" feature is enabled
### STR

```
$ cargo new --bin app && cd $_
```

```
$ edit Cargo.toml && tail -n4 $_
[dependencies.compiler_builtins]
branch = "rustbuild"
features = ["weak"]
git = "https://github.com/japaric/compiler-builtins"
```

```
$ edit src/main.rs && cat $_
#![feature(compiler_builtins_lib)]
#![feature(core_intrinsics)]
#![feature(lang_items)]
#![no_main]
#![no_std]

extern crate compiler_builtins;

use core::intrinsics;

#[no_mangle]
pub fn _start() {
    unsafe {
        intrinsics::copy_nonoverlapping(0x0 as *const u32, 0x20 as *mut u32, 0x20);
    }
}

#[no_mangle]
pub fn __aeabi_unwind_cpp_pr0() {}

#[lang = "panic_fmt"]
extern "C" fn panic_fmt() {}
```

```
$ xargo rustc --target thumbv6m-none-eabi -- -C link-args=-nostartfiles
error: linking with `arm-none-eabi-gcc` failed: exit code: 1
  |
  = note: "arm-none-eabi-gcc" "-L" "/home/japaric/.xargo/lib/rustlib/thumbv6m-none-eabi/lib" "/home/japaric/tmp/app/target/thumbv6m-none-eabi/debug/deps/app-6d9087d3fe6b93f5.0.o" "-o" "/home/japaric/tmp/app/target/thumbv6m-none-eabi/debug/deps/app-6d9087d3fe6b93f5" "-Wl,--gc-sections" "-nodefaultlibs" "-L" "/home/japaric/tmp/app/target/thumbv6m-none-eabi/debug/deps" "-L" "/home/japaric/tmp/app/target/debug/deps" "-L" "/home/japaric/.xargo/lib/rustlib/thumbv6m-none-eabi/lib" "-Wl,-Bstatic" "-Wl,-Bdynamic" "/home/japaric/tmp/app/target/thumbv6m-none-eabi/debug/deps/librlibc-ca4e4855ef16b6a5.rlib" "/home/japaric/.xargo/lib/rustlib/thumbv6m-none-eabi/lib/libcore-539305ece2e53820.rlib" "/home/japaric/tmp/app/target/thumbv6m-none-eabi/debug/deps/libcompiler_builtins-71942e6a7f812dde.rlib" "-nostartfiles"
  = note: /home/japaric/tmp/app/target/thumbv6m-none-eabi/debug/deps/libcompiler_builtins-71942e6a7f812dde.rlib(compiler_builtins-71942e6a7f812dde.0.o): In function `compiler_builtins::arm::__aeabi_memcpy4':
/home/japaric/.cargo/git/checkouts/compiler-builtins-299ea7fa6d11e5d0/b59c789/src/arm.rs:117: undefined reference to `memcpy'
/usr/lib/gcc/arm-none-eabi/6.2.0/../../../../arm-none-eabi/bin/ld: /home/japaric/tmp/app/target/thumbv6m-none-eabi/debug/deps/app-6d9087d3fe6b93f5: hidden symbol `memcpy' isn't defined
/usr/lib/gcc/arm-none-eabi/6.2.0/../../../../arm-none-eabi/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
```

Note the order of the linker arguments:

- `app.0.o`
- `librlibc.rlib`
- `libcore.rlib`
- `libcompiler_builtins.rlib`

And the relevant bits of running `nm` over these object files:

```
$ arm-none-eabi-nm -g **/app*.0.o
         U __aeabi_memcpy4

$ arm-none-eabi-nm -g **/librlibc*.rlib
00000000 W memcmp
00000000 W memcpy
00000000 W memmove
00000000 W memset

$ arm-none-eabi-nm -g ~/.xargo/lib/rustlib/thumbv6m-none-eabi/lib/libcore*.rlib
         U __aeabi_memclr4
         U __aeabi_memcpy
         U __aeabi_memcpy4
         U __aeabi_memset

$ arm-none-eabi-nm -g **/libcompiler_builtins*.rlib
         U memcpy
         U memmove
         U memset
00000000 T __aeabi_memclr
00000000 T __aeabi_memclr4
00000000 T __aeabi_memclr8
00000000 T __aeabi_memcpy
00000000 T __aeabi_memcpy4
00000000 T __aeabi_memcpy8
00000000 T __aeabi_memmove
00000000 T __aeabi_memmove4
00000000 T __aeabi_memmove8
00000000 T __aeabi_memset
00000000 T __aeabi_memset4
00000000 T __aeabi_memset8
```

For this to actually work, `librlibc.rlib` should be passed to the linker *after* `libcompiler_builtins.rlib` but that's not possible because `compiler-builtins` is marked as `#![compiler_builtins]` and will always be the last rlib to be passed to the linker.

Not quite sure how to solve this other than copy paste rlibc's mem* functions into compiler-builtins or special case rlibc in rustc ...

May be worth to mention that this is not a problem in no_std programs that link to libc and not use the "weak" Cargo feature as the linker arguments would look look like: `libcore.rlib .. libcompiler_builtins.rlib -lc`. In that case `-lc` would provide `memcpy` et al.

cc @alexcrichton

japaric:
behind the "mem" Cargo feature, which used to be named "weak"

fixes #126 

Possible solution to #126 
r? @alexcrichton

### DIFF
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -1,0 +1,59 @@
+#[no_mangle]
+pub unsafe extern "C" fn memcpy(dest: *mut u8,
+                                src: *const u8,
+                                n: usize)
+                                -> *mut u8 {
+    let mut i = 0;
+    while i < n {
+        *dest.offset(i as isize) = *src.offset(i as isize);
+        i += 1;
+    }
+    dest
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn memmove(dest: *mut u8,
+                                 src: *const u8,
+                                 n: usize)
+                                 -> *mut u8 {
+    if src < dest as *const u8 {
+        // copy from end
+        let mut i = n;
+        while i != 0 {
+            i -= 1;
+            *dest.offset(i as isize) = *src.offset(i as isize);
+        }
+    } else {
+        // copy from beginning
+        let mut i = 0;
+        while i < n {
+            *dest.offset(i as isize) = *src.offset(i as isize);
+            i += 1;
+        }
+    }
+    dest
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn memset(s: *mut u8, c: i32, n: usize) -> *mut u8 {
+    let mut i = 0;
+    while i < n {
+        *s.offset(i as isize) = c as u8;
+        i += 1;
+    }
+    s
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn memcmp(s1: *const u8, s2: *const u8, n: usize) -> i32 {
+    let mut i = 0;
+    while i < n {
+        let a = *s1.offset(i as isize);
+        let b = *s2.offset(i as isize);
+        if a != b {
+            return a as i32 - b as i32;
+        }
+        i += 1;
+    }
+    0
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,12 +8,6 @@ version = "0.1.0"
 rustc-cfg = "0.2.0"
 gcc = "0.3.36"
 
-[dependencies]
-
-[dependencies.rlibc]
-git = "https://github.com/alexcrichton/rlibc"
-optional = true
-
 [dev-dependencies]
 quickcheck = "0.3.1"
 rand = "0.3.14"
@@ -26,7 +20,8 @@ c = []
 # Mark this crate as the #![compiler_builtins] crate
 compiler-builtins = []
 default = ["compiler-builtins"]
+# Include implementations of memory operations like memcpy
+mem = []
 rustbuild = ["compiler-builtins"]
-weak = ["rlibc/weak"]
 
 [workspace]
--- a/src/arm.rs
+++ b/src/arm.rs
@@ -1,5 +1,8 @@
 use core::intrinsics;
 
+#[cfg(feature = "mem")]
+use mem::{memcpy, memmove, memset};
+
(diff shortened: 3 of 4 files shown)


src/mem.rs
```diff
@@ -0,0 +1,59 @@
...
+        while i < n {
+            *dest.offset(i as isize) = *src.offset(i as isize);
+            i += 1;
+        }
+    }
+    dest
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn memset(s: *mut u8, c: i32, n: usize) -> *mut u8 {
```
japaric: It should be `c_int` but it's better not to depend on `libc`, which is where `c_int` is defined. Right now, `c_int` is defined as `type c_int = i32` in `libc`.

> Will it work on all architectures?

That's a good question. AVR seems to define `int` as 16-bit. I don't know about MSP430. You probably know that better. It's likely that all these implementations don't work correctly on architectures where `sizeof(usize)` is neither 32 or 64.

pftbest: Yes, on MSP430 `int` is 16bit. And I would like to have a working memset on MSP430. My point is that if you make it `isize` instead of `i32` it will be compatible with all architectures that rust supports AFAIK.

PR #129, opened 2016-12-19
Fix memset arguments for MSP430 target.

pftbest:
On MSP430 second argument to memset has `i16` type instead of `i32`.


japaric:
I think we could simplify the logic by creating a "conditional" (cfg) `type c_int = i32` or `= i16` and using that in the function signature because the implementations are the same.

japaric:
Thanks, @pftbest!

@bors r+

PR #132, opened 2017-01-02
mul.rs: use wrapping_mul to avoid infinite recursion

### LINKED ISSUE #127: __aeabi_lmul on thumbv6m-none-eabi
Managed to track a crash when using lmul down to what seems to be a stack overflow, with a loop between the __aeabi_lmul and __muldi3 functions. I think it's due to 0x6ccc in this [dump](https://gist.github.com/seeday/2fe1814b4b26f00e2b3e9eacd9624535).

parched:
rustc in debug mode with a plain multiplication will call @llvm.umul.with.overflow.* which may call the builtin resulting in infinite recursion.

Fixes #127.

Cc @nezza this should be the fix you can try. Note I made this patch in github's web Ui so it's completely untested.


japaric:
Thanks, @parched!

@bors r+

PR #133, opened 2017-01-04
i128 intrinsics

est31:
Adds i128 intrinsics.

Note that this PR doesn't do float intrinsics, due to the missing presence of macros, those are however still required in order for rustc to switch to this crate.


est31:
Builds fail because they are still on an outdated nightly (in fact, nightly with i128 support is released yet).

I'm currently only developing with `RUSTFLAGS="--cfg stage0" cargo test --no-default-features`, and after nightlies will be available, I'll drop the `RUSTFLAGS` argument.

est31:
Okay, I think adding tests is a little bit complicated, as the tests will only work on 64 bit non windows platforms... Is it okay if this doesn't get quickcheck like tests for now? if yes, r? @japaric

est31:
Okay I've rebased and also due to https://github.com/rust-lang/rust/pull/39463 removed the support for `stage0`'s which don't support 128 bit integers yet. Should be now nicer to look at.

japaric:
LGTM. But this needs tests.

> Okay, I think adding tests is a little bit complicated, as the tests will only work on 64 bit non windows platforms

Could you elaborate on this? We can and [are running quickcheck like tests on Windows](https://ci.appveyor.com/project/rust-lang-libs/compiler-builtins/build/1.0.92/job/f1tc4kb5atf7g21o#L148). What's the problem with Windows here? There are no intrinsics on Windows to compare to? In that case, I'd like that we at least test these intrinsics on Linux (i.e. `cfg(not(windows))` the tests).

est31:
> There are no intrinsics on Windows to compare to?

Yup. All the intrinsics in compiler-rt are inside `#ifdef CRT_HAS_128BIT` blocks ([example](https://github.com/llvm-mirror/compiler-rt/blob/d89810afdc2862ea865f5c297ef4ea5839941d25/lib/builtins/fixdfti.c#L13)), and [this variable](https://github.com/llvm-mirror/compiler-rt/blob/d89810afdc2862ea865f5c297ef4ea5839941d25/lib/builtins/int_types.h#L66) is defined only on LP64, but Windows has LLP64. Also, 32 bit platforms don't have LP64 either.

That the intrinsics are not implemented on all platforms was the main reason we needed to do a Rust implementation when i128 support was added to Rust.

Regarding tests, I could try to do quickcheck like tests that are enabled only if we are on 64 bit linux. Alternatively, I could also try to port the compiler-rt testsuite for those intrinsics to Rust. What seems better to you?

Which sounds better to you?

japaric:
> Alternatively, I could also try to port the compiler-rt testsuite for those intrinsics to Rust.

That would be best but if the quickcheck alternative is easier to implement we can start with that.

est31:
@japaric I've added quickcheck tests. r?

src/int/mul.rs
```diff
@@ -120,3 +120,27 @@ mod tests {
...
+            Some(f(a.0, b.0))
+        }
+        fn __muloti4(f: extern fn(i128, i128, &mut i32) -> i128,
+                     a: I128,
+                     b: I128) -> Option<(i128, i32)> {
+            let (a, b) = (a.0, b.0);
+            let mut overflow = 2;
+            let r = f(a, b, &mut overflow);
+            if overflow != 0 && overflow != 1 {
+                return None
```
japaric: wouldn't a value different that 0 or 1 indicate a bug in our implementation? I think such case should panic instead of being skipped.

src/int/udiv.rs
```diff
@@ -380,3 +380,48 @@ mod tests {
...
+        }
+
+        fn __udivmodti4(f: extern fn(u128, u128, Option<&mut u128>) -> u128,
+                        n: U128,
+                        d: U128) -> Option<u128> {
+            let (n, d) = (n.0, d.0);
+            if d == 0 {
+                None
+            } else {
+                // FIXME fix the segfault when the remainder is requested
```
japaric: Could you open an issue about this?

japaric:
Thanks, @est31. r=me with with the `overflow` change.

src/int/udiv.rs
est31: Sure!

est31:
Changed it. Note that travis failed due to bug #136.

japaric:
> Changed it. Note that travis failed due to bug #136.

That should be fixed in #138 so I'm going to put this in the queue.

@bors r+

japaric:
I have manually cleared the caches. Let's try again.

@bors retry

est31:
hmm seems that `mips64-unknown-linux-gnuabi64` and `mips64el-unknown-linux-gnuabi64` tests failed for `__modti3` and `__divti3`:

```
__divti3 - Args: 119123594998676976834373841781095175696 75802332446715572479975239465689939968 
 compiler-builtins: Some(-102509915410633448775648222814998102015)
 compiler_rt:       Some(1)
 gcc_s:             Some(Some(1))
__divti3 - Args: 59561797499338488417186920890547587848 75802332446715572479975239465689939968 

 compiler-builtins: Some(-51254957705316724387824111407499051008)
 compiler_rt:       Some(0)
 gcc_s:             Some(Some(0))
```

and

```
__modti3 - Args: 27708956818918212506987268272516826162 322710947762631927081672291152409133056 
 compiler-builtins: Some(10137537660611676125284951993157747760)
 compiler_rt:       Some(10137537660611676125284951993157747762)
 gcc_s:             Some(Some(10137537660611676125284951993157747762))
__modti3 - Args: 13854478409459106253493634136258413081 322710947762631927081672291152409133056 
 compiler-builtins: Some(13854478409459106253493634136258413080)
 compiler_rt:       Some(13854478409459106253493634136258413081)
 gcc_s:             Some(Some(13854478409459106253493634136258413081))
```

est31:
Well seems mips [does not support i128 integers](http://llvm.org/viewvc/llvm-project?view=revision&revision=224488) on clang, which means we should disable it anyway. Will disable all quickcheck tests for mips.

PR #141, opened 2017-02-07
use AAPCS calling convention on all aeabi intrinsics

### LINKED ISSUE #116: __aeabi_ functions should use the AAPCS calling convention, at least with LLVM 4.0+
I'm not exactly sure about the specifics of this issue but based on 
https://github.com/rust-lang/compiler-rt/pull/26#issuecomment-259976063
I think the __aeabi aliases should use the AAPCS calling convention. `extern "aapcs" fn` seems to work. If the functions that are being aliased don't also use `extern "aapcs"` there may be some shuffling around of registers which may hamper inlining.

japaric:
also, on ARM, inline(always) the actual implementation of the intrinsics so we
end with code like this:

```
00000000 <__aeabi_dadd>:
    (implementation here)
```

instead of "trampolines" like this:

```
00000000 <__aeabi_dadd>:
    (shuffle registers)
    (call __adddf3)

00000000 <__adddf3>:
    (implementation here)
```

closes #116

cc #66
r? @alexcrichton
cc @mattico

### DIFF
--- a/src/arm.rs
+++ b/src/arm.rs
@@ -62,44 +62,44 @@ pub unsafe fn __aeabi_ldivmod() {
 
 // TODO: These aeabi_* functions should be defined as aliases
 #[cfg_attr(not(test), no_mangle)]
-pub extern "C" fn __aeabi_dadd(a: f64, b: f64) -> f64 {
+pub extern "aapcs" fn __aeabi_dadd(a: f64, b: f64) -> f64 {
     ::float::add::__adddf3(a, b)
 }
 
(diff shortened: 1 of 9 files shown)


alexcrichton:
r=me, but looks like arm tests are failing?

japaric:
Yes, looking into it. Seems that there no aeabi intrinsics for i128 operations so LLVM is lowering those to the "x86 named" ones on ARM.

mattico:
:+1:

Issue #144, opened 2017-02-28
explore using utest for testing the thumb targets


japaric:
[utest](https://github.com/japaric/utest) is a custom test crate that supports `no_std` targets like the thumb targets. We could use it to test the thumb targets but there are a few complications:

- utest-cortex-m-qemu, the test runner for thumb targets that works with QEMU emulation, depends on this crate. This means that this crate will appear twice in the dependency graph and that could cause problems.

- even if we get `#[test]` working for the thumb targets, quickcheck still won't work / compile for those targets. So we'll have to create unit tests for the thumb targets that don't rely on quickcheck. We could port the compiler-builtins tests, those don't depend on quickcheck and have the advantage that we could move away from comparing compiler-builtins results to libgcc / libcompiler-rt results as well.

japaric:
I have a new test suite in [compiler-builtins-tester] that works on both `std` and `no_std` targets and doesn't require compiling or linking to compiler-rt / gcc_s. I have checked that it mostly (\*) passes on x86_64, the arm targets and the thumb targets.

[compiler-builtins-tester]: https://github.com/japaric/compiler-builtins-tester

(\*) The powi{s,d}f2 tests are disabled because they don't compile on the thumb targets as we are missing some required intrinsics (e.g. `aeabi_dmul`). And testing on the thumbv6m target is also disabled because of #150.

The test suite works like this:

For each intrinsic, we randomly generate an integration test file (`tests/divdi3.rs`) that contains 10,000 (can be changed) *different* test cases (e.g. for `addsf3`, `1.0 + 1.0` must equal `2.0`). Since the test cases are generated on a x86_64 host the expected results of the test cases are "ground truth" as they were obtained by computing the operation, that the intrinsic emulates, using hardware instructions.

Then we just run `cargo test` / `cross test` / `xargo test` (`utest` is supported out of the box) as appropriate.

While checking that this new test suite works, I found these issues: #145, #148 #150 and #151. Two of them affected `std` targets and should have been caught by the current test suite but weren't. I think the new test suite found them because it does a better job at picking extreme values for the test cases.

Also, with this new test suite, we can test the ARM targets which we are ignoring in the current test suite as some unit tests return results, which are correct, but that don't match compiler-rt / gcc_s results, which are wrong / not as accurate. (cc @mattico)

I propose we move to this new test system as it has wider (target) coverage and doesn't rely on compiler-rt / gcc_s C code bases.

cc @alexcrichton

[...]

PR #146, opened 2017-03-06
fix infinite recursion in divmoddi4 / mulodi4

### LINKED ISSUE #145: ARMv7-M: infinite recursion with __divmoddi4 / __mulodi4
### STR

``` rust
let x: i64 = 6940082214040621058;
let y: i64 = 2305846955181360688;

let z = x % y;
// or
let z = sdiv::__divmoddi4(x, y, &mut 0);
```

with the target: `thumbv7m-none-eabi`.

The above program works fine (it terminates) with the `x86_64-unknown-linux-gnu` target.

### Meta

```
$ rustc -V
rustc 1.17.0-nightly (b1e31766d 2017-03-03)
```

japaric:
on ARMv7-M processors, divmoddi4 was calling mulodi4 and mulodi4 was calling
divmoddi4 leading to infinite recursion. This commit breaks the cycle by using
wrapping multiplication in divmoddi4.

fixes #145

r? @alexcrichton

### DIFF
--- a/src/int/sdiv.rs
+++ b/src/int/sdiv.rs
@@ -55,7 +55,9 @@ macro_rules! divmod {
                 #[cfg(all(feature = "c", any(target_arch = "x86")))]
                 () => unsafe { $div(a, b) },
             };
-            *rem = a - (r * b);
+            // NOTE won't overflow because it's using the result from the
+            // previous division
+            *rem = a - r.wrapping_mul(b);
             r
         }
     }


alexcrichton:
@bors: r+

Out of curiosity, do you think it'd be feasible to CI against this? It seems like a common failure mode...

japaric:
> Out of curiosity, do you think it'd be feasible to CI against this? It seems like a common failure mode...

We would have to test the intrinsics on `no_std` targets like `thumbv7m-none-eabi` to catch these. That's possible with [utest](https://github.com/japaric/utest#testing-on-an-emulated-cortex-m-processor) but `quickcheck` doesn't compile on `no_std` so we would have to add unit tests that don't depend on quickcheck specifically to test these targets. See #144.

Ideally, I'd like a much simpler test system. One that doesn't depend on compiling `libgcc` or `compiler-rt` or even quickcheck. Basically have something like this:

``` rust
#[test]
fn muldi3() {
    assert_eq!(__muldi3(1, 1), 1);
    assert_eq!(__muldi3(1, 2), 2);
    assert_eq!(__muldi3(2, 3), 6);
    // ...
}
```

Where the assertions are randomly generated before we run the test suite. compiler-rt test suite is like this but has fixed test cases rather than randomly generated ones.

alexcrichton:
Could we perhaps have a separate project which is compatible with utest, not using quickcheck? (and then use both on CI)

japaric:
> Ideally, I'd like a much simpler test system.

I got a proof of concept for this system [here](https://github.com/japaric/compiler-builtins-tester). I have tested that it works for the `x86_64-unknown-linux-gnu`, `armv7-unknown-linux-gnueabihf` and `thumbv7m-linux-eabi` targets. I have only tested `muldi3` and `mulodi4` so far but apparently I've already found another bug in `mulodi4`:

```
thread 'mulodi4' panicked at 'attempt to subtract with overflow', /cargo/git/checkouts/compiler-builtins-ec094dc45a0179c8/644a1c9/src/int/sdiv.rs:13
```

Issue #154, opened 2017-04-10
can't test powi*f2 intrinsics on thumb


japaric:
because we are missing Rust implementations of aeabi_dmul, etc.

AaronKutch:
I think this was fixed a long time ago, new tests I wrote confirm powi works on thumb

AaronKutch:
Better testing has been implemented by #397, this should be closed

Issue #150, opened 2017-03-13
thumbv6m: debug assertion in divmoddi4


japaric:
``` rust
#[test]
fn divmoddi4() {
    let (a, b) = (4165187992, 3876330169);
    let mut rem = 0;
    let c = __divmoddi4(a, b, &mut rem_);
}
```

```
$ qemu-arm target/thumbv6m-linux-eabi/debug/divmoddi4-04f11ecf75829660
running 1 tests
test divmoddi4 ...
panicked at 'attempt to multiply with overflow', /home/japaric/.cargo/git/checkouts/compiler-builtins-ec094dc45a0179c8/9681358/src/int/udiv.rs:95
qemu: uncaught target signal 4 (Illegal instruction) - core dumped
[1]    21208 illegal hardware instruction (core dumped)  qemu-arm target/thumbv6m-linux-eabi/debug/divmoddi4-04f11ecf75829660
```

thumbv7m targets don't have this problem.

japaric:
Another case that results in an assertion `(1750458948, 2611493155)` with `divdi3`.

japaric:
Testing all these intrinsics fail due to this debug assertion

- divdi3
- divmoddi4
- moddi3
- modsi3
- udivmoddi4
- udivmodsi4
- umoddi3
- umodsi3

parched:
This is probably fix by https://github.com/rust-lang/llvm/pull/79

parched:
@japaric I think this should be fixed in the latest nightly now that https://github.com/rust-lang/rust/pull/42410 has landed.

japaric:
I tested with `rustc 1.19.0-nightly (fe7227f6c 2017-06-16)` and I'm still seeing the debug assertion. I'll try again with the next nightly release.

parched:
Ok, that commit had the fix in, so this must be caused by something else.

parched:
@japaric I'd like to debug this, how exactly did you run the test? Do I have to build all of std with Xargo for `thumbv6m-linux-eabi`?

japaric:
@parched I actually ran the whole test suite with `sh ci/run.sh thumbv6m-linux-eabi`. But you can run a single custom test like this:

``` console
$ edit tests/custom.rs && cat $_
```

``` rust
#![feature(compiler_builtins_lib)]
#![no_std]

extern crate compiler_builtins;
extern crate utest_cortex_m_qemu;
#[macro_use]
extern crate utest_macros;

[...]

Issue #137, opened 2017-02-04
Segfault in __udivmodti4 quickcheck when remainder is requested


est31:
PR #133 adds quickcheck tests for `__udivmodti4` with the code that does checking with remainder enabled commented out. I commented it out as I got a segfault.

est31:
The segfault happens at:

```
0x00007ffff55fc22e in __udivmodti4 (a=11837846198047054611, b=10468428215946443262, rem=0x7ffff67fc388)
    at compiler-rt/lib/builtins/udivmodti4.c:47
47                      *rem = n.s.low % d.s.low;
```

japaric:
According to [this Travis build](https://travis-ci.org/rust-lang-nursery/compiler-builtins/builds/220750542), this appears to only happens on MIPS.

est31:
@japaric as you can see from this travis log, it failed on multiple 64 bit platforms (x86_64, aarch64, mips64, powerpc64, ...): https://travis-ci.org/rust-lang-nursery/compiler-builtins/builds/198218713

japaric:
@est31 AFAICS, all the errors in the build you have linked seem to be due to "missing symbol" which indicates a bad cache.

est31:
@japaric ah you are right. I *think* I remember though reproducing the issue on my native platform (x86_64). Maybe you can try to comment out [this code](https://github.com/rust-lang-nursery/compiler-builtins/blob/9aa3a257d61ed51d5a38360b2aa37e1c571d594b/src/int/udiv.rs#L423-L426) and try whether you can reproduce?

japaric:
> Maybe you can try to comment out this code and try whether you can reproduce?

I can  locally repro on x86_64. The weird thing is that we are not hitting this bug on x86_64 and other arches in #155.

PR #155, opened 2017-04-10
no-std friendly test suite

### LINKED ISSUE #87: Failed to build as a dependency
Added this to my Cargo.toml

``` toml
[dependencies]
rustc_builtins = { git = "https://github.com/japaric/rustc-builtins" }
```

tried to compile using

```
RUST_BACKTRACE=1 xargo build --target=stm32f429zi --verbose
```

got this error

```
error: failed to run custom build command for `rustc_builtins v0.1.0 (https://github.com/japaric/rustc-builtins#89ebc46f)`
process didn't exit successfully: `/Users/vadzim/Downloads/rplayground/target/debug/build/rustc_builtins-c925dd23196c1b21/build-script-build` (exit code: 101)
--- stdout
cargo:rerun-if-changed=build.rs

--- stderr
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()', ../src/libcore/result.rs:799
stack backtrace:
   1:        0x109c560d8 - std::sys::backtrace::tracing::imp::write::h22f199c1dbb72ba2
   2:        0x109c5b56f - std::panicking::default_hook::{{closure}}::h9a389c462b6a22dd
   3:        0x109c5a605 - std::panicking::default_hook::h852b4223c1c00c59
   4:        0x109c5ab56 - std::panicking::rust_panic_with_hook::hcd9d05f53fa0dafc
   5:        0x109c5a9f4 - std::panicking::begin_panic::hf6c488cee66e7f17
   6:        0x109c5a912 - std::panicking::begin_panic_fmt::hb0a7126ee57cdd27
   7:        0x109c5a877 - rust_begin_unwind
   8:        0x109c80c30 - core::panicking::panic_fmt::h9af671b78898cdba
   9:        0x109c05c01 - core::result::unwrap_failed::hb3293571464cc40d
  10:        0x109c01575 - <core::result::Result<T, E>>::unwrap::hebcc09568bcc73ba
  11:        0x109c0fa1a - build_script_build::main::hfbb6c5966e861548
  12:        0x109c5bb2a - __rust_maybe_catch_panic
  13:        0x109c5a146 - std::rt::lang_start::h14cbded5fe3cd915
  14:        0x109c10e69 - main

error: `cargo` process didn't exit successfully
```

### LINKED ISSUE #90: Floating point add gives different results than compiler-rt on arm-unknown-linux-gnueabi

### LINKED ISSUE #136: Removal of target dir required when adding symbols to compiler-rt-cdylib
I was attempting to add some new symbols to compiler-rt-cdylib. It worked, but the tests failed. I had to remove the target directory and do a clean re-build of the entire project to update the so file. This should not be required.

### LINKED ISSUE #144: explore using utest for testing the thumb targets

japaric:
this PR moves the test suite away from rand and quickcheck by generating the
unit tests on the host rather than on the target. Because the host is usually
x86_64 (which has hardware ops for most of the operations that need intrinsics
on other architectures), we compute the right answer on the host instead of
comparing against the libgcc_s / compiler-rt implementation on the target. With
this change, we can drop the compiler-rt-cdylib and gcc_s subprojects.

Finally we add a target specific dependency on utest, a no-std testing
framework, to additionally run the test suite on the thumb targets.

r? @alexcrichton
cc @mattico
closes #87 (we no longer depend on the rustc-cfg crate)
closes #90 (we now compare against the result of using hardware ops)
closes #136 (compiler-rt-cdylib has been removed)
closes #144 (we are now running tests on the thumb targets)

---

I have tested this locally on x86_64, ARM and the 4 thumb targets. x86_64 and
ARM pass without a problem. On all thumb targets, I had to disable the test of
the powi*f2 intrinsic as other intrinsics are needed to run that intrinsics
(cf. #154). And on thumbv6, I had to disable several tests due to the debug
assertion reported in #150.


japaric:
I'm trying to debug this crash from the MIPS build job:

```
test divti3 ... qemu: uncaught target signal 11 (Segmentation fault) - core dumped
```

but AFAICT, when running this unit test:

``` rust
#[test]
fn divti3() {
    // A
    for &((a, b), c) in TEST_CASES {
        // B
        let c_ = __divti3(a, b);
        assert_eq!(((a, b), c), ((a, b), c_));
    }
}
```

the program crashes *before* reaching the first iteration of the for loop (so it never reaches point B). I thought it could be may due to the size of the TEST_CASES array which contains 10_000 test cases but reducing it to 1 test case still causes the crash. Other unit tests that make use of i128 types, like multi3, work fine so I have no idea what's going on here.

alexcrichton:
Looks great to me! It'd be nice to reduce the verbosity of the build script, but not critical. r=me when you feel like this should land.

japaric:
So, I ignored the divti3 on MIPS to see if the other i128 tests pass I got this assertion while testing the modti3 intrinsic:

```
running 1 test
test modti3 ... FAILED

failures:

---- modti3 stdout ----
        thread 'modti3' panicked at 'assertion failed: `(left == right)` (left: `((-922337203685477580757, -922337203685477580700), -115292150460684697557)`, right: `((-922337203685477580757, -922337203685477580700), -115292150460684697557)`)', /target/mips-unknown-linux-gnu/debug/build/compiler_builtins-86a4e859175c573b/out/modti3.rs:10039
```

If you can't tell from glancing at it both the LHS and the RHS are exactly equal! Yet the assertion for equality fails.

There doesn't seem to be a problem with equality of i128. That operation doesn't even use an intrinsic on MIPS.

This program works fine on x86_64 and on MIPS:

``` rust
#![feature(i128_type)]

fn main() {
    let x: ((i128, i128), i128) = ((-922337203685477580757, -922337203685477580700), -115292150460684697557);
    let y: ((i128, i128), i128) = ((-922337203685477580757, -922337203685477580700), -115292150460684697557);

    assert_eq!(x, y);
    println!("OK");
}
```

```
$ cargo run
OK

$ cross run --target mips-unknown-linux-gnu
OK
```

Seems like this could some undefined behavior or ABI problem related to #137 as the div / mod / divmod intrinsics are all related. @est31, do you remember if that problem was MIPS specific? I'm only seeing the problem on 32-bit MIPS but not on MIPS64.

japaric:
I don't know what's up with bors but tests have passed on [AppVeyor] and on [Travis] so I'm going to merge manually.

[AppVeyor]: https://ci.appveyor.com/project/rust-lang-libs/compiler-builtins/build/1.0.175
[Travis]: https://travis-ci.org/rust-lang-nursery/compiler-builtins

PR #165, opened 2017-06-01
remove unused macros

japaric:
to fix warnings

### DIFF
--- a/src/float/conv.rs
+++ b/src/float/conv.rs
@@ -1,17 +1,6 @@
 use float::Float;
 use int::Int;
 
-macro_rules! fp_overflow {
-    (infinity, $fty:ty, $sign: expr) => {
-        return {
-            <$fty as Float>::from_parts(
-                $sign,
-                <$fty as Float>::exponent_max() as <$fty as Float>::Int,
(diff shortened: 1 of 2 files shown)


japaric:
MIPSel failure while testing floattidf. This has nothing to do with this change:

> 	thread 'floattidf' panicked at 'assertion failed: `(left == right)` (left: `((-3518437208883100,), 14318300716730535424, 14318300716730535168, true)`, right: `((-3518437208883100,), 14318300716730535424, 14318300716730535168, false)`)', /target/mipsel-unknown-linux-gnu/debug/build/compiler_builtins-8619b5d3e04f0666/out/floattidf.rs:10055

Looks like the value computed on the host as the correct answer is wrong. `-3518437208883100_i128` casted to f64 is `-3518437208883100_f64` which transmuted to u64 is `14062771311441084216_u64`. That value doesn't match the `14318300716730535424_u64` from the failure message. Maybe there was some UB in the generation of a previous test case? cc @est31 

@bors retry

est31:
Yeah the tests are unfortunately a bit spurious.

PR #166, opened 2017-06-22
Test with the 'c' feature enabled on CI

alexcrichton:


alexcrichton:
As some background on this, I'm looking to integrate this repo upstream into rust-lang/rust again. I initially ran into some bugs with the `c` feature enabled on Windows, so the main goal here is to enable the `c` feature on CI. Unfortunately testing with the `c` feature is relatively broken right now (lots of stuff is cfg'd out). As a result, this is going to be a growing patch as I fix the things I come across.

The biggest change I've made so far is how to define intrinsics. I found the definitions duplicated and somewhat hard to read due to the indirections through macros. I settled on a strategy where the intrinsics are generated inside of an `intrinsics!` macro which uses Rust syntax but helps recognize "special attributes" to control various aspects. The thinking is that this'll mean that there's only one definition of an intrinsic, and it can get decorated with weird platform hacks and whatnot.

The second change is to move intrinsic definitions from macros to traits. Generic programming with integers isn't the best but I found it to be much easier to navigate than macros at least, personally. Feedback is of course always welcome!

A question I've run across though, @japaric what are [these attributes doing](https://github.com/rust-lang-nursery/compiler-builtins/blob/master/src/int/mul.rs#L77-L78)? Is there a reason to ensure that those symbols aren't mangled other than "hygiene"?

I'll be working to get this green over time, but it may take some time. I'll ping this once it's all good!

est31:
Great design choices, to port to a trait based system from macros, and to realize the windows abi hack via a pseudo-attribute `#[win64_128bit_abi_hack]`. Makes its usage much nicer :).

alexcrichton:
Huzzah everything's finally green!

r? @japaric

src/macros.rs
```diff
@@ -0,0 +1,286 @@
...
+        }
+
+        #[cfg(target_arch = "arm")]
+        pub mod $name {
+            intrinsics! {
+                pub extern "aapcs" fn $alias( $($argname: $ty),* ) -> $ret {
+                    super::$name($($argname),*)
+                }
+
+                pub extern $abi fn $name( $($argname: $ty),* ) -> $ret {
```
japaric: I think we can drop this. LLVM won't emit calls to x86-style intrinsics (e.g. addsf); instead it will always use the aeabi names (e.g. aeabi_fadd). The reason we didn't do this before is that our use of cfg would have become even more unwieldy. It seems that this be easy to fix with the new approach.

src/qc.rs
```diff
@@ -1,311 +0,0 @@
-//  When testing functions, QuickCheck (QC) uses small values for integer (`u*`/`i*`) arguments
```
japaric: I had forgotten we no longer use this. Thanks for the cleanup!

appveyor.yml
```diff
@@ -1,21 +1,45 @@
...
   - cargo -V
 
 build: false
 
 test_script:
-  - cargo build --target %TARGET%
-  - cargo build --release --target %TARGET%
-  - cargo test --no-default-features --features gen-tests --target %TARGET%
-  - cargo test --no-default-features --features gen-tests --release --target %TARGET%
+  - sh ci/run.sh %TARGET%
```
alexcrichton: 🌠

src/macros.rs
alexcrichton: At some point during testing this was required for... whatever reason. But it can indeed be removed now, yay!

ci/run.sh
```diff
@@ -87,22 +84,70 @@ case $TRAVIS_OS_NAME in
...
+        exit 1
+    fi
+    set -ex
+done
+
+rm -f $path
+
+# Verification of the `intrinsics` program doesn't work on thumb targets right
+# now.
+case $1 in
```
alexcrichton: Ah so right now it actually has more to do with `panic_unwind`. Linking the example on MinGW required the panic runtime for all the requisite symbols (hard to get around otherwise manually). I figured that the this wouldn't get us much more coverage than the ARM targets already get, but do you think it's worth the `#[cfg]` to only link `panic_unwind` on Windows and get this working again for Thumb?

japaric: > do you think it's worth the #[cfg] to only link panic_unwind on Windows and get this working again for Thumb?

Yeah, I think we should. The thumb targets, specially the thumbv6m one, actually require more intrinsics (mainly float ops iirc) than the ARM targets so it would be good to test that we don't drop, by mistake, an intrinsic that those targets need.

mattico:
The refactorings you applied to the integer intrinsics (using traits vs macros, adding wrapping_ ops to the trait) could also be applied to the float intrinsics, no? I'll make an issue for it and work on it at some point. Constantly converting between Wrapping() types and regular types really muddies the fp_add implementation.

alexcrichton:
I believe so, yeah, I was just running a little low on steam around that point :(

It'd be a most welcome PR though I'd imagine! (I'm hoping to land this in the next few hours)

alexcrichton:
Hm that looks spurious, so I'm going to merge.

PR #175, opened 2017-07-06
Add `__rust_probestack` intrinsic

alexcrichton:
Will be required for rust-lang/rust#42816


mattico:
Just curious: does this obsolete `___chkstk_ms`?  Was that even used in the first place?

alexcrichton:
Nah we'll still need to keep those around. On non-Windows platforms `__rust_probestack` will be used but on Windows the `__chkstk` business is still used.

It may, however, make sense to consolidate all these intrinsics in one module

PR #181, opened 2017-07-22
Avoid memcpy references in unoptimized code

TimNN:
Fixes rust-lang/rust#43411.


alexcrichton:
@bors: r+

I wonder if it'd be possible to add a test for this as well?

PR #189, opened 2017-09-12
Update rustbuild commentary

tamird:
r? @nikomatsakis

This is the reason that https://github.com/rust-lang/rust/pull/44509 doesn't work - the `rustbuild` feature _is_ actually used, it was just incorrectly documented here and I missed it.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -4106,9 +4106,11 @@ mod c {
         // also needs to satisfy intrinsics that jemalloc or C in general may
         // need, so include a few more that aren't typically needed by
         // LLVM/Rust.
-        sources.extend(&[
-            "ffsdi2.c",
-        ]);
+        if cfg!(feature = "rustbuild") {
+            sources.extend(&[
+                "ffsdi2.c",
+            ]);
+        }
 
         if target_os != "ios" {
             sources.extend(
@@ -4359,7 +4361,9 @@ mod c {
             sources.remove(&["aeabi_cdcmp", "aeabi_cfcmp"]);
         }
 
-        let root = if env::var_os("CARGO_FEATURE_RUSTBUILD").is_some() {
+        // When compiling in rustbuild (the rust-lang/rust repo) this build
+        // script runs from a directory other than this root directory.
+        let root = if cfg!(feature = "rustbuild") {
             Path::new("../../libcompiler_builtins")
         } else {
             Path::new(".")


build.rs
```diff
@@ -4102,14 +4103,6 @@ mod c {
             ],
         );
 
-        // When compiling in rustbuild (the rust-lang/rust repo) this library
-        // also needs to satisfy intrinsics that jemalloc or C in general may
-        // need, so include a few more that aren't typically needed by
-        // LLVM/Rust.
-        sources.extend(&[
-            "ffsdi2.c",
```
alexcrichton: How come this was moved up?

tamird: I could leave it where it is, but that invites questions (why is it separate from the rest of the files?).

The comment here suggests an answer, but doesn't make sense given that the file's inclusion is unconditional (the condition was removed in 91eaa859027ea4806997112cb9f53b25e51106ca, but I don't know why).

So since the inclusion is unconditional, I figured I'd just move it into the main slice.

build.rs
```diff
@@ -4003,16 +4003,15 @@ mod c {
 
     /// Compile intrinsics from the compiler-rt C source code
     pub fn compile(llvm_target: &[&str]) {
-        let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
-        let target_env = env::var("CARGO_CFG_TARGET_ENV").unwrap();
-        let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
```
alexcrichton: These all need to stay as environment variables due to cross compilation

build.rs
alexcrichton: This is currently in a dedicated section because of the comment. This file is not needed to compile Rust but it's needed to compile Rust's dependencies, namely jemalloc.

build.rs
tamird: Got it, thanks

build.rs
tamird: OK, so the conditional bit should be restored?

tamird:
@alexcrichton I think this ready for another look - `mipsel-unknown-linux-gnu` failed, but the failure looks spurious/unrelated.

tamird:
@alexcrichton I think #188 was maybe incorrect; should we restore that feature and add tests that exercise it?

alexcrichton:
Sure!

tamird:
Hm, it's not exactly clear to me how to meaningfully test that outside of "does rustc still compile", so I'm going to leave it as is. Is this good to go?

alexcrichton:
@bors: r+

Looks good!

tamird:
@alexcrichton any idea what that failure is?
```
---- floattidf stdout ----
	thread 'floattidf' panicked at 'assertion failed: `(left == right)`
  left: `((-28147497671065500,), 14331630114050693152, 14331630114050693120, true)`,
 right: `((-28147497671065500,), 14331630114050693152, 14331630114050693120, false)`', /target/mips-unknown-linux-gnu/debug/build/compiler_builtins-04fe505231ccc40a/out/floattidf.rs:10055:8
```
I can't even make sense of the output.

alexcrichton:
[spurious failure](https://github.com/rust-lang-nursery/compiler-builtins/issues/168)

Issue #183, opened 2017-07-24
Parts of libcompiler_builtins are compiled as WX


djc:
Similar to https://github.com/rust-lang/rust/issues/34482, I still see a number of warnings like this when compiling rust-1.19.0:

```
 * !WX --- --- usr/lib64/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-701380125126dfef.rlib:chkstk.o
 * !WX --- --- usr/lib64/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-701380125126dfef.rlib:chkstk2.o
 * !WX --- --- usr/lib64/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-701380125126dfef.rlib:floatundidf.o
 * !WX --- --- usr/lib64/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-701380125126dfef.rlib:floatundisf.o
 * !WX --- --- usr/lib64/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-701380125126dfef.rlib:floatundixf.o
```

alexcrichton:
Oh dear definitely seems like something to fix! @djc would you be up for submitting a patch for at least the easy piece, removing compilation of chkstk files on non-windows platforms?

djc:
Uh, I guess. Do you mean like this?

```diff
diff --git a/build.rs b/build.rs
index 25cc520..787906f 100644
--- a/build.rs
+++ b/build.rs
@@ -4172,8 +4172,6 @@ mod c {
                     &[
                         "i386/ashldi3.S",
                         "i386/ashrdi3.S",
-                        "i386/chkstk.S",
-                        "i386/chkstk2.S",
                         "i386/divdi3.S",
                         "i386/floatdidf.S",
                         "i386/floatdisf.S",
@@ -4188,6 +4186,14 @@ mod c {
                         "i386/umoddi3.S",
                     ],
                 );
+                if target_os == "windows" {
+                    sources.extend(
+                        &[
+                            "i386/chkstk.S",
+                            "i386/chkstk2.S",
+                        ],
+                    );
+                }
             }
         }
```

alexcrichton:
I think that should do it yeah!

If you'd like as well, it'd be awesome to add a test into this repo to ensure that we don't regress this.

djc:
[...]

PR #194, opened 2017-09-15
Implement x86 chkstk in "rust"

mattico:
cc #183 

Basically the same as the x86_64 ones, except `__alloca` doesn't need to fix the parameter register. I've manually verified that the disassembly is the same, and that these work in a compiled rust program.

The second commit disables compiling probestack functions for `feature = mangled-names`. They aren't needed during testing because they aren't comparison tested and the unmangled versions are the ones that actually get used.

r? @alexcrichton


alexcrichton:
@bors: r+

Thanks!

Issue #197, opened 2017-09-28
Int <-> float conversion broken by #192


est31:
#192 seems to have broken conversion between 128 bit ints and floats (https://github.com/rust-lang/rust/pull/44785):

```
[01:01:43] ------------------------------------------
[01:01:43] stderr:
[01:01:43] ------------------------------------------
[01:01:43] thread 'main' panicked at 'assertion failed: `(left == right)`
[01:01:43]   left: `0`,
[01:01:43]  right: `547625059298595101446575784722432`', /checkout/src/test/run-pass/u128.rs:56:4
[01:01:43] note: Run with `RUST_BACKTRACE=1` for a backtrace.
[01:01:43] 
[01:01:43] ------------------------------------------
```

cc @alexcrichton

est31:
I'm working on the issue

est31:
Hmmm, my attempt to isolate this failed:

```rust
#![feature(compiler_builtins_lib)]
#![feature(i128_type, test)]

extern crate test;
use test::black_box as b;

extern "C" {
    pub fn __floatuntisf(i: u128) -> f32;
    pub fn __floatuntidf(i: u128) -> f64;
    pub fn __fixunssfti(f: f32) -> u128;
    pub fn __fixunsdfti(f: f64) -> u128;
}

#[test]
fn float_int_conv() {
    unsafe {
        let l :u128 = b(432 << 100);
        assert_eq!(__fixunssfti(__floatuntisf(l)), l);
        assert_eq!(__fixunsdfti(__floatuntidf(l)), l);
    }
}
```

... works fine.

est31:
So I could reproduce the issue by doing `optimize-tests = false` in config.toml and `./x.py test src/test/run-pass --test-args 128 --stage 1`. If you swap the `--stage 1` by `--stage 0`, its not reproducible. Very weird.

aidanhs:
Unsure if related, but I'm attempting to build the latest commit of rust and am getting
```
warning: this expression will panic at run-time
  --> src/rustc/compiler_builtins_shim/../../libcompiler_builtins/src/float/conv.rs:56:21
   |
56 |             if (a & (1 << mant_dig)) != 0 {
   |                     ^^^^^^^^^^^^^^^ attempt to shift left with overflow
...
80 |         int_to_float!(i, i32, f64)
   |         -------------------------- in this macro invocation

warning: this expression will panic at run-time
[...]

PR #198, opened 2017-09-29
Partial revert of the float conversion refactor in #192

est31:
PR to make updating compiler-builtins for newer rustc possible again (cc #197). I don't really know what #197 is about yet, so I want to un-block updates of compiler-builtins by reverting the conversion implementation to its older macro form. The other changes of #192 are left, even the changes to `mod.rs` done by the two partially reverted commits.

I've tested the changes of this PR with a local build of rustc and it can't reproduce #197. However, it does not close the issue, as optimally we'd have the code back in master again.

r? @alexcrichton

### DIFF
--- a/src/float/conv.rs
+++ b/src/float/conv.rs
@@ -1,5 +1,5 @@
 use float::Float;
-use int::{Int, CastInto};
+use int::Int;
 
 macro_rules! int_to_float {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #199, opened 2017-10-06
Remove redundant rerun-if-changed directive

tamird:
The docs at doc.crates.io/build-script.html state:

> Note that if the build script itself (or one of its dependencies)
> changes, then it's rebuilt and rerun unconditionally, so
> cargo:rerun-if-changed=build.rs is almost always redundant (unless you
> want to ignore changes in all other files except for build.rs).

So remove this redundant thing.

r? @alexcrichton

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,8 +3,6 @@
 use std::env;
 
 fn main() {
-    println!("cargo:rerun-if-changed=build.rs");
-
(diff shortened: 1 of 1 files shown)


alexcrichton:
The final clause,

> unless you want to ignore changes in all other files except for build.rs

is why this exists

(unfortunately we can't remove it)

PR #200, opened 2017-10-07
drop binfmt_misc dependency

japaric:
Instead use the Cargo runner feature. The binfmt_misc approach requires running a privileged
container for setup. Not all docker setups support privileged containers so the test suite should be
more accessible with this change as no privileged container is needed.

r? @alexcrichton


alexcrichton:
@bors: r+

Nice!

PR #202, opened 2017-11-01
Update compiler-rt submodule

ishitatsuyuki:
This brings in LLVM 5 related changes. Not tested.


ishitatsuyuki:
CI failure due to homebrew failing on outdated Ruby.

alexcrichton:
I don't think we're ready for LLVM 5 to land upstream in rustc, so why would we land this ahead of time?

ishitatsuyuki:
There's sanitizer linkage issues; and we're behind the schedule.

alexcrichton:
That's... not really an answer though? We can't land this until we can upgrade to LLVM 5, right?

Amanieu:
compiler-rt doesn't depend on LLVM and can be updated independently.

ishitatsuyuki:
* the current code base has compile issues with glibc 2.26.
* the current code base is incompatible with latest LLVM sanitizers.
* the upgrade should be back compat, but it may cause the sanitizers to fail weird. Needs test.

alexcrichton:
Aren't the newer sanitizers not guaranteed to be compatible with older LLVM? Are we sure that this upgrade of the compiler-rt submodule actually works with the LLVM rustc is using?

ishitatsuyuki:
Let's test it. I will open a WIP PR to rust repo tomorrow.

alexcrichton:
According to https://github.com/rust-lang/rust/pull/45779 sounds like we're not quite ready for this, so closing

PR #205, opened 2017-11-08
Implement mulsf3 and muldf3

aeleos:
It is testing and working with randomly generated unit tests. I had to add a extra generator function to generate larger floats as the previous one only generated extremely small floats that when multiplied together always produced 0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,11 +170,11 @@ features = ["c"]
 - [x] lshrdi3.c
 - [x] moddi3.c
 - [x] modsi3.c
-- [ ] muldf3.c
+- [x] muldf3.c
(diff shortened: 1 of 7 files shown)


alexcrichton:
Nice! I think that there may be some overlap here with https://github.com/rust-lang-nursery/compiler-builtins/pull/204 perhaps?

cc @vlad9486

This looks pretty great to me though! I'm curious to hear what @vlad9486 thinks though

aeleos:
@alexcrichton yea it seems that we ended up working on the same thing at the same time randomly.

alexcrichton:
Ok let's merge!

PR #206, opened 2017-11-12
Implement divsf3 and divdf3

aeleos:
It has two different functions for 32 and 64 bit as the 64 bit version but has some differences in how it does the calculations. Both are integrated into the testing system and pass all tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,11 +133,11 @@ features = ["c"]
 - [ ] arm/unordsf2vfp.S
 - [x] ashldi3.c
 - [x] ashrdi3.c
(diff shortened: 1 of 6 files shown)


alexcrichton:
Thanks! Just to confirm, this is all basically ported from the C version, right?

aeleos:
Yep it is pretty much copied line by line from C version. I had to get a little creative due to the type system and the C versions liberal use of casting between types and unsigned number negations.

alexcrichton:
Ok sounds great, thanks so much!

PR #207, opened 2017-11-15
Add "volatile" and "memory" clobber to asm! that doesn't fall through.

sunfishcode:
Use the "volatile" option and the "memory" clobber on inline asm that does
things like return directly, to reduce the chances of compilers rearranging
the code.

This fixes the conflict with TrapUnreachable in https://github.com/rust-lang/rust/pull/45920#issuecomment-344718264.


alexcrichton:
:+1:

PR #210, opened 2017-11-25
Implement all the 128-bit operator lang items (Rust 45676 part 2)

scottmcm:
https://github.com/rust-lang/rust/pull/46093 declared lang items for all the 128-bit operators.  This PR implements them.

I was unsure how testing should work, and was unable to get the tests in `/tests` to even compile locally.  

cc https://github.com/rust-lang/rust/issues/45676 @nagisa @est31


est31:
> I was unsure how testing should work, and was unable to get the tests in /tests to even compile locally.

@scottmcm https://github.com/rust-lang-nursery/compiler-builtins/issues/160#issuecomment-293914018

src/int/add.rs
```diff
@@ -0,0 +1,124 @@
...
+#[allow(dead_code)]
+fn rust_u128_addo(a: u128, b: u128) -> (u128, bool) {
+    let mut oflow = 0;
+    let r = a.addo(b, &mut oflow);
+    (r, oflow != 0)
+}
+
+#[test]
+fn test_add() {
+    assert_eq!(rust_u128_add(1, 2), 3);
```
est31: you should put those into the tests/ directory, that's where all the other tests are. I guess one doesn't need to go the full way that most of the other tests do by autogenerating values though.

alexcrichton:
Thanks @scottmcm! Additionally thanks for the tests! Everything here looks fantastic to me

PR #218, opened 2017-12-20
Fix __kuser_cmpxchg inline asm

Amanieu:
The wrong register was passed to BLX. However I just changed them to normal function calls to simplify things.


Amanieu:
On second thought, I think I'm just going to treat these as normal function calls rather than inline asm, like @parched suggested in https://github.com/rust-lang-nursery/compiler-builtins/pull/115#discussion_r157832778.

alexcrichton:
@bors: r+

Thanks!

PR #220, opened 2017-12-26
Use kernel user helpers for armv4t

malbarbo:
Rename armv5te cfg to kernel_user_helpers and enable it for armv4t.


malbarbo:
The armv4t target is being add in https://github.com/rust-lang/rust/pull/47018

alexcrichton:
er,

@bors: retry

PR #223, opened 2018-01-19
Skip aeabi_cdcmp and aeabi_cfcmp on big-endian targets

paoloteti:
First of all aeabi_cdcmp and aeabi_cfcmp are never called LLVM.
Second are little-endian only, so build fail on big-endian targets as show below:

```
cargo:warning=./compiler-rt/lib/builtins/arm/aeabi_cdcmp.S:13:2: error: #error big endian support not implemented
cargo:warning= #error big endian support not implemented
cargo:warning=  ^~~~~
```

Temporally workaround: exclude these files for big-endian targets.

Tested on a Texas Instruments TMS570LS3137 (Cortex-R4F big-endian) based board.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -5464,7 +5464,7 @@ mod c {
                 ],
             );
 
-            // First of all aeabi_cdcmp and aeabi_cfcmp are never called LLVM.
+            // First of all aeabi_cdcmp and aeabi_cfcmp are never called by LLVM.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Looks like there may be a build failure though?

alexcrichton:
Thanks!

PR #224, opened 2018-01-25
Fix some typos/bugs with float comparison intrinsics

alexcrichton:
* I believe `__gtdf2` erroneously used `f32` instead of `f64`
* Most of these needed `#[arm_aeabi_alias]` to ensure they're correctly called
  through the alias
* Some existing aliases were corrected with the right names


alexcrichton:
@whitequark mind double checking?

whitequark:
LGTM, though I am far from an EABI expert.

PR #225, opened 2018-01-29
Add support for mul[s/d]f3vfp and div[s/d]f3vfp

paoloteti:
Here using `"C"` the compiler will use `"aapcs"` or `"aapcs-vfp"` depending on target configuration.

Of course this translates in a call to `__aeabi_fdiv` / `__aeabi_fmul` on non-HF targets.

On `eabi` targets with +vfpv2/vfpv3 LLVM generate:

```
   vmov	s0, r1
   vmov	s2, r0
   vdiv.f32	s0, s2, s0
   vmov	r0, s0
   bx	lr
```

On `eabihf` targets with +vfpv3-d16/d32/f32 +fp-only-sp LLVM generate:

```
  vdiv.f32	s0, s0, s1
  bx	lr
```

That's exactly what We need for [div/mul][s/d]f3vfp.S

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,9 +96,9 @@ features = ["c"]
 - [x] arm/aeabi_memset.S
 - [x] arm/aeabi_uidivmod.S
 - [x] arm/aeabi_uldivmod.S
-- [ ] arm/divdf3vfp.S
+- [x] arm/divdf3vfp.S
 - [ ] arm/divmodsi4.S (generic version is done)
-- [ ] arm/divsf3vfp.S
+- [x] arm/divsf3vfp.S
 - [ ] arm/divsi3.S (generic version is done)
 - [ ] arm/eqdf2vfp.S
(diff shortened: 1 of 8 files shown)


alexcrichton:
Thanks! Could you be sure to add some tests for these as well?

paoloteti:
Done! Of course on QEMU/TravisCI We still have an ICE:

`error: internal compiler error: librustc_metadata/cstore_impl.rs:131: get_optimized_mir: missing MIR for `DefId(3/0:10 ~ utest_cortex_m_qemu[f0f1]::lang_items[0]::start[0])``

alexcrichton:
Thanks! Ah yeah it's fine to ignore the thumb targets on CI (feel free to add an allow_failure if you're enterprising!) but looks like other targets may be failing as well?

alexcrichton:
Thanks!

PR #227, opened 2018-02-04
Fix __aeabi_fcmple and add test cases

paoloteti:
`le` in __aeabi_fcmple means `less or equal`.

I don't know why a test case is missing, so let's add it now.

Of course it can fail because `__aeabi_fcmple` is already in some nightly build, so We can't test it against a trusted implementation.


alexcrichton:
Oops, thanks!

PR #228, opened 2018-02-04
Add support for sub*f3vfp and add*f3vfp

paoloteti:
As done before for `mul` and `div` let's use extern "C" to generate `"aapcs"` or `"aapcs-vfp"` depending on target configuration.


alexcrichton:
@bors: r+

Thanks!

PR #229, opened 2018-02-08
Add generic conversion from a narrower to a wider FP type

paoloteti:
Add `extend` module to implement conversion from a narrower to a wider floating-point type.

This implementation is only intended to support *widening* operations.
Module to convert a *narrower* floating-point will be added in the future.


alexcrichton:
@bors: r+

awesome, thanks!

PR #230, opened 2018-02-09
Fix issue extending f32::MIN/MAX to f64 and improve testcrate.

paoloteti:
I was able to trigger an issue extending f32::MAX or f32::MIN to a f64.
Issue was **not** triggered by `testcrate` mainly because f32::MAX/MIN are
not in the list of special values to generate.

This PR fix the issue and improve `testcrate` adding MAX/MIN/MIN_POSITIVE
in the list of special values.

### DIFF
--- a/src/float/extend.rs
+++ b/src/float/extend.rs
@@ -41,7 +41,7 @@ fn extend<F: Float, R: Float>(a: F) -> R where
         let abs_dst: R::Int = a_abs.cast();
         let bias_dst: R::Int = exp_bias_delta.cast();
         abs_result = abs_dst.wrapping_shl(sign_bits_delta);
-        abs_result |= bias_dst.wrapping_shl(dst_sign_bits);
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! 

It looks like though this may have revealed another (maybe unrelated?) failure!

```
---- __subdf3::test stdout ----
	thread '__subdf3::test' panicked at 'assertion failed: `(left == right)`
  left: `-inf`,
 right: `-inf`: inputs (18442240474082181119, 9218868437227405311)', /target/i586-unknown-linux-gnu/release/build/testcrate-f709bcd1a9863f7e/out/generated.rs:32516:13
```

Do you think you may have a change to look into that?

paoloteti:
Yes, I'm going to have a look on it.

alexcrichton:
@bors: r+

Thanks!

PR #232, opened 2018-02-18
Fix CI for thumb* targets

paoloteti:
Main fix is inside `utest` (implement Termination to avoid ICE) all the rest are minor fixes due to the new `testcrate`.

Thumb6m needs some extra fixes: some undefined reference and a panic raised removing assembly implementation not in the list.

Still allow failures on TravisCI and keep my `utest` fork for a while. Let's see if is stable.


testcrate/Cargo.toml
```diff
@@ -14,12 +14,15 @@ rand = { version = "0.4", features = ["i128_support"] }
...
-features = ["mangled-names", "no-lang-items"]
+features = ["no-lang-items"]
 
 [target.'cfg(all(target_arch = "arm", not(any(target_env = "gnu", target_env = "musl")), target_os = "linux"))'.dev-dependencies]
-test = { git = "https://github.com/japaric/utest" }
-utest-cortex-m-qemu = { default-features = false, git = "https://github.com/japaric/utest" }
-utest-macros = { git = "https://github.com/japaric/utest" }
+test = { git = "https://github.com/paoloteti/utest", branch = "ci" }
+utest-cortex-m-qemu = { default-features = false, git = "https://github.com/paoloteti/utest", branch = "ci" }
+utest-macros = { git = "https://github.com/paoloteti/utest", branch = "ci" }
```
alexcrichton: For tracking purposes, are the changes here available as upstream PRs perhaps?

testcrate/Cargo.toml
```diff
@@ -14,12 +14,15 @@ rand = { version = "0.4", features = ["i128_support"] }
...
 
 [features]
 c = ["compiler_builtins/c"]
+mem = ["compiler_builtins/mem"]
+mangled-names = ["compiler_builtins/mangled-names"]
```
alexcrichton: Could this perhaps be turned on by default? I was hoping to keep `cargo test` working in the `testcrate` directory "mostly working by default" if possible

testcrate/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg_attr(feature = "no_std", no_std)]
```
alexcrichton: Could this unconditionally be `#![no_std]`?

alexcrichton: (or this entire file could be emptied)

testcrate/Cargo.toml
paoloteti: Yes it is upstream now, I'll update the URL

testcrate/Cargo.toml
paoloteti: Do you mean 'default = ["mangled-names"]' only ?

paoloteti:
Recent `rust-lang/rust` PR [48143](https://github.com/rust-lang/rust/pull/48143) broke again `utest`. 

If accepted on `utest` upstream PR [2](https://github.com/japaric/utest/pull/2) will fix the issue, so let's wait.

testcrate/Cargo.toml
alexcrichton: Indeed!

alexcrichton:
Ok! Regardless though this all looks good to me, so let's land this and then we can edit .travis.yml to start blocking again in a follow-up?

@bors: r+

Issue #216, opened 2017-12-09
Is there a clever reason why __floatdisf and __floatundisf are unimplemented?


ketsuban:
Looking at the code it seems like they *should* be really easy - all the functions in this group use a dead-simple macro. Is it just that nobody got around to doing it, or are these functions special somehow (e.g. problems a la rust-lang/rust#10184)?

kevinaboos:
I also need these functions, specifically `__floatundisf`. I'm happy to contribute them to this crate for x86_64, but currently I have no idea where to begin. If someone can point me in the right direction as to how to implement them, I can get started.

alexcrichton:
Sure yeah! Currently we mostly just port the implementations from C in the `compiler-rt` project directly to Rust (then add some tests and such). You can find the `compiler-rt` intrinsics in `compiler-rt/lib/builtins` typically, and I normally grep around for the symbol name to find where it's defined

Ketsuban:
Back when I first created this issue I forked the repo and [added these two intrinsics](https://github.com/Ketsuban/compiler-builtins/commit/4c412d273512b4623958a9adb4685b33ef32b1d0) using `__floatdidf` and `__floatundidf` as a template, purely to address my personal need. I didn't make a pull request because I was cognizant of the fact I was engaging in cargo-cult programming, but now there's eyes on this issue I'll make sure everyone knows about it even if only so someone can tell me I was doing it wrong and should feel bad.

PR #234, opened 2018-03-14
Add __floatdisf and __floatundisf intrinsics

ketsuban:
I created this some months back for personal use - I didn't create a PR then because I suspected the actual solution should be more involved than cargo-culting the existing implementation of `__floatdidf` and `__floatundidf` and changing the type, but @alexcrichton said I should be bold.

Resolves issue #216.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,11 +150,11 @@ features = ["c"]
 - [x] fixunssfdi.c
 - [x] fixunssfsi.c
 - [x] floatdidf.c
-- [ ] floatdisf.c
+- [x] floatdisf.c
 - [x] floatsidf.c
 - [x] floatsisf.c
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! Looks like CI may be failing though, but I think that's just due to the need to remove these files getting built from the build script?

alexcrichton:
It also looks like there's an implementation optimized for i386, perhaps that could still be compiled in and used by default?

alexcrichton:
Oh no worries! I think this should be fixed by removing [these two lines](https://github.com/rust-lang-nursery/compiler-builtins/blob/2a2f6d96c8dc578d2474742f14c9bab0b36b0408/build.rs#L177-L178)

PR #235, opened 2018-03-18
i128 is being stabilized

mark-i-m:
cc rust-lang/rust#35118 rust-lang/rust#49101

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,7 @@
 #![feature(linkage)]
 #![feature(lang_items)]
 #![allow(unused_features)]
+#![feature(i128_type)]
 #![no_builtins]
 #![cfg_attr(feature = "compiler-builtins", feature(staged_api))]
 #![cfg_attr(feature = "compiler-builtins",
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,6 @@
 #![feature(compiler_builtins)]
 #![feature(core_intrinsics)]
 #![feature(naked_functions)]
-#![feature(i128_type)]
 #![feature(repr_simd)]
 #![feature(abi_unadjusted)]
 #![feature(linkage)]


src/lib.rs
```diff
@@ -12,7 +12,6 @@
 #![feature(compiler_builtins)]
 #![feature(core_intrinsics)]
 #![feature(naked_functions)]
-#![feature(i128_type)]
```
est31: You'd most likely have to do `#[cfg_attr(stage0, feature(i128_type))]` here, as on stage0 it is still needed.

mark-i-m: Hmm... It seemed to work fine when I made this change locally?

alexcrichton:
Thanks! I think it's ok if this is here, right? Does this fail the build?

alexcrichton:
(in that I'd prefer to land the rustc change first and then land the change here after that's gone through)

PlasmaPower:
It's sort of a deadlock. Rustc fails without this, but this fails without rustc.

alexcrichton:
The "failure" I think is `deny(warnings)`, right? Can that be removed?

PlasmaPower:
Yeah I think that's the best solution. If we want we can re-add if after the rustc change goes through.

mark-i-m:
Moving the `feature(i128_type)` after `allow(unused_features)` seems to make the compiler build for me locally.

kennytm:
We could use the old external tool update workflow here:

1. Push this PR into a new, non-master branch on this repository 
2. Update the submodule in rust-lang/rust#35518 to point to that new branch
3. Wait for rust-lang/rust#35518 to merge and a new nightly is released
4. Merge this PR

PlasmaPower:
That's what I originally suggested in the rustc PR but just ignoring the warning temporarily seems easier.

mark-i-m:
Umm... does anyone understand the current Travis failure:

```
thread 'rustc' panicked at 'slice index starts at 4294967295 but ends at 30759529', libcore/slice/mod.rs:791:5
error: internal compiler error: unexpected panic
note: the compiler unexpectedly panicked. this is a bug.
note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports
note: rustc 1.26.0-nightly (adf2135ad 2018-03-17) running on x86_64-apple-darwin
```

mark-i-m:
It looks like the current change in the PR might be sufficient if we fix the build failures... alternately, I could do @kennytm's suggestion. Just let me know :)

alexcrichton:
Ok this strategy seems fine by me yeah

mark-i-m:
Hmm... this doesn't seem to have worked... I don't understand why.

src/lib.rs
```diff
@@ -12,12 +12,12 @@
...
 #![feature(repr_simd)]
 #![feature(abi_unadjusted)]
 #![feature(linkage)]
 #![feature(lang_items)]
 #![allow(unused_features)]
```
PlasmaPower: Oh this is `#![allow(unused_features)]`, but you need `#[allow(stable_features)]`.

PR #236, opened 2018-03-21
stable_features allowed temporarily

mark-i-m:
cc rust-lang/rust#35118 rust-lang/rust#49101

My second attempt to get this right :/


PlasmaPower:
Appveyor failure is unrelated.

PR #238, opened 2018-03-27
Fix incorrect names used / generated on ARM

### LINKED ISSUE #237: Undefined symbols on arm-ios
Originally reported at https://github.com/TimNN/cargo-lipo/issues/11.

There appear to be several undefined symbols relating to integer definition on `armv7(s)` (iOS) targets.

Steps to reproduce:

Checkout this repo and run:

```sh
cargo build --release --target armv7s-apple-ios
mkdir tmp
cd tmp
ar x ../target/armv7s-apple-ios/release/libcompiler_builtins.rlib
ld -r -o tmp.o *.o
nm -u tmp.o
```

The output for `armv7` and `armv7s` is:

```
___divsi3
___udivsi3
__aeabi_idiv
__divmoddi4
__udivmoddi4
__udivmodsi4
```

Note that the four symbols prefixed  with **two** underscores exist in the binary prefixed with **three** underscores.

TimNN:
Fixes #237.

### DIFF
--- a/src/arm.rs
+++ b/src/arm.rs
@@ -1,7 +1,10 @@
 use core::intrinsics;
 
 // NOTE This function and the ones below are implemented using assembly because they using a custom
-// calling convention which can't be implemented using a normal Rust function
+// calling convention which can't be implemented using a normal Rust function.
+// NOTE The only difference between the iOS and non-iOS versions of those functions is that the iOS
+// versions use 3 leading underscores in the names of called functions instead of 2.
+#[cfg(not(target_os = "ios"))]
 #[naked]
 #[cfg_attr(not(feature = "mangled-names"), no_mangle)]
 pub unsafe fn __aeabi_uidivmod() {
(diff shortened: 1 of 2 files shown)


TimNN:
So, while this fixes 32bit iOS, it apparently breaks `arm-unknown-linux-gnueabi`...

src/arm.rs
```diff
@@ -8,7 +8,7 @@ pub unsafe fn __aeabi_uidivmod() {
     asm!("push {lr}
           sub sp, sp, #4
           mov r2, sp
-          bl __udivmodsi4
+          bl ___udivmodsi4
```
TimNN: Apparently the `gnueabi` targets **don't** need the extra underscore... (while the iOS versions of those targets **do**).

alexcrichton:
Thanks! Looks like CI may be failing for ARM though?

(the windows failure can be ignored)

TimNN:
> Thanks! Looks like CI may be failing for ARM though?

Yes, see the inline comment (https://github.com/rust-lang-nursery/compiler-builtins/pull/238#discussion_r177384304).

I don't know how to best fix this. Should I just duplicate the entire functions, one version for `gnueabi` and one for `iOS`? Do you have any idea what the actual rules are, when to use two and when three underscores?

alexcrichton:
Oh oops sorry I misinterpreted that comment, but yeah I think duplication here should be fine. Unfortunately I personally have no idea about the underscore business :(

alexcrichton:
Thanks!

PR #241, opened 2018-05-12
use system library on openbsd

semarie:
compiler_rt is provided by default on OpenBSD, so use it instead of
rebuilding it from source.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -10,6 +10,13 @@ fn main() {
         return;
     }
 
+    // OpenBSD provides compiler_rt by default, use it instead of rebuilding it from source
+    if target.contains("openbsd") {
+        println!("cargo:rustc-link-search=native=/usr/lib");
+        println!("cargo:rustc-link-lib=static=compiler_rt");
+        return;
+    }
+
     // Forcibly enable memory intrinsics on wasm32 as we don't have a libc to
     // provide them.
     if target.contains("wasm32") {


semarie:
The purpose is to avoid recompiling compiler_rt (what `build.rs` does), and make the crate uses the system library instead of.

I used this way to avoid a missing symbol (from emutls) while compiling rustc (since 1.23.0, I think). It was related to `librustc_binaryen` if I correctly remember. On OpenBSD the symbol is only available via libcompiler_rt, but I couldn't mix the system library with the compiled one from `compiler-builtins` (where emutls isn't compiled).

Without such change, I have the following error while compiling rustc (nightly):
```
rustc:/data/semarie/build-rust/build_dir/build/x86_64-unknown-openbsd/stage1/lib/rustlib/x86_64-unknown-openbsd/codegen-backends/librustc_trans-llvm.so: undefined symbol '__emutls_get_address'
error: couldn't load codegen backend "/data/semarie/build-rust/build_dir/build/x86_64unknown-openbsd/stage1/lib/rustlib/x86_64-unknown-openbsd/codegen-backends/librustc_trans-llvm.so": "Cannot load specified object"
```

Avoiding to compiling the C part at all and using `libcompiler_rt.a` from system solve the issue.

If I didn't mess myself `__rust_probestack` is in the rust part. So it will be compiled and linked with the system library.

alexcrichton:
Ah right yes sorry, got mixed up!

PR #244, opened 2018-05-27
Fix segfault on probestack with dynamic alloca.

qnighy:
When dynamic `alloca` is used, `__rust_probestack` is called even if the allocation is small. This patch fixes the segfault (false stack smash) occuring when dynamic `alloca` attempts to allocate a small fragment of stack memory.

I need it fixed to implement rust-lang/rust#48055.


qnighy:
Added a comment.

alexcrichton:
👍

PR #251, opened 2018-07-27
Implement __mulsi3.

dvc94ch:
Adds __mulsi3 intrinsic for RISCV target required when no hardware multiplication unit is available.

PR adding riscv support to rust: https://github.com/rust-lang/rust/pull/52787

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,6 +58,9 @@ pub mod arm;
 #[cfg(all(kernel_user_helpers, target_os = "linux", target_arch = "arm"))]
 pub mod arm_linux;
 
+#[cfg(any(target_arch = "riscv32"))]
+pub mod riscv32;
+
 #[cfg(target_arch = "x86")]
 pub mod x86;
 
--- a/src/riscv32.rs
+++ b/src/riscv32.rs
@@ -1,0 +1,17 @@
+intrinsics! {
+    // Implementation from gcc
+    // https://raw.githubusercontent.com/gcc-mirror/gcc/master/libgcc/config/epiphany/mulsi3.c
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -58,6 +58,9 @@ pub mod arm;
 #[cfg(all(kernel_user_helpers, target_os = "linux", target_arch = "arm"))]
 pub mod arm_linux;
 
+#[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
```
japaric: NOTE: we'll be submitting a PR to rust-lang/rust that adds these two target architectures.

src/riscv.rs
```diff
@@ -0,0 +1,16 @@
+// Implementation from gcc
+// https://raw.githubusercontent.com/gcc-mirror/gcc/master/libgcc/config/epiphany/mulsi3.c
+#[cfg_attr(not(feature = "mangled-names"), no_mangle)]
+pub unsafe extern "C" fn __mulsi3(mut a: usize, mut b: usize) -> usize {
```
japaric: The signature should be `fn(u32, u32) -> u32` to match the linked source. (iirc, "si" is used for 32-bit ops; "di" is used for 64-bit ops; etc.).

alexcrichton:
Thanks for the PR! Could this use the `intrinsics!` macro like other intrinsics? Other than that looks good to me!

dvc94ch:
@japaric @alexcrichton Is this ok to merge? It's not required by `riscv32imac-unknown-none-elf`, but is if someone builds a target without the +m extension.

alexcrichton:
Thanks @dvc94ch!

PR #255, opened 2018-08-18
Fix __sync_fetch_and_nand_* for pre-v6 ARM

AdrianBunk:
gcc changed semantics for __sync_fetch_and_nand_* in gcc 4.4,
and this was implementing the old semantics:
https://gcc.gnu.org/onlinedocs/gcc-8.2.0/gcc/_005f_005fsync-Builtins.html


alexcrichton:
Thanks!

PR #256, opened 2018-08-20
Support windows/arm target

jordanrh1:
Do not compile assembly files that cannot be compiled with the MSVC toolchain.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -348,7 +348,7 @@ mod c {
             }
         }
 
-        if llvm_target[0] == "armv7" && target_env != "msvc" {
+        if llvm_target[0] == "armv7" /* XXX && target_env != "msvc" */ {
             sources.extend(
                 &[
                     "arm/sync_fetch_and_add_4.S",
--- a/build.rs
+++ b/build.rs
@@ -297,7 +297,7 @@ mod c {
             }
         }
 
-        if target_arch == "arm" && target_os != "ios" {
+        if target_arch == "arm" && target_os != "ios" && target_env != "msvc" {
             sources.extend(
                 &[
                     "arm/aeabi_div0.c",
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -117,7 +117,7 @@ mod c {
 
         cfg.warnings(false);
 
-        if target_env == "msvc" {
+        if target_env == "msvc" && !target_arch_arm {
```
alexcrichton: Hm is this right? This means that if we compile anything it'll use gcc-style flags, right?

build.rs
```diff
@@ -297,7 +297,7 @@ mod c {
             }
         }
 
-        if target_arch == "arm" && target_os != "ios" {
+        if target_arch == "arm" && target_os != "ios" && target_env != "msvc" {
```
alexcrichton: There's some C files below in addition to assembly files, should they be included even on MSVC?

build.rs
jordanrh1: You're right, this change is incorrect.

build.rs
jordanrh1: My understanding is that LLVM uses different builtins when targeting the msvc subsystem, for example `__rt_udiv` instead of `__aeabi_uidivmod`. I don't think the C files are required, and I haven't hit any errors yet due to their absence. I checked a few of the C files and they contain helper code for the assembly code.

alexcrichton: Ok!

build.rs
jordanrh1: Undid the unnecessary changes. Let me know if you'd like me to squash.

alexcrichton:
Thanks!

alexcrichton:
Ah no worries! We're not too strict about the history

PR #259, opened 2018-09-08
Remove superfluous cc flags

paoloteti:
Latest `cc-rs` already manage all arm/thumb flags, so We can safely remove this piece of code.


alexcrichton:
Thanks!

PR #261, opened 2018-09-14
Attempt to fix TravisCI

paoloteti:
- Set `DEBUG_LTO_BUILD_DOESNT_WORK` for `i686-apple-darwin` and `x86_64-apple-darwin` to avoid the undefined reference problem with debug-assertions+lto

- Remove `#[panic_handler]` attribute since this feature is now stable


alexcrichton:
Awesome, thanks! Progress being made works for me :)

Looks like Windows may still have issues, but that's fine to fix at a future time.

PR #263, opened 2018-11-21
Add SGX target

jethrogb:
As suggested by @alexcrichton in https://github.com/rust-lang/rust/pull/56066#discussion_r235078452 and https://github.com/rust-lang/rust/pull/56066#discussion_r235086404.

This also adds f32 versions of acosf, asinf, atan2f, atanf, cbrtf, coshf, expm1f, fdimf, hypotf, log1pf, sinhf, tanf, tanhf to the wasm target. These have recently been implemented. With this change, the conversion functions in https://github.com/rust-lang/rust/blob/b74215a/src/libstd/sys/wasm/cmath.rs#L11-L75 are obsolete and can be swapped out for their f32 versions. I have included that change in my SGX PR for std which updates the compiler-builtins submodule.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -17,9 +17,9 @@ fn main() {
         return;
     }
 
-    // Forcibly enable memory intrinsics on wasm32 as we don't have a libc to
+    // Forcibly enable memory intrinsics on wasm32 & SGX as we don't have a libc to
     // provide them.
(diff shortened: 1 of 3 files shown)


jethrogb:
apple-darwin is also failing on master

alexcrichton:
👍

PR #268, opened 2018-12-27
expose ceil, floor and trunc on sgx-target

akash-fortanix:
The compiler intrinsics will link to these.

Do I need to do this for wasm32 as well? I didn't get any errors when I tried it on the playground. But I am not sure if that suffices.

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -66,6 +66,16 @@ no_mangle! {
     fn tanhf(n: f32) -> f32;
 }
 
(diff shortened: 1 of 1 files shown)


akash-fortanix:
I believe AppVeyor build failures are not caused by this PR.

alexcrichton:
Thanks! For wasm there's native instructions for these so LLVM doesn't lower to these operations, but seems fine to add for sgx!

PR #269, opened 2019-01-03
Fix SGX target_env collision

dingelish:
As discussed in issue [57231](https://github.com/rust-lang/rust/issues/57231) and a merged PR [57243](https://github.com/rust-lang/rust/pull/57243), all `target_env = "sgx"` should be accompanied with a target_vendor check. This commit adds these checks to compiler-builtins.

Signed-off-by: Yu Ding <dingelish@gmail.com>


alexcrichton:
Thanks!

PR #271, opened 2019-01-07
Version Bump for count leading zeros

Lokathor:
We've added a new function. It's my understanding that, given how the Rust community uses SemVer, and since we're at a 0.x.y version at the moment, a new function means we bump the `y` version number.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Jorge Aparicio <japaricious@gmail.com>"]
 name = "compiler_builtins"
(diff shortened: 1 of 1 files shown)


alexcrichton:
Hm I think this may have been from an outdated branch? (hence the merge conflict)

Lokathor:
Oh my I didn't even notice! I thought that was the normal "you can't merge this yourself" message since I didn't have permissions. Let me have a look here.

Lokathor:
I think we're good now

PR #276, opened 2019-03-14
Fix compilation errors for Armv8-M Baseline and Mainline with FPU

hug-dev:
The larger goal of this pull request is to add `thumbv8m.base-none-eabi` and `thumbv8m.main-none-eabihf` targets to the Rust CI so that the artefacts are available via `rustup`
This pull request specifically fixes the compilation errors encountered when compiling for those targets.

On the way, I added some modifications, not strictly related with those goals and that need to be checked:

* removed from exclusion some Arm assembly files for the FPU targets. I tried to assemble those files independantly with `arm-none-eabi-gcc` with the same options used by `cc-rs` and it was working correctly. Maybe I am missing something on why are these files excluded?
* removed the `cfg(not(thumbv6m))` from the intrinsics example as the "real" functions were compiling fine for `thumbv6m` (and then also `thumbv8m.base`)? I suppose that the intrinsics were implemented for those targets after the `cfg` was made. I suppose then that #75 should be closed? Again, maybe I am missing something.

I was not sure how to test it so I compiled the intrinsics example for the Arm targets:
```
cargo build --example intrinsics --features="c compiler-builtins" --target thumbv6m-none-eabi
cargo build --example intrinsics --features="c compiler-builtins" --target thumbv6m-none-eabi
cargo build --example intrinsics --features="c compiler-builtins" --target thumbv7m-none-eabi
cargo build --example intrinsics --features="c compiler-builtins" --target thumbv7em-none-eabi
cargo build --example intrinsics --features="c compiler-builtins" --target thumbv7em-none-eabihf 
```

Also I patched the Rust compiler with my local modified version of `compiler-builtins` and produced the distribution artefacts, successfully, for all the Arm targets again:
```
./x.py dist --target thumbv6m-none-eabi
./x.py dist --target thumbv7m-none-eabi
./x.py dist --target thumbv7em-none-eabihf
./x.py dist --target thumbv7em-none-eabi
./x.py dist --target thumbv8m.main-none-eabi
./x.py dist --target thumbv8m.main-none-eabihf
./x.py dist --target thumbv8m.base-none-eabi
```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -49,10 +49,11 @@ fn main() {
         println!("cargo:rustc-cfg=thumb")
     }
 
-    // compiler-rt `cfg`s away some intrinsics for thumbv6m because that target doesn't have full
-    // THUMBv2 support. We have to cfg our code accordingly.
-    if llvm_target[0] == "thumbv6m" {
-        println!("cargo:rustc-cfg=thumbv6m")
+    // compiler-rt `cfg`s away some intrinsics for thumbv6m and thumbv8m.base because
+    // these targets do not have full Thumb-2 support but only original Thumb-1.
+    // We have to cfg our code accordingly.
+    if llvm_target[0] == "thumbv6m" || llvm_target[0] == "thumbv8m.base" {
+        println!("cargo:rustc-cfg=thumb_1")
     }
 
     // Only emit the ARM Linux atomic emulation on pre-ARMv6 architectures.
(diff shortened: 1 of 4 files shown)


Amanieu:
LGTM, and travis is happy.

alexcrichton:
Thanks @hug-dev!

hug-dev:
Thanks!
Should a release be made prior to my pull request to the Rust repository that will use this PR changes? I could also make the PR first.

alexcrichton:
We can make a release when necessary yeah, but for now you'll probably want to use `[patch]` while developing. When the PR is ready to go we can publish these changes and then update the PR to use the published version

hug-dev:
Cool, will prepare the PR :)

PR #284, opened 2019-05-02
Include `ldexp*` intrinsics on the wasm target

alexcrichton:
Looks like LLVM optimizes programs like:

    fn foo(a: u8) -> f32 {
        2.0f32.powf(a as f32)
    }

to actually invoking `ldexpf`, so let's be sure to include bindings so
there's not undefined symbols.


jethrogb:
Between SGX and Wasm we keep finding things like these... Is there a more structured approach we can take to figuring out the symbols LLVM codegen might pull in?

PR #289, opened 2019-05-14
don't force-on c feature when working in rustc workspace

RalfJung:
libstd turns on `c` in [it's `Cargo.toml`](https://github.com/rust-lang/rust/blob/80e7cde2238e837a9d6a240af9a3253f469bb2cf/src/libstd/Cargo.toml#L52). That should be enough. Is there a way to verify this?

Fixes https://github.com/rust-lang-nursery/compiler-builtins/issues/288

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ mangled-names = []
 no-lang-items = []
 
 # Only used in the compiler's build system
-rustc-dep-of-std = ['c', 'compiler-builtins', 'core']
+rustc-dep-of-std = ['compiler-builtins', 'core']
 
 [[example]]
 name = "intrinsics"


alexcrichton:
FWIW after some digging alloc as well will need to enable the c feature

RalfJung:
I could mirror std's setup -- but care to elaborate why? Shouldn't it be enough to do that in one place?

alexcrichton:
There's a [branch in rustbuild](https://github.com/rust-lang/rust/blob/f59c71eb8ed808347c1e4245b842d673c75daeb6/src/bootstrap/compile.rs#L164-L205) which special cases no-std targets where we only compile the `alloc` crate and its dependencies for those target, not `std` and below. The `std` crate by default enables the C feature of compiler-builtins but the `alloc` crate doesn't currently. It's also fine to just centralize the logic in the `alloc` crate though.

PR #296, opened 2019-05-21
Support linking against system clang libs

infinity0:
Some platforms already have clang runtime libs available. In this case we can link against them directly, rather than rebuilding them (which requires the sources to be available, which has been removed recently).


build.rs
```diff
@@ -434,3 +448,41 @@ mod c {
...
+    pub fn compile(llvm_target: &[&str]) {
+        let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+
+        let llvm_config = env::var("LLVM_CONFIG").expect("LLVM_CONFIG not set");
+        let (subpath, libname) = match target_os.as_str() {
+            "linux" => ("linux", format!("clang_rt.builtins-{}", llvm_target[0])),
+            "macos" => ("darwin", "clang_rt.builtins_osx_dynamic".to_string()),
+            _ => panic!("unsupported target os: {}", target_os),
+        };
+        let cmd = format!("ls -1d $({} --libdir)/clang/*/lib/{}", llvm_config, subpath);
```
alexcrichton: Instead of shelling out to `ls` and such, can this just use `read_dir` and syscalls manually here?

infinity0:
Thanks for the comments, I'll get around to them at some point. But first I need to understand this crate better before proceeding with this PR, since I'm not entirely sure if this approach is viable.

In particular, LLVM does not built this static library (`libclang_rt.builtins-$arch.a`) for some architectures, e.g. [s390x Fedora](https://koji.fedoraproject.org/koji/rpminfo?rpmID=17094785) and [sparc64 Debian](https://packages.debian.org/sid/sparc64/libclang-common-8-dev/filelist). However the equivalent Rust build on those architectures *does* build the C code as rust's own `libcompiler-rt.a` as with all architectures.

Furthermore I am not sure how to actually meaningfully test if any of what I'm doing is actually working. Even if I comment out the `c_system::compile` line to disable system linking, both the build, the tests, and the tests in `testcrate/` *all pass successfully*. So I guess they are falling back to the Rust implementation or something? Also if I change the sources map to have bogus entries, e.g. renaming `__clzdi2` to `__clzdi2XXX`, everything also still works.

alexcrichton:
Yeah the best way to test this would be to run it through rustc's CI or something similar, the CI here in this repository has been really difficult to get right for that reason becaues it's so easy to accidentally pull in the system and/or Rust version. I don't know of a great way to test this.

mati865:
@infinity0 Rust ships with C code compiled from `llvm-project` submodule, see https://github.com/rust-lang/rust/pull/60981/files

Also see last line here:
```
$ rg "compiler_builtins" src/*/Cargo.toml 
src/liballoc/Cargo.toml
15:compiler_builtins = { version = "0.1.10", features = ['rustc-dep-of-std'] }
35:compiler-builtins-mem = ['compiler_builtins/mem']
36:compiler-builtins-c = ["compiler_builtins/c"]
```

infinity0:
@mati865 Yes I know the source C files are shipped as part of the source release, but they don't appear to be *enabled*. The "compiler-builtins-c" feature you indicated is not enabled by default, and none of the other features depending on it within rustc are enabled by default either:

~~~~
$ rg "compiler-builtins-c" src/*/Cargo.toml
src/liballoc/Cargo.toml
36:compiler-builtins-c = ["compiler_builtins/c"]

src/libstd/Cargo.toml
67:compiler-builtins-c = ["alloc/compiler-builtins-c"]
~~~~

It seems that one has to enable the feature "compiler-builtins-c", but I don't see that this is done by default anywhere. For example I don't see this in https://github.com/rust-lang/rust-forge or https://github.com/rust-lang/rust-central-station, and I don't know where else to look. I'm not even sure how one *could* enable those features - e.g. `./x.py build` doesn't appear to take a `--features` flag.

As I said I can still proceed with this PR since I think it's useful, I'm just trying to avoid getting confused by various comments since I only half-know what I'm doing here.

mati865:
@infinity0 look closely at https://github.com/rust-lang/rust/pull/60981/files#diff-fa9668c926a2788f7849514fe3ed66a9. It will enable `compiler-builtins-c` feature if `src/llvm-project/compiler-rt` path exists.

infinity0:
@mati865 Ah ok thanks, that was the missing link, sorry for the confusion. Then I will finish this PR as discussed above and deploy it in Debian as well. Do you know of a good way to test (even manually) whether the optimised C versions are actually being used? Presumably I could objdump libcompiler_builtins.rlib and examine the assembly?

mati865:
>  Do you know of a good way to test (even manually) whether the optimised C versions are actually being used?

I don't know the whole process here, sorry.

infinity0:
OK, so I looked into the generated rlib in a bit more detail and I think the current approach won't work, because `rustc-link-lib` ends up including (for some symbols) both the C version and the Rust version, which will cause a duplicate symbol error at link time. I'll need to filter out the library first, probably using the `ar` crate. This should be straightforward though.

infinity0:
OK, this version should work and I've added a script to manually check for duplicate symbols. We are waiting on mdsteele/rust-ar#14 however.

ci/check_no-duplicate-symbols.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/python3
```
alexcrichton: This doesn't seem to be executed anywhere on CI?

build.rs
```diff
@@ -439,3 +453,150 @@ mod c {
...
+                match files.get(&format!("{}.c.o", sym_)) {
+                    Some(i) => i,
+                    None => match files.get(&format!("{}.S.o", sym_)) {
+                        Some(i) => i,
+                        None => panic!("could not find expected symbol {} in {:?}", sym, &fullpath),
+                    },
+                }
+            };
+            let mut entry = orig.jump_to_entry(i).unwrap();
+            // TODO: ar really should have an append_entry to avoid the clone
```
bjorn3: That clone is cheap (only some int amd the name of the entry)

ci/check_no-duplicate-symbols.sh
infinity0: I thought it would be better to place the script somewhere other than the top-level directory, but then after I moved it into `ci/` I noticed that `run.sh` already has a duplicate symbols check. My script prints out some more details though, to help debug the situation.

infinity0:
> why is ar needed to extract from the archives? How come linking it in with rustc doesn't work?

clang's builtins contains all the symbols, but this crate defines rust versions of some symbols. If we link in clang's builtins directly, we'll link in both the C versions and the rust versions, leading to duplicate symbols errors.

ci/check_no-duplicate-symbols.sh
alexcrichton: Oh sure yeah, but I didn't actually see this run anywhere on CI, and there's also not any documentation as to what this script does, so I wasn't sure why it was added.

alexcrichton:
Oh right sorry, I understand now. Makes sense

ci/check_no-duplicate-symbols.sh
infinity0: Yeah, `ci/` is probably not the most suitable place for it. Shall I move it back up to the top-level directory? Or something like `dev/` or `scripts/`

alexcrichton: Er sorry my point isn't where this should go, it's that I still don't know what this does, how it's supposed to be run, or what it runs on. If it's not run on CI then it's probably best to leave it out for now.

infinity0: OK, I've removed it from this PR.

alexcrichton:
Great! I think though CI may be failing?

infinity0:
I tried to fix it in the latest commit but it seems now CI is not running at all? Are you able to see what's wrong with it?

infinity0:
I just had to rebase to the latest master, but having trouble figuring out the current failure. Will keep trying.

ognevny:
any news on this?

infinity0:
I never finished this, and have no intention to at the moment. Somebody else is welcome to clone my fork and take it over.

tgross35:
I'm going to close this as superseded by https://github.com/rust-lang/compiler-builtins/commit/83da87fbcf17bde3f3d5f8efc33473b10c538965. That patch solves the same problem of allowing reuse of a compiler-rt you have built, but a lot of the responsibility for getting the right path is moved to outside of the crate so our build.rs remains a bit simpler.

Thank you for the work here!

PR #298, opened 2019-06-05
Update libm for fmin/fmax/fminf/fmaxf

varkor:


alexcrichton:
I think the symbols may need to be added [here](https://github.com/rust-lang-nursery/compiler-builtins/blob/master/src/math.rs) as well?

varkor:
Ah, I completely overlooked that file. Fixed.

PR #300, opened 2019-07-13
avoid ptr::write which might panic in debug mode

RalfJung:
See discussion at https://github.com/rust-lang/rust/pull/62103#issuecomment-508981384.


alexcrichton:
I've bumped to 0.1.17 with this change as well

PR #302, opened 2019-07-13
Replace {u,i}128_* lang items with __rust_{u,i}128_* unmangled functions

bjorn3:
The -Zlower-128bit-ops feature is completely broken, as libcore needs
those lang items to compile with this feature, but they are only
provided by compiler_builtins, which itself depends on libcore.
According to rust-lang/rust#58969 the feature never got finished.

This commit removes the associated lang items and replaces them with
normal unmangled functions, when there is no existing intrinsic. This
makes it easier for alternative codegen backends to implement 128bit
integer support.

cc https://github.com/bjorn3/rustc_codegen_cranelift/pull/627

### DIFF
--- a/src/int/addsub.rs
+++ b/src/int/addsub.rs
@@ -90,44 +90,43 @@ where
 impl Subo for i128 {}
 impl Subo for u128 {}
 
-u128_lang_items! {
-    #[lang = "i128_add"]
-    pub fn rust_i128_add(a: i128, b: i128) -> i128 {
-        rust_u128_add(a as _, b as _) as _
+intrinsics! {
+    pub extern "C" fn __rust_i128_add(a: i128, b: i128) -> i128 {
+        __rust_u128_add(a as _, b as _) as _
     }
-    #[lang = "i128_addo"]
-    pub fn rust_i128_addo(a: i128, b: i128) -> (i128, bool) {
-        let mut oflow = 0;
-        let r = a.addo(b, &mut oflow);
-        (r, oflow != 0)
-    }
-    #[lang = "u128_add"]
-    pub fn rust_u128_add(a: u128, b: u128) -> u128 {
-        a.add(b)
-    }
-    #[lang = "u128_addo"]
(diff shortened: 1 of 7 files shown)


alexcrichton:
Thanks! 

Do you know if this works with presumably accompanied changes to the trans backend for rustc? If not I might recommend not landing this here and just using `[patch]` locally to develop against this and make sure it works in conjunction with updates to rustc.

bjorn3:
I plan to remove `-Zlower-128bit-ops` from rustc after this is merged. The llvm codegen backend will likely not use those new `__rust_*` functions, but I want to use them in bjorn3/rustc_codegen_cranelift#627. Removing the lang item definitions should not break anything, as they are only used when `-Zlower-128bit-ops` is passed, which was broken anyway.

alexcrichton:
Ok makes sense! I'm basically asking that things be working before we merge this. I'd prefer to not have to merge this and then publish a number of subsequent bugfix releases. I'm also not sure if there's much benefit to merging ahead of time before it's tested to work, but I could be wrong!

bjorn3:
> I'd prefer to not have to merge this and then publish a number of subsequent bugfix releases.

I understand. I will probably make the rustc changes later this week and test it.

> I'm also not sure if there's much benefit to merging ahead of time before it's tested to work, but I could be wrong!

Not much. bjorn3/rustc_codegen_cranelift#627 is blocked on CraneStation/cranelift#795 anyway.

alexcrichton:
Ok, thanks! Let me know when it's all ship shape and I can merge/publish!

bjorn3:
Sure

bjorn3:
I tested https://github.com/rust-lang/rust/pull/62801 together with this and it seemed to work on x86_64. (didnt have time to run all tests though, but libstd and rustc built)

alexcrichton:
Ok thanks!

alexcrichton:
I've published this as 0.1.18

PR #305, opened 2019-07-23
Add CFI information to __rust_probestack

### LINKED ISSUE #304: `__rust_probestack` needs CFI annotations to assist in backtracing
In a stack overflow situations, under Rust 1.36, gdb fails to backtrace. Also, various other backtracer mechanisms fail to work. The reason is that `__rust_probestack` meddles with the stack register, and gdb is missing the required information in order to find the stack frame which triggered the overflow. Even with `-C debuginfo=1` this still happens, because the relevant assembly code is manually written without the required annotations.

Here how it looks like:
```
>>> bt
#0  0x0000555555576f73 in __rust_probestack () at /cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.14/src/probestack.rs:55
Backtrace stopped: Cannot access memory at address 0x7fffff7fee50
```

Example program `test.rs`:

```rust
fn main() {
    let arr : [u8; 0xf000000] = [0x1; 0xf000000];
    let mut sum : u64 = 0;

    for i in arr.iter() {
        sum += *i as u64;
    }

    println!("{}", sum);
}
```

Observing the assembly code, it is possible to figure out based on `$rdi` and `$rsp` what should be the offset to add to `$rsp` in order to bring `gdb` to the correct analysis.

```
>>> set $rsp = $rsp + 0x7fe000
>>> bt
#0  0x0000555555576f83 in __rust_probestack () at /cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.14/src/probestack.rs:55
#1  0x000055555555835a in test::main () at a.rs:1
```

One workaround that can be implemented by various backtracers (the internal, and the [backtrace](https://docs.rs/backtrace) crate) is automatic hand-coded compensation, by figuring out the state of the `__rust_probestack` function.

Proof-of-concept (assuming `ctx` captured CPU state):

```rust
// `Pseudo`-code

#[cfg(target_arch = "x86_64")]
if symbol_name == "__rust_probestack" {
    let ptr = ctx.rip as *const u8;
    let back = unsafe { std::slice::from_raw_parts(ptr.offset(-6), 6) };
    let mut array = [0u8; 6];
    array.copy_from_slice(&back);

    let matched = {
        if array == [0x81, 0xec, 0x00, 0x10, 0x00, 0x00] {
            let diff = 0x1000 + (ctx.rax - ctx.r11);
            ctx.rsp += diff;
            true
        } else if array == [0x00, 0x77, 0xe4, 0x4c, 0x29, 0xdc] {
            ctx.rsp += ctx.rax;
            true
        } else {
            false
        }
    };

    if matched { // Unwind
        ctx.rip = unsafe { *(ctx.rsp as *const u64) };
    }
}

backtrace_from(&ctx);
```

While I am against this particular workaround in the general sense, one might find it useful as a stop-gap, at least until `compiler-builtins` is fixed.

Related issue: https://github.com/rust-lang/rust/issues/51405

da-x:
In order for GDB and other stack unwinders to correctly backtrace a stack overflow, it needs CFI information in __rust_probestack.

This turns the following backtrace,

```
>> bt
 #0  0x0000555555576f73 in __rust_probestack () at /cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.14/src/probestack.rs:55
Backtrace stopped: Cannot access memory at address 0x7fffff7fedf0
```

Into this:

```
>>> bt
 #0  0x0000555555574e47 in __rust_probestack ()
 #1  0x00005555555595ba in test::main ()
 #2  0x00005555555594f3 in std::rt::lang_start::{{closure}} ()
 #3  0x0000555555561ae3 in std::panicking::try::do_call ()
 #4  0x000055555556595a in __rust_maybe_catch_panic ()
 #5  0x000055555555af9b in std::rt::lang_start_internal ()
 #6  0x00005555555594d5 in std::rt::lang_start ()
 #7  0x000055555555977b in main ()
```

Fixes #304.


da-x:
I've done some work to test this with a patched `rustc`, having `compiler_builtins = { path = "src/libcompiler_builtins" }`, and it seems to work.

src/probestack_x86_64.S
```diff
@@ -0,0 +1,57 @@
...
+.type   __rust_probestack, @function
+__rust_probestack:
+    .cfi_startproc
+    pushq   %rbp
+    .cfi_def_cfa_offset 16
+    .cfi_offset 6, -16
+    movq    %rsp, %rbp
+    .cfi_def_cfa_register 6
+    mov    %rax,%r11
+    // duplicate %rax as we're clobbering %r11
```
bjorn3: Please move this comment one line up. In is about the `mov` above.

build.rs
```diff
@@ -28,6 +28,22 @@ fn main() {
         println!("cargo:rustc-cfg=feature=\"mem\"");
     }
 
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+    if target_os != "windows" {
```
alexcrichton: I think it's probably fine to just check the `target` local variable like the rest of this function

build.rs
```diff
@@ -28,6 +28,22 @@ fn main() {
...
+        let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+        if target_arch == "x86_64" {
+            let cfg = &mut cc::Build::new();
```
alexcrichton: This compilation will want to be in the `c` module as well, and if possible it'd be great to use the same compilation support that all the other C intrinsics use below.

src/probestack.rs
```diff
@@ -41,88 +41,4 @@
 //! probes on any other architecture like ARM or PowerPC64. LLVM I'm sure would
 //! be more than welcome to accept such a change!
 
-#![cfg(not(windows))] // Windows already has builtins to do this
-
-#[naked]
-#[no_mangle]
-#[cfg(all(target_arch = "x86_64", not(feature = "mangled-names")))]
-pub unsafe extern "C" fn __rust_probestack() {
```
alexcrichton: I don't think we'll want to delete these because a C compiler isn't always available and we'll still want definitions of these functions.

src/probestack_x86.S
```diff
@@ -0,0 +1,38 @@
...
+// Unix we're expected to restore everything as it was, this function basically
+// can't tamper with anything.
+//
+// The ABI here is the same as x86_64, except everything is 32-bits large.
+
+.text
+.globl  __rust_probestack
+.type   __rust_probestack, @function
+__rust_probestack:
+    .cfi_startproc
```
alexcrichton: To confirm, are we *certain* that this can't be done with LLVM's inline assembly? For example does updating ebp/rbp work for gdb/backtraces? Does LLVM inline assembly not support the `.cfi_*` directives?

(in general it'll just be easier if we can stick to inline assembly I think)

da-x: I've tried adding the CFI directives both in the original inline assembly, also via `global_asm!`, and both cases they did not result in a binary containing the directives, as observed via `dwarfdump`. Only a separate compilation worked. I haven't tested yet whether `gdb` picks on the `ebp`/`rbp` usage without the directives.

build.rs
da-x: Indeed, it makes sense to refactor that way.

src/probestack.rs
da-x: Actually, we wouldn't really need C because we are compiling assembly files. Passing '.S' files to a C compiler is just wrapping around an assembler. Can we rely solely on our bundled-in LLVM for it somehow?

src/probestack_x86.S
bjorn3: gdb is very picky about what kind of prelude it accepts. Even placing a `rex` prefix before the `push` breaks it: https://github.com/bjorn3/rustc_codegen_cranelift/issues/146#issuecomment-449474527. Because of this, I wouldnt be suprised if using a different register as base pointer confuses gdb.

src/probestack.rs
alexcrichton: Nah unfortunately we don't ship an assembler that uses LLVM, so we're left basically looking for `cc`. You're right that we otherwise just need `as`, but that's not actually that much easier to acquire than a C compiler most of the time.

src/probestack_x86.S
alexcrichton: I think we can use the standard base register prelude here, and if that works I think we probably want to stick with that since we should be able to continue using inline assembly. 

@da-x want to test on your end with the base pointer management added but still using inline assembly and seeing if it works?

da-x: It works as expected. Please consider #306 and close this PR.

alexcrichton:
Indeed I'm gonna close this in favor of https://github.com/rust-lang-nursery/compiler-builtins/pull/306, thanks for investigating!

PR #309, opened 2019-08-19
Support deterministic builds by remapping the __FILE__ prefix if the …

jgalenson:
…compiler supports it.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -423,6 +423,10 @@ mod c {
             panic!("RUST_COMPILER_RT_ROOT={} does not exist", root.display());
         }
 
+        // Support deterministic builds by remapping the __FILE__ prefix if the
+        // compiler supports it.
+        cfg.flag_if_supported(&format!("-ffile-prefix-map={}=.", root.display()));
+
         let src_dir = root.join("lib/builtins");
         for (sym, src) in sources.map.iter() {
             let src = src_dir.join(src);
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! Do you know which file is using `__FILE__`? Or is there perhaps a command line option we could pass like `NDEBUG` which would disable its usage? It'd be interesting if we could get MSVC to be more deterministic here as well.

jgalenson:
The problematic line in compiler-rt is https://github.com/llvm-mirror/compiler-rt/blob/master/lib/builtins/int_util.h#L21, which is used by (at least) absvdi2, absvsi2, absvti2, addvdi3, addvsi3, addvti3, mulvdi3, mulvsi3, mulvti3, negvdi2, negvsi2, negvti2, subvdi3, subvsi3, and subvti3.

There is currently no NDEBUG macro we can use.  An alternative to this patch is of course to add something like it, but since compiler-rt doesn't seem to use NDEBUG at all, I decided to do this.  We can do DNDEBUG in addition to (or instead of) this patch if you would prefer.

alexcrichton:
Nah ok this sounds good to me in that case! Mind including that header file reference in the comment in the source and I'll merge?

jgalenson:
Done.  I suppose it might be worth also trying to get NDEBUG into compiler-rt to support compilers that don't have -ffile-prefix-map, but merging this will help anyway.  Thanks!

alexcrichton:
👍

PR #311, opened 2019-08-20
Implement LLVM's elementwise unordered atomic memory intrinsics

Ralith:
Allows uses of intrinsics of the form llvm.(memcpy|memmove|memset).element.unordered.atomic.* to be linked. Unblocks https://github.com/rust-lang/rust/pull/59155, https://github.com/rust-lang/rust/issues/58599.

These functions are not based on an existing implementation in compiler-rt, because one does not appear to exist. Reviewing [their introduction](https://reviews.llvm.org/D27133), this appears to be because existing users of these intrinsics are not compiler-rt users. Fortunately [the semantics](https://llvm.org/docs/LangRef.html#element-wise-atomic-memory-intrinsics) are simple, so I went ahead and implemented them as the trivial unordered-atomic-ification of the corresponding non-atomic operations.

LLVM calls these functions with [intptr](https://llvm.org/doxygen/classllvm_1_1DataLayout.html#a7a2619d0e489a4ba9c19a0d86a041d59)s rather, so I'm taking a minor liberty by representing them with accurate pointer types. If that's incorrect, it's easy to switch to explicit casts. I'm also assuming that rustc will always generate calls with `usize` for the third argument, which is true for https://github.com/rust-lang/rust/pull/59155.

CC @joshlf

### DIFF
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -5,6 +5,10 @@ type c_int = i16;
 #[cfg(not(target_pointer_width = "16"))]
 type c_int = i32;
 
+use core::intrinsics::{atomic_load_unordered, atomic_store_unordered, unchecked_div};
+use core::mem;
+use core::ops::{BitOr, Shl};
+
 #[cfg_attr(all(feature = "mem", not(feature = "mangled-names")), no_mangle)]
 pub unsafe extern "C" fn memcpy(dest: *mut u8, src: *const u8, n: usize) -> *mut u8 {
     let mut i = 0;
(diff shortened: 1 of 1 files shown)


Ralith:
One thing I'm unclear on is whether this needs to account for targets that don't support a particular width of ` by e.g. `cfg`in things out. I don't see any evidence of that being done elsewhere, but there's not many examples to reference.

CI seems to be failing with linker errors like `unresolved external symbol _ZN4core9panicking5panic17h6e86172dee32b6faE referenced in function _ZN4core6option15Option$LT$T$GT$6unwrap17hb3f9b8d564ddbf0cE`, which is odd, as I don't think I've introduced anything more able to panic than existing code, or involving `Option` at all.

src/mem.rs
```diff
@@ -58,3 +62,110 @@ pub unsafe extern "C" fn memcmp(s1: *const u8, s2: *const u8, n: usize) -> i32 {
...
+        $(
+            #[cfg_attr(not(feature = "mangled-names"), no_mangle)]
+            pub unsafe extern "C" fn $fun($($iid: $ity),+) {
+                $($body)*
+            }
+        )+
+    }
+}
+
+trampoline! {
```
alexcrichton: Could this ideally try to use the existing `intrinsics!` macro?

alexcrichton:
The implementations here all look good to me, although I'm presuming this is all pretty experimental so I'm not necessarily reviewing too too closely. As for the CI errors it looks like this is related to possible panics in debug mode, although I also don't really know how they'd arise. I'd recommend doing what CI is doing (running the build then looking at the rlib) and trying to trace back from there to see why panics are included.

Centril:
cc also @RalfJung

src/mem.rs
Ralith: I was reluctant to do that because `intrinsics!` makes safe functions, while the raw pointer access means they very much are not.

Ralith:
Fixed CI by replacing division with `intrinsics::unchecked_div` and a for loop with a while loop.

src/mem.rs
```diff
@@ -58,3 +62,112 @@ pub unsafe extern "C" fn memcmp(s1: *const u8, s2: *const u8, n: usize) -> i32 {
     }
     0
 }
+
+unsafe fn memcpy_element_unordered_atomic<T: Copy>(dest: *mut T, src: *const T, bytes: usize) {
+    let n = unchecked_div(bytes, mem::size_of::<T>());
```
HadrienG2: I suppose that Bad Things will happen if this is called with a zero-sized T (which could happen by accident in generic code). Is it the responsibility of the caller to refrain from doing that?

Ralith: These functions are only exposed via the monomorphic __llvm_* wrappers, which ensure that `T` is a primitive unsigned integer type.

HadrienG2: Ah, thanks, I missed that.

src/mem.rs
alexcrichton: Ah I think that's ok, basically everything in this library is `unsafe` anyway so I think it's fine for the functions to look safe and they just have an initial `unsafe` block to call the right intrinsic.

alexcrichton: (also I'd prefer to prioritize reducing duplication around mangling and such)

alexcrichton:
Interesting! Was both `unchecked_div` and the for loop required or just the for loop?

src/mem.rs
Ralith: If you're sure. Change applied.

Ralith:
Both are definitely necessary.

alexcrichton:
Thanks!

src/mem.rs
```diff
@@ -58,3 +62,105 @@ pub unsafe extern "C" fn memcmp(s1: *const u8, s2: *const u8, n: usize) -> i32 {
...
+                atomic_store_unordered(dest.add(i), atomic_load_unordered(src.add(i)));
+                i += 1;
+            }
+        }
+    }
+}
+
+fn memset_element_unordered_atomic<T>(s: *mut T, c: u8, bytes: usize)
+where
+    T: Copy + From<u8> + Shl<u32, Output = T> + BitOr<T, Output = T>,
```
RalfJung: This looks odd, seems worth adding a comment explaining how exactly the value `x` being written is computed?

What is the reason this does not match the normal `memset`, which works on `*mut u8`?

src/mem.rs
```diff
@@ -58,3 +62,105 @@ pub unsafe extern "C" fn memcmp(s1: *const u8, s2: *const u8, n: usize) -> i32 {
     }
     0
 }
+
+fn memcpy_element_unordered_atomic<T: Copy>(dest: *mut T, src: *const T, bytes: usize) {
+    unsafe {
+        let n = unchecked_div(bytes, mem::size_of::<T>());
```
RalfJung: Can we assume that `bytes` is a multiple of `size_of::<T>`? If not, what happens when there is rounding going on?

src/mem.rs
Ralith: The difference is that we must atomically store an entire `T` at a time, which would be impossible if we worked byte-at-a-time. Instead, we build up a `T` made out of repeated `c`s, and atomically store that instead. What I'd really have liked here is `T: PrimitiveInteger`, but of course we don't have that.

src/mem.rs
Ralith: Yes, this guaranteed by LLVM:
> [If len is not a positive integer multiple of element_size, then the behaviour of the intrinsic is undefined.](https://llvm.org/docs/LangRef.html#id942)

src/mem.rs
RalfJung: So the spec says that e.g. the setting happens in 4-byte-atomic chunks?

LLVM is allowed to transform adjacent subsequent atomic operations into a single one. Basically what you are doing here is realizing that optimization by hand.

>  What I'd really have liked here is T: PrimitiveInteger, but of course we don't have that.

Fair. Could you add a comment explaining that?
I also realized this function is private, so we actually know it will be called only with unsigned primitive integer types.

src/mem.rs
RalfJung: Seems like you could use https://doc.rust-lang.org/nightly/std/intrinsics/fn.exact_div.html then. Also IMO this is a precondition worth stating in a comment here.

Ralith: Ooh, didn't see that one. Thanks!

src/mem.rs
Ralith: > So the spec says that e.g. the setting happens in 4-byte-atomic chunks?

Yes, writing each complete `T` atomically is what distinguishes this from a regular `memset`.

> LLVM is allowed to transform adjacent subsequent atomic operations into a single one. Basically what you are doing here is realizing that optimization by hand.

Right, but we need to guarantee this happens for correctness, so relying on an optimization wouldn't be appropriate.

eddyb:
Would it be possible to gate this on LLVM version? System LLVM 7 and LLVM 8 versions don't seem to allow building Rust's standard library for a `i686-unknown-linux-gnu` target, e.g. LLVM 8 fails with:
```llvm
 cmpxchg instructions must be atomic.
  %10 = cmpxchg i64* %9, i64 0, i64 0 unordered not_atomic
in function __llvm_memcpy_element_unordered_atomic_8
LLVM ERROR: Broken function found, compilation aborted!
error: could not compile `compiler_builtins`.
```

And LLVM 7 just segfaults, which made https://github.com/rust-lang/rust/pull/70989 confusing to debug (I initially assumed `rustc` was getting miscompiled, but that doesn't seem likely and the LLVM 8 failure explains it).

I will likely attempt to find a target that doesn't have this problem, since it seems easier if possible.

eddyb:
[`i686-unknown-linux-gnu` has](https://github.com/rust-lang/rust/blob/14061868b3960d8a68a079bd276dde85936970ac/src/librustc_target/spec/i686_unknown_linux_gnu.rs#L6):
```rust
    base.max_atomic_width = Some(64);
```

But [`i586-unknown-linux-gnu`](https://github.com/rust-lang/rust/blob/14061868b3960d8a68a079bd276dde85936970ac/src/librustc_target/spec/i586_unknown_linux_gnu.rs) doesn't have that, so there's hope still.

RalfJung:
> System LLVM 7 and LLVM 8 versions don't seem to allow building Rust's standard library for a i686-unknown-linux-gnu target

Even if we have a work-around, isn't that still a pretty serious bug?

RalfJung:
Opened issue: https://github.com/rust-lang/compiler-builtins/issues/356

src/mem.rs
RalfJung: (Resolved in https://github.com/rust-lang/compiler-builtins/pull/312.)

PR #313, opened 2019-09-06
Lokathor tries to use the .gitmodules

Lokathor:
@alexcrichton requested that I try to make a PR for this crate that points at the new libm PR before we land the libm PR.

I don't know if it's set right, but CI will tell.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -9,7 +9,7 @@ fn main() {
     println!("cargo:compiler-rt={}", cwd.join("compiler-rt").display());
 
     // Activate libm's unstable features to make full use of Nightly.
-    println!("rustc-cfg=libm/unstable");
+    println!("cargo:rustc-cfg=feature=\"unstable\"");
 
     // Emscripten's runtime includes all the builtins
     if target.contains("emscripten") {
--- a/build.rs
+++ b/build.rs
@@ -8,6 +8,9 @@ fn main() {
 
     println!("cargo:compiler-rt={}", cwd.join("compiler-rt").display());
 
+    // Activate libm's unstable features to make full use of Nightly.
+    println!("rustc-cfg=libm/unstable");
+
     // Emscripten's runtime includes all the builtins
     if target.contains("emscripten") {
         return;
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "libm"]
 	path = libm
-	url = https://github.com/rust-lang-nursery/libm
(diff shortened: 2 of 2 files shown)


alexcrichton:
One thing this will want to do is to ensure that the `unstable` feature is "activated" for libm, probably by printing a `rustc-cfg` annotation in the build script.

build.rs
```diff
@@ -8,6 +8,9 @@ fn main() {
 
     println!("cargo:compiler-rt={}", cwd.join("compiler-rt").display());
 
+    // Activate libm's unstable features to make full use of Nightly.
+    println!("rustc-cfg=libm/unstable");
```
alexcrichton: Er sorry I'm running out of time today, but I don't really have the time to basically write this PR second-hand unfortunately. This is not the right syntax for enabling the right feature here, and I'd recommend running the build locally and poking around with the output to make sure it's activated correctly.

Lokathor: Oh, oops, I just assumed that the Cargo.toml syntax still applied. I'll give it a look soon.

Lokathor: It seems that the correct syntax is
```rust
println!("cargo:rustc-cfg=feature=\"unstable\"");
```
I missed the `cargo:` at first, and that's my fault, but the [build scripts docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html) just say `rustc-cfg=FEATURE`. which was the confusing part.

So now it's
```rust
println!("cargo:rustc-cfg=feature=\"unstable\"");
```

Which managed to trigger

```rust
#[cfg(feature = "unstable)]
compile_error!("unstable is on.");
```

when I tested it out.

The build doesn't appear to build the libm crate itself, it just directly grabs source files out of the libm source tree, so I guess we just pretend that compiler-builtins also now has a feature named `unstable`?

alexcrichton: Yes that'll basically be how this works, and this looks green so that's good!

alexcrichton:
@Lokathor ok want to revert the update to `.gitmodules`, update the libm hash to one that's upstream, and I can merge?

Lokathor:
Yes, I will try to do that within a few hours.

Lokathor:
Uh, I tried to do the reverse of what I did before and then it suddenly said there was a 400+ commit difference and it refused to merge.

Lokathor:
I think that it's fixed, and if CI passes feel free to merge.

alexcrichton:
👍

PR #317, opened 2019-10-02
Emit `_fltused` on `uefi` targets as a short-term workaround

oli-obk:
fixes https://github.com/rust-lang/rust/issues/62785

### DIFF
--- a/ci/docker/x86_64-unknown-uefi/Dockerfile
+++ b/ci/docker/x86_64-unknown-uefi/Dockerfile
@@ -1,0 +1,4 @@
+FROM ubuntu:18.04
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    gcc libc6-dev ca-certificates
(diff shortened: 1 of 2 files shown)


ci/docker/x86_64-unknown-uefi/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM ubuntu:18.04
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    gcc libc6-dev ca-certificates
```
alexcrichton: Is this intended to be hooked up to CI? (it doesn't happen automatically, extra configuration would be needed)

josephlr: @oli-obk can we do CI configs in a follow up PR, so that this fix can get merged?

pnkfelix:
This is blocking rust-lang/rust#62785

@alexcrichton can you respond to @josephlr 's question regarding handling of CI?

alexcrichton:
I've removed the CI changes for now.

alexcrichton:
I've published 0.1.20 with this change to get pulled into the compiler

PR #306, opened 2019-07-24
probestack: add frame pointers for easier traceback

da-x:
This turns the following backtrace,

```
>> bt
 #0  0x0000555555576f73 in __rust_probestack () at /cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.14/src/probestack.rs:55
Backtrace stopped: Cannot access memory at address 0x7fffff7fedf0
```

To this:

```
>>> bt
 #0  0x0000555555574e47 in __rust_probestack ()
 #1  0x00005555555595ba in test::main ()
 #2  0x00005555555594f3 in std::rt::lang_start::{{closure}} ()
 #3  0x0000555555561ae3 in std::panicking::try::do_call ()
 #4  0x000055555556595a in __rust_maybe_catch_panic ()
 #5  0x000055555555af9b in std::rt::lang_start_internal ()
 #6  0x00005555555594d5 in std::rt::lang_start ()
 #7  0x000055555555977b in main ()
```

PR #328, opened 2019-12-04
Add control flow information to __rust_probestack

tmandry:
This adds control flow information (CFI) on x86 and x86_64 platforms, so that DWARF-based backtraces work when the stack probe fails.

The implementation builds on the work done in #305 and #306. It manages to keep the CFI in inline assembly by defining a "function within a function." This has been done elsewhere; see e.g. [this LibFuzzer code](https://github.com/llvm-mirror/compiler-rt/blob/69445f095c22aac2388f939bedebf224a6efcdaf/lib/fuzzer/FuzzerUtilFuchsia.cpp#L176).

### DIFF
--- a/src/probestack.rs
+++ b/src/probestack.rs
@@ -41,95 +41,149 @@
 //! probes on any other architecture like ARM or PowerPC64. LLVM I'm sure would
 //! be more than welcome to accept such a change!
 
-#![cfg(not(windows))] // Windows already has builtins to do this
+#![cfg(not(feature = "mangled-names"))]
+// Windows already has builtins to do this.
+#![cfg(not(windows))]
+// We only define stack probing for these architectures today.
+#![cfg(any(target_arch = "x86_64", target_arch = "x86"))]
 
-#[naked]
-#[no_mangle]
-#[cfg(all(target_arch = "x86_64", not(feature = "mangled-names")))]
-pub unsafe extern "C" fn __rust_probestack() {
-    // Our goal here is to touch each page between %rsp+8 and %rsp+8-%rax,
-    // ensuring that if any pages are unmapped we'll make a page fault.
-    //
-    // The ABI here is that the stack frame size is located in `%eax`. Upon
-    // return we're not supposed to modify `%esp` or `%eax`.
-    asm!("
-        pushq  %rbp
-        movq   %rsp, %rbp
-
-        mov    %rax,%r11        // duplicate %rax as we're clobbering %r11
-
-        // Main loop, taken in one page increments. We're decrementing rsp by
-        // a page each time until there's less than a page remaining. We're
-        // guaranteed that this function isn't called unless there's more than a
-        // page needed.
-        //
-        // Note that we're also testing against `8(%rsp)` to account for the 8
-        // bytes pushed on the stack orginally with our return address. Using
-        // `8(%rsp)` simulates us testing the stack pointer in the caller's
-        // context.
-
-        // It's usually called when %rax >= 0x1000, but that's not always true.
-        // Dynamic stack allocation, which is needed to implement unsized
-        // rvalues, triggers stackprobe even if %rax < 0x1000.
-        // Thus we have to check %r11 first to avoid segfault.
-        cmp    $$0x1000,%r11
-        jna    3f
-    2:
-        sub    $$0x1000,%rsp
-        test   %rsp,8(%rsp)
-        sub    $$0x1000,%r11
-        cmp    $$0x1000,%r11
-        ja     2b
-
-    3:
-        // Finish up the last remaining stack space requested, getting the last
-        // bits out of r11
-        sub    %r11,%rsp
-        test   %rsp,8(%rsp)
-
-        // Restore the stack pointer to what it previously was when entering
-        // this function. The caller will readjust the stack pointer after we
-        // return.
-        add    %rax,%rsp
-
-        leave
-        ret
-    " ::: "memory" : "volatile");
-    ::core::intrinsics::unreachable();
+extern "C" {
+    pub fn __rust_probestack();
 }
 
-#[naked]
-#[no_mangle]
-#[cfg(all(target_arch = "x86", not(feature = "mangled-names")))]
-pub unsafe extern "C" fn __rust_probestack() {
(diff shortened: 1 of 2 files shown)


src/probestack.rs
```diff
@@ -43,18 +43,32 @@
...
 #[cfg(all(target_arch = "x86_64", not(feature = "mangled-names")))]
-pub unsafe extern "C" fn __rust_probestack() {
+pub unsafe extern "C" fn __rust_probestack_wrapper() {
     // Our goal here is to touch each page between %rsp+8 and %rsp+8-%rax,
     // ensuring that if any pages are unmapped we'll make a page fault.
     //
     // The ABI here is that the stack frame size is located in `%eax`. Upon
     // return we're not supposed to modify `%esp` or `%eax`.
     asm!("
+        .cfi_endproc
```
frobtech: I think the endproc/pushsection/popsection/startproc dance merits a comment explaining why endproc/startproc are needed.

src/probestack.rs
```diff
@@ -43,18 +43,32 @@
...
 #[naked]
 #[no_mangle]
 #[cfg(all(target_arch = "x86_64", not(feature = "mangled-names")))]
-pub unsafe extern "C" fn __rust_probestack() {
+pub unsafe extern "C" fn __rust_probestack_wrapper() {
     // Our goal here is to touch each page between %rsp+8 and %rsp+8-%rax,
     // ensuring that if any pages are unmapped we'll make a page fault.
     //
     // The ABI here is that the stack frame size is located in `%eax`. Upon
     // return we're not supposed to modify `%esp` or `%eax`.
```
frobtech: Note these should be `%rsp` and `%rax`.

src/probestack.rs
```diff
@@ -43,18 +43,32 @@
...
     asm!("
+        .cfi_endproc
+        .pushsection .text.__rust_probestack
+        .p2align 4,,15
+        .globl __rust_probestack
+        .type  __rust_probestack, @function
+    __rust_probestack:
+        .cfi_startproc
         pushq  %rbp
+        .cfi_def_cfa_offset 16
```
frobtech: I'd use `.cfi_adjust_cfa_offset 8` here--that exactly matches the change made by the push insn, rather than freshly encoding the absolute state.

src/probestack.rs
```diff
@@ -43,18 +43,32 @@
...
+        .cfi_offset rbp, -16
```
frobtech: I believe these can use the `%regname` syntax and being consistent with the instructions that way seems preferable.

src/probestack.rs
```diff
@@ -43,18 +43,32 @@
...
+pub unsafe extern "C" fn __rust_probestack_wrapper() {
     // Our goal here is to touch each page between %rsp+8 and %rsp+8-%rax,
     // ensuring that if any pages are unmapped we'll make a page fault.
     //
     // The ABI here is that the stack frame size is located in `%eax`. Upon
     // return we're not supposed to modify `%esp` or `%eax`.
     asm!("
+        .cfi_endproc
+        .pushsection .text.__rust_probestack
+        .p2align 4,,15
```
frobtech: This alignment isn't really required and I wouldn't recommend it.

src/probestack.rs
```diff
@@ -93,7 +107,13 @@ pub unsafe extern "C" fn __rust_probestack() {
         add    %rax,%rsp
 
         leave
+        .cfi_def_cfa rsp, 8
```
frobtech: Again this could be `.cfi_def_cfa_register %rsp`;`.cfi_adjust_cfa_offset -8` to exactly match the relative effect of the instruction rather than encoding knowledge of the absolute state.

tmandry:
Unfortunately the mach-o assembler doesn't support the directives I'm using to keep the assembly inline. I don't know how to express what I want to in that assembler (it doesn't like some of the assembly used in #305 either).

That means we'll have to split the implementations for Mac out :(

tmandry:
Actually, it seems #305's implementation did compile on apple _except_ for the `.type __rust_probestack, @function` directive. I'm not sure what the implication would be of removing it, but if we can do that then it's probably best to go with that approach (which would require using external assembly files).

There doesn't seem to be any substitute for `.pushsection`/`.popsection` in Apple's assembler, and I don't know how to keep the assembly inline without those.

tmandry:
Hmm, so wrapping our inner implementation in `.cfi_endproc`/`.cfi_startproc` is required in debug builds, but fails in release builds. _Not_ doing this succeeds in release builds, but fails in debug builds.

It seems like the enclosing `.cfi_startproc`/`.cfi_endproc` gets "optimized out" in release builds. Not sure what to do about this yet.

src/probestack.rs
```diff
@@ -43,18 +43,94 @@
...
+            // We are about to define a 'function within a function.' Because the
+            // compiler will have emitted a .cfi_startproc at the beginning of
+            // __rust_probestack_wrapper, we need .cfi_endproc before we can define
+            // the contents of __rust_probestack.
+            .cfi_endproc
+
+            .pushsection .text.__rust_probestack
+            .globl __rust_probestack
+            .type  __rust_probestack, @function
+        __rust_probestack:
```
bjorn3: I believe on Mach-O the linker would be allowed to move this symbol and following content to another location, as it uses symbols to determine the start and end of every function.

bjorn3:
Maybe use `global_asm!` instead of inline asm?

src/probestack.rs
tmandry: I think that would be fine. `__rust_probestack` is the function that gets called, not `__rust_probestack_wrapper`.

tmandry:
> Maybe use `global_asm!` instead of inline asm?

I didn't know about `global_asm!`, thanks! Hopefully that can get me out of this startproc/endproc mess.

alexcrichton:
Now that's an impressive set of green :)

I think it'd be good to hook this up into rust-lang/rust before merging though (via `[patch]` most likely). Would you be ok running the test suite locally with this merged into make sure it's all still working at least at a high level?

tmandry:
Done! Had to make one change, now all tests pass on both linux and mac.

alexcrichton:
👍

PR #337, opened 2020-01-14
Don't modify the intrinsic abi for aarch64 windows

danielframpton:
I was running rustc on aarch64-pc-windows-msvc and ran into compilation errors when evaluating constants.

After some hunting, it appeared that 128 bit integer remainder operations were getting the incorrect value (the result of the division, not the remainder).

This behavior can be observed with a simple program that performs remainder operations on the aarch64-pc-windows-msvc target.

This change disables a macro from adjusting some intrinsic operations and resolves the issue I was seeing with remainder, but there may be a better way to resolve the problem.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Jorge Aparicio <japaricious@gmail.com>"]
 name = "compiler_builtins"
-version = "0.1.23"
+version = "0.1.24"
 license = "MIT/Apache-2.0"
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! 

TBH I know very little about this attribute and why it's needed on x86_64 Windows, but hey if it works it works on arm64 windows. If it fixed issues for you I think it's fine to land this, and if it ever crops up again we can dig more into it.

PR #340, opened 2020-02-12
Set probestack visibility to hidden on ELF targets

tmandry:
Fix for https://github.com/rust-lang/rust/issues/68794. Tested across a few different platforms in https://github.com/rust-lang/rust/pull/69045, and it looks like it hasn't broken anything.

cc @Mark-Simulacrum @Amanieu


Mark-Simulacrum:
Once this lands into master and is confirmed by downstream users as both not breaking anything and fixing the bug, we will likely want to consider a beta backport as well. (Though the regression is stable/stable at this point, so it is less important than for a stable/beta regression, in some sense).

PR #344, opened 2020-02-28
aarch64: Exclude FP intrinsics on +nofp or +nosimd

andre-richter:
`AArch64` GCCs exit with an error condition when they encounter any kind of
floating point code if the `nofp` and/or `nosimd` compiler flags have been set.

Therefore, evaluate if those flags are present and set a boolean that causes any
compiler-rt intrinsics that contain floating point source to be excluded for
this target.

This patch prepares https://github.com/rust-lang/rust/pull/68334


alexcrichton:
Thanks!

PR #351, opened 2020-04-29
Switch to using `llvm_asm!` instead of `asm!`

alexcrichton:


Amanieu:
Could you release a new version of compiler-builtins with this change? It is blocking https://github.com/rust-lang/rust/pull/69171.

alexcrichton:
Sure thing, done now!

alexcrichton:
As an FYI as well libs is added to this crate on crates.io (and core too) in case publishes are needed and I'm not around. There's some [instructions to publish in the repo](https://github.com/rust-lang/compiler-builtins/blob/master/PUBLISHING.md) as well. I certainly don't mind continuing to publish, but if I'm unresponsive or busy I'd want to make sure I'm not accidentally blocking anything!

Amanieu:
I tried to do it myself but I am missing direct push access to the repo, which is needed to push a new tag for the version.

alexcrichton:
Oh good point, should be fixed now!

PR #358, opened 2020-05-26
Use crate visibility for traits

tmiasko:
Those traits are an implementation detail of compiler builtins and not intended
for general use. The fact that they are public not only makes it possible to
use them, but also leads to situation where rustc is providing them as a suggestion:

* https://github.com/rust-lang/rust/issues/71090
* https://github.com/rust-lang/rust/issues/71532

### DIFF
--- a/src/float/mod.rs
+++ b/src/float/mod.rs
@@ -13,7 +13,7 @@ pub mod pow;
 pub mod sub;
 
 /// Trait for some basic operations on floats
-pub trait Float:
+pub(crate) trait Float:
     Copy
     + PartialEq
     + PartialOrd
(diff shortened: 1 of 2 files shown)


alexcrichton:
Is there a particular reason for this? Could you explain in the commit or PR message as to why this change is being made?

bjorn3:
The fact that these traits are exported has been useful to me while debugging miscompilations by cg_clif around 128bit int handling. Because these traits are exported, I can just import them after copying an intrinsic provided by compiler-builtins to the test crate. I don't have to copy all the methods provided by these traits too until I actually want to change them.

tmiasko:
I added rationale to pull request description.

alexcrichton:
Ok thanks! This sounds like a reasonable stopgap to me for now until a better solution is worked out. @bjorn3 I think that's not really the intended use of this crate so while it may be helpful I think it's more important to cater to rust-lang/rust's issues it's having right now.

jethrogb:
I think maybe this should've led to a minor version bump due to the public API change?

PR #359, opened 2020-05-27
Manually patch ret instruction for LVI

jethrogb:
In order to land https://github.com/rust-lang/rust/pull/72655 this change needs to be applied.

More info on the change: https://software.intel.com/security-software-guidance/insights/deep-dive-load-value-injection#specialinstructions

### DIFF
--- a/src/probestack.rs
+++ b/src/probestack.rs
@@ -109,7 +109,12 @@ macro_rules! define_rust_probestack {
 //
 // The ABI here is that the stack frame size is located in `%rax`. Upon
 // return we're not supposed to modify `%rsp` or `%rax`.
-#[cfg(target_arch = "x86_64")]
+//
+// Any changes to this function should be replicated to the SGX version below.
+#[cfg(all(
+    target_arch = "x86_64",
+    not(all(target_env = "sgx", target_vendor = "fortanix"))
+))]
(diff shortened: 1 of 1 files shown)


jethrogb:
cc @raoulstrackx

alexcrichton:
To avoid on the `#[cfg]` soup, could this perhaps just have an entirely separate definition for sgx?

jethrogb:
Are you suggesting to duplicate the whole 50-line function? That seems like unnecessary duplication to me and it would also introduce the risk that only one of the two functions would get updated when needed.

alexcrichton:
Yes, that is what I'm suggesting. I'm not sure this function has changed since it was added, and it's easy enough that if a substantial change happens it's mirrored elsewhere. This crate already suffers from lots of `#[cfg]` and trying to conditionalize one instruction makes it even harder to further understand.

src/probestack.rs
```diff
@@ -163,6 +168,69 @@ global_asm!(define_rust_probestack!(
     "
 ));
 
+// This function is the same as above, except that some instructions are
```
jethrogb, line 171: Actually it may be reasonable for SGX to have its own implementation in the future because it likely knows a priori how big the stack is. We'll leave that for a future change though.

PR #360, opened 2020-05-29
Expand wasm32 testing on CI

alexcrichton:
Run the full `run.sh` test script to get full assertions, including that
nothing in the wasm compiler-builtins is panicking. Unfortunately it's
currently panicking, so this is good to weed out!


alexcrichton:
Ok I think https://github.com/rust-lang/libm/pull/244 will fix CI here, so after that's merged I'll land this here.

PR #364, opened 2020-07-08
Fix warnings on latest nightly

josephlr:
https://github.com/rust-lang/rust/pull/72700 caused the existing
`allow(improper_ctypes)` guard to stop working, we now need
`allow(improper_ctypes_definitions)` instead.

We keep the old one to avoid any issues with older nightlies.

Signed-off-by: Joe Richey <joerichey@google.com>


alexcrichton:
👍

PR #365, opened 2020-07-08
Use REP MOVSQ/STOSQ on x86_64

### LINKED ISSUE #339: memcpy and co. are really unoptimized
I've seen the assembly of the memcpy here for both powerpc and wasm (where the one here is always used unless you use the WASI target) and in both cases really simplistic byte wise loops are emitted. I haven't really done any benchmarks but these likely perform worse than more optimized loops that work on 32-bit or so at a time.

josephlr:
Addresses part of #339 (see also https://github.com/rust-osdev/cargo-xbuild/issues/77)

The implementations for these functions are quite simple and (on many recent processors) are the fastest way to implement `memcpy`/`memmove`/`memset`. The implementations [in MUSL](https://git.musl-libc.org/cgit/musl/tree/src/string/x86_64/memcpy.s) and in [the Linux kernel](https://elixir.bootlin.com/linux/v5.9/source/arch/x86/lib/memcpy_64.S#L38) were used for inspiration. 

Benchmarks for the memory functions were also added in this PR, and can be invoked by running `cargo bench --package=testcrate`. [The results of running this benchmarks on different hardware](https://docs.google.com/spreadsheets/d/1H-ubR-xCJWomUYDI9D2JH19BNUD7R9kfkl_OHSv6vMk/edit#gid=1115059537) show that the qword-based variants are almost always faster than the byte-based variants.

While the implementations for `memcmp`/`bcmp` could be made faster though use of SIMD intrinsics, using `rep cmpsb`/`rep cmpsq` makes them slower, so they are left as-is in this PR.

Note that #164 added some optimized versions for memcmp on ARM.


alexcrichton:
Thanks for this!

I forget, but do we already have tests for memset/memcmp/etc? If not, could you add some as part of this PR? 

Additionally, do you have some benchmark numbers for how these perform?

src/mem/x86_64.rs
```diff
@@ -0,0 +1,91 @@
...
+// See "Intel® 64 and IA-32 Architectures Optimization Reference Manual"
+//  - Enhanced REP MOVSB and STOSB operation (ERMSB) (section 3.7.7)
+//  - Fast Short REP MOV (FSRM)
+//
+// While the "rep cmpsb" instructions have not seen similar optimizations, they
+// are still faster than our simple fallback implementations for memcmp/bcmp.
+
+#[inline(always)]
+unsafe fn copy_forward(dest: *mut u8, src: *const u8, count: usize) {
+    asm!(
```
bjorn3, line 15: :sob: cg_clif doesn't support inline asm. Can you add a feature flag to disable `asm!` usage?

josephlr: This isn't the only ASM in this crate, as many of the functions in this crate must me implemented in assembly. How is cg_clif currently building this crate if it doesn't support inline ASM?

Or is this a distinction between `llvm_asm!` and `asm!`

alexcrichton: I think it's ok to not support cg_clif in the first pass, that's not an officially supported backend and it's not at the point where it should be adding maintenance burden to this crate. @bjorn3 you're welcome to have a follow-up PR to update this crate if it stops working.

bjorn3: The other inline asm is only used for stack probes. For this reason I disabled stack probes in Cranelift and made cg_clif recognize the specific inline asm of the stack probes and emit a trap (for `___chkstk_ms`) or nothing (for the `__rust_probestack` `global_asm!`. Combined this means that no inline asm is ever used. This PR would however introduce mandatory inline asm, that I would then need to emulate.

josephlr:
> Additionally, do you have some benchmark numbers for how these perform?

I updated this PR to add `memcpy`/`memset`/`memcmp` benchmarks to the `testcrate` crate. It allows comparing the libc functions (via `copy_from_slice`/`slice.cmp`/etc...) to the Rust functions provided by this crate. 

I ran a [bunch of trials](https://docs.google.com/spreadsheets/d/1H-ubR-xCJWomUYDI9D2JH19BNUD7R9kfkl_OHSv6vMk/edit?usp=sharing), results are below, the main takeaways are:
- Using the `rep movsb/stosb` makes the Rust `memcpy`/`memset` implementation as fast as musl/glibc's
  - About a **70-80%** improvement (for 4K blocks).
  - Given that musl essentially has [an identical implementation](http://git.musl-libc.org/cgit/musl/tree/src/string/x86_64/memcpy.s), this makes sense
- Using `repe cmpsb` for `memcmp` actually **makes things worse**, so I'll remove it.
  - This is consistent with the Intel's optimization guide.
  - The `memcmp` implementation is **very slow**, there's room to improve here.

## memcpy

| Implementation | 4 KiB blocks (GiB/sec) | 1 MiB blocks (GiB/sec) |
|----------|-------------:|------:|
| Current, simple Rust loop |  57.7 | 30.1 |
| **This PR (`rep movsb`)** |    **98.7 (+71%)**  | **37.3 (+24%)** |
| x86_64 Linux musl libc | 94.9 (+64%) | 38.1 (+27%) |
| x86_64 Linux GNU libc | 126.0 (+118%) |    35.7 (+19%) |

## memset

| Implementation | 4 KiB blocks (GiB/sec) | 1 MiB blocks (GiB/sec) |
|----------|-------------:|------:|
| Current, simple Rust loop |  68.6 | 45.3 |
| **This PR (`rep stosb`)** |    **121.7 (+77%)**  | **63.6 (+40%)** |
| x86_64 Linux musl libc | 112.2 (+63%) | 63.8 (+41%) |
| x86_64 Linux GNU libc | 112.7 (+64%) |    63.7 (+41%) |

## memcmp

| Implementation | 4 KiB blocks (GiB/sec) | 1 MiB blocks (GiB/sec) |
|----------|-------------:|------:|
| Current, simple Rust loop |  3.6 | 3.6 |
| **This PR (`rep movsb`)** |    **2.2 (-38%)**  | **2.2 (-37%)** |
| x86_64 Linux musl libc | 3.5 (-1%) | 3.6 (-1%) |
| x86_64 Linux GNU libc | 78.8 (+2110%) | 81.9 (+2182%) |

alexcrichton:
Nice! Thos are some pretty slick wins and also nice find that `memcmp` doesn't speed up all that much. Also that's pretty crazy how much faster glibc is for `memcmp` than a simple loop!

CryZe:
This claims to close my issue, but isn't this only about x86, while the performance problems seem to be happening across the board? (Especially on WASM it seems rather bad atm)

alexcrichton:
We can leave it open for other platforms, but FWIW there's not really much else we can do for wasm. The bulk memory proposal fixes this issue, however, because the `memory.copy` instructions is basically the exact same as `call memcpy` and it implemented by the engine so is much faster.

CryZe:
If you use the WASI target it does a loop copying 32-bit values rather than individual bytes because it uses the musl implementation in the wasilibc then. I haven't done any real benchmarking, but I'd expect that to be faster. But yeah the bulk memory proposal also fixes that issue if you use that target feature.

src/mem/x86_64.rs
```diff
@@ -0,0 +1,69 @@
...
+        "mov ecx, {byte_count:e}",
+        "add rdi, 7",
+        "add rsi, 7",
+        "rep movsb [rdi], [rsi]",
+        "cld",
+        byte_count = in(reg) byte_count,
+        inout("rcx") qword_count => _,
+        inout("rdi") dest_end.sub(8) => _,
+        inout("rsi") src_end.sub(8) => _,
+        options(nostack, preserves_flags)
```
Amanieu, line 49: `add` clobbers the flags.

alexcrichton:
Also, to confirm, do we have tests for this in-repo? If not could some be added?

josephlr:
> Also, to confirm, do we have tests for this in-repo? If not could some be added?

I would want better test coverage than what we currently have. I'm planning to add a bunch before this CL is ready for review again.

src/mem/x86_64.rs
```diff
@@ -0,0 +1,69 @@
+use super::c_int;
+
+// On recent Intel processors, "rep movsb" and "rep stosb" have been enhanced to
```
nagisa, line 3: How does this implementation fare on non-intel implementations of x86_64?

josephlr: I've been investigating performance on AMD hardware (the only other x86 platform where anyone cares about performance). This has led me to modify the implementation. When I have some time, I'll post the results and update this comment to clarify the impact on AMD as well.

src/mem/x86_64.rs
Soveu: Also, it clears the directions flag, which could be set before

Soveu:
These implementations assume that the direction flag is not set, which could not be always the case

src/mem/x86_64.rs
petrochenkov: On Intel fast string support is detectable by two flags in CPUID and one enable bit in IA32_MISC_ENABLE, if necessary. AMD may provide the same detection tools.

src/mem/x86_64.rs
```diff
@@ -0,0 +1,69 @@
...
+    let qword_count = count >> 3;
+    let byte_count = count & 0b111;
+    asm!(
+        "rep stosq [rdi], rax",
+        "mov ecx, {byte_count:e}",
+        "rep stosb [rdi], al",
+        byte_count = in(reg) byte_count,
+        inout("rcx") qword_count => _,
+        inout("rdi") dest => _,
+        in("rax") (c as u8 as u64) * 0b0101010101010101,
```
andyhhp, line 65: This wants to be a hex constant, not a binary one

src/mem/x86_64.rs
josephlr: So it looks like virtually all newish AMD and Intel processors support some sort of "REP MOVS enhancement" (i.e. rep movs is somehow better than a normal loop). However, if the ermsb feature flag isn't present (like on all AMD processors) then `rep movsq` seems better than `rep movsb`. With ermsb the two variants are about the same speed.

Given this, I just implemented the `rep movsq` version unconditionally without any CPUID checking. Variants that use `rep movsb` when Intel's ERMSB/FSRB feature is enabled could be added later, but there doesn't seem to be much of a gain (at least with the benchmarks I'm running here).

src/mem/x86_64.rs
josephlr: Thanks, fixed.

src/mem/x86_64.rs
josephlr: Fixed (due to the add). Note that the direction flag must be cleared before exiting an ASM block, andit doesn't affect `preserves_flags`.

josephlr:
> These implementations assume that the direction flag is not set, which could not be always the case

Per the [`asm!` docs](https://doc.rust-lang.org/beta/unstable-book/library-features/asm.html), "On x86, the direction flag (DF in EFLAGS) is clear on entry to an asm block and must be clear on exit.", so I think these impls are fine.

src/mem/x86_64.rs
josephlr: I gated the changes behind the `asm` feature flag.

josephlr:
For AMD performance, I'm getting some conflicting results about which is better `rep movsb` or `rep movsq`. I think it might be because I'm using a VM. I don't have any real AMD CPU hardware, could anyone in this thread run the benchmarks against https://github.com/rust-lang/compiler-builtins/pull/365/commits/dd7b7dc8f247c2d7eaae299b63cc99c61b62a344 and https://github.com/rust-lang/compiler-builtins/pull/365/commits/2f9f61ff35563ca52dd1138c612a8fc1620fd189 and tell me what they find?

andyhhp:
> DF

Even C doesn't tolerate DF being set generally.  There are two legitimate uses of it which I have encountered.  One is memmove(), and one is code dumps in backtraces, where you need to be wary of hitting page/permission boundaries (see https://github.com/xen-project/xen/blob/master/xen/arch/x86/traps.c#L175-L204).

For performance, things are very tricky, and once size does not fit all.  Presumably here we're talking about `mem*()` calls which have survived through LLVM's optimisations passes, and are the variations which don't decompose nicely?

If alignment information is available at compile time, then `rep stos{l,q}` is faster than `rep stosb` on earlier hardware.  Intel have some forthcoming features literally named `Fast Zero-Length MOVSB`, `Fast Short STOSB`, `Fast short CMPSB/SCASB` (https://software.intel.com/content/dam/develop/external/us/en/documents/intel-tdx-module-1eas.pdf Page 120.  Not sure if this was intended to be public right now, but it is.) which should give anyone a hint that the current variations aren't great for small `%ecx` inputs.

Frankly, study a popular libc and follow their lead.  A **lot** of time and effort has gone into optimising them generally across multiple generations of processor.  Alternatively, if you do feel like doing feature-based dispatch, that will get better results if you can pick the optimum algorithm for the CPU you're on.

Soveu:
Linux encourages to use `rep movsb/stosb` with memcpy/memset
[memcpy](https://github.com/torvalds/linux/blob/1e6d1d96461eb350a98c1a0fe9fd93ea14a157e8/tools/arch/x86/lib/memcpy_64.S#L50)
[memset](https://github.com/torvalds/linux/blob/1e6d1d96461eb350a98c1a0fe9fd93ea14a157e8/tools/arch/x86/lib/memset_64.S#L48)

josephlr:
I did some final benchmarking of various approaches [here are the results](https://docs.google.com/spreadsheets/d/1H-ubR-xCJWomUYDI9D2JH19BNUD7R9kfkl_OHSv6vMk/edit?usp=sharing). Long story short, the `rep movsq`/`rep stosq` are faster the their byte-based equivalents. This is especially notable on AMD where it's not even close (6x faster). The byte-based instructions were only faster in two senarios:
  - `memcpy`, in the forward direction, on Intel hardware with the `erms` capability.
  - `memset`, in the forward direction, on Intel hardware with the `erms` capability.

The PR title/description and the comments have been updated. The only thing left is to add tests for the memory functions.

Thanks for the feedback @andyhhp, comments inline.

> Presumably here we're talking about `mem*()` calls which have survived through LLVM's optimisations passes, and are the variations which don't decompose nicely?

Correct. This is also why we aren't really concerned about short `memcpy`s here, as they are usually directly inlined by LLVM.

> If alignment information is available at compile time, then `rep stos{l,q}` is faster than `rep stosb` on earlier hardware.

Yes, this is true on essentially all AMD hardware and all Intel hardware without ERMSB ("Enhanced REP MOVSB/STOSB").

> Intel have some forthcoming features literally named `Fast Zero-Length MOVSB`, `Fast Short STOSB`, `Fast short CMPSB/SCASB` (https://software.intel.com/content/dam/develop/external/us/en/documents/intel-tdx-module-1eas.pdf Page 120. Not sure if this was intended to be public right now, but it is.)

It looks like the TDX manual has the CPUID listing for Sapphire Rapids, and that the main Intel manual hasn't been updated yet. I would imagine these features will appear in the main SDM once it gets updated. I've update the file comment to also mention those features. 

> Frankly, study a popular libc and follow their lead. A **lot** of time and effort has gone into optimising them generally across multiple generations of processor.

Oh I definitely realize this. This PR is based off the implementations [in MUSL](https://git.musl-libc.org/cgit/musl/tree/src/string/x86_64/memcpy.s) and in [the Linux kernel](https://elixir.bootlin.com/linux/v5.9/source/arch/x86/lib/memcpy_64.S#L38) . I've updated the comments/descriptions to make this more clear (I referenced it in the comments). As @Soveu noted above, the kernel prefers `rep movsb` over `rep movsq`, but _only_ when the `erms` feature is present. The above results would be seem to validate that approach.

> Alternatively, if you do feel like doing feature-based dispatch, that will get better results if you can pick the optimum algorithm for the CPU you're on.

We may do compile-time feature dispatch in a followup CL (if people care). It would basically switch to the `rep movsb`-only variants if `#[cfg(target_feature = "ermsb")]`.

Soveu:
Are you sure you have been benchmarking it properly?
[gamozo found very interesting patterns in `rep movsb` behaviour](https://www.twitch.tv/videos/758359144?t=06h02m59s), that are reproducible on my sandybridge laptop (tommorow i'll test it on ryzen)

josephlr:
> Are you sure you have been benchmarking it properly?

The benchmark code is in this PR, let me know if you find any issues. The results I found above seem consistent with what the Intel/AMD optimization manuals would suggest. Note that the benchmarks here are not intended to be fully comprehensive, just to give a rough idea of the perf.

> [gamozo found very interesting patterns in `rep movsb` behaviour](https://www.twitch.tv/videos/758359144?t=06h02m59s), that are reproducible on my sandybridge laptop (tommorow i'll test it on ryzen)

As they mention, it seems mostly like a cache effect, so I'm not sure how much the actual memcpy implementation matters here, but it looks like going with `rep movsq` is the right move.

Soveu:
idk if i messed up something with constants or something, ryzen 2500u performance looks at _best_ like `movsb` from sandybridge

josephlr:
> idk if i messed up something with constants or something, ryzen 2500u performance looks at _best_ like `movsb` from sandybridge
> movsb is as fast as movsq

Was that with the benchmarks in this PR?

Soveu:
No, I use my own code to measure the impact of alignment on performance and to visualize it with gnuplot

Soveu:
These are results from your benchmarks
```
running 16 tests
test memcmp_builtin_1048576  ... bench:      27,377 ns/iter (+/- 594)
test memcmp_builtin_4096     ... bench:          78 ns/iter (+/- 2)
test memcmp_rust_1048576     ... bench:     369,738 ns/iter (+/- 12,626)
test memcmp_rust_4096        ... bench:       1,187 ns/iter (+/- 21)
test memcpy_builtin_1048576  ... bench:      26,713 ns/iter (+/- 7,254)
test memcpy_builtin_4096     ... bench:          77 ns/iter (+/- 1)
test memcpy_rust_1048576     ... bench:      90,658 ns/iter (+/- 1,872)
test memcpy_rust_4096        ... bench:         102 ns/iter (+/- 0)
test memmove_builtin_1048576 ... bench:      26,674 ns/iter (+/- 542)
test memmove_builtin_4096    ... bench:          77 ns/iter (+/- 1)
test memmove_rust_1048576    ... bench:      60,246 ns/iter (+/- 1,063)
test memmove_rust_4096       ... bench:         269 ns/iter (+/- 3)
test memset_builtin_1048576  ... bench:      19,357 ns/iter (+/- 464)
test memset_builtin_4096     ... bench:          75 ns/iter (+/- 1)
test memset_rust_1048576     ... bench:      20,946 ns/iter (+/- 259)
test memset_rust_4096        ... bench:         106 ns/iter (+/- 1)
```

josephlr:
@Soveu and what do you get with the 'rep movsb' implementation on AMD. I.e. running the benchmarks against https://github.com/rust-lang/compiler-builtins/pull/365/commits/2a0132cfae362f429b6da7b5ba206ee20ec9aa1f

Soveu:
```
running 16 tests
test memcmp_builtin_1048576  ... bench:      27,336 ns/iter (+/- 793)
test memcmp_builtin_4096     ... bench:          78 ns/iter (+/- 2)
test memcmp_rust_1048576     ... bench:     585,392 ns/iter (+/- 3,531)
test memcmp_rust_4096        ... bench:       2,304 ns/iter (+/- 9)
test memcpy_builtin_1048576  ... bench:      29,355 ns/iter (+/- 482)
test memcpy_builtin_4096     ... bench:          76 ns/iter (+/- 1)
test memcpy_rust_1048576     ... bench:      89,706 ns/iter (+/- 2,593)
test memcpy_rust_4096        ... bench:          96 ns/iter (+/- 1)
test memmove_builtin_1048576 ... bench:      26,435 ns/iter (+/- 337)
test memmove_builtin_4096    ... bench:          78 ns/iter (+/- 1)
test memmove_rust_1048576    ... bench:     434,864 ns/iter (+/- 10,232)
test memmove_rust_4096       ... bench:       1,732 ns/iter (+/- 26)
test memset_builtin_1048576  ... bench:      19,934 ns/iter (+/- 149)
test memset_builtin_4096     ... bench:          75 ns/iter (+/- 1)
test memset_rust_1048576     ... bench:      21,058 ns/iter (+/- 283)
test memset_rust_4096        ... bench:         101 ns/iter (+/- 0)
```

Soveu:
i wonder why memmove got faster than memcpy lol

josephlr:
@alexcrichton the tests have been added so this is now ready for final review and merging. 

This implementation sticks with the `rep movsq`/`rep stosq` implementation used by MUSL and Linux (see the links in the PR description). The final assembly [looks optimal](https://rust.godbolt.org/z/a8cnxW) (`memcpy`/`memset` are identical to Linux's implementation).

For performance numbers, see my link in https://github.com/rust-lang/compiler-builtins/pull/365#issuecomment-710842993

alexcrichton:
Thanks again for this! This all looks great to me. As one final thing, though, I'm not sure if the `asm` feature is actually ever enabled on CI, so could you add a line [here](https://github.com/rust-lang/compiler-builtins/blob/bc235bccb9a34a7f9ceca226981922b0c31780e4/ci/run.sh#L10-L14) to test the feature?

testcrate/benches/mem.rs
```diff
@@ -0,0 +1,154 @@
...
+extern crate test;
+use test::{black_box, Bencher};
+
+extern crate compiler_builtins;
+use compiler_builtins::mem::{memcmp, memcpy, memmove, memset};
+
+fn memcpy_builtin(b: &mut Bencher, n: usize) {
+    let v1 = vec![1u8; n];
+    let mut v2 = vec![0u8; n];
+    b.iter(|| {
```
nagisa, line 12: minor nit: set `b.bytes = n as u64;`?

nagisa (approved):

testcrate/benches/mem.rs
josephlr: Nice, that's much better than doing the math normally.

josephlr:
> Thanks again for this! This all looks great to me. As one final thing, though, I'm not sure if the `asm` feature is actually ever enabled on CI, so could you add a line [here](https://github.com/rust-lang/compiler-builtins/blob/bc235bccb9a34a7f9ceca226981922b0c31780e4/ci/run.sh#L10-L14) to test the feature?

Done (for tests and builds), also the testcrate enables the `"asm"` feature by default, should that be changed?

alexcrichton:
Hm yeah ideally that would change but that's probably best left to another PR, thanks again!

PR #372, opened 2020-08-14
Add uefi arch x86 probestack support

xiaoyuxlu:
1. In UEFI x86 arch, probestack need triple underscore.

2.  In UEFI, probestack function do things like _chkstk(in MSVC).
    MSVC x32's _chkstk and cygwin/mingw's _alloca adjust %esp themselves
    MSVC x64's __chkstk and cygwin/mingw's ___chkstk_ms do not adjust
    %rsp themselves.
    But current probestack doesn't adjust esp. And LLVM doesn't generate
    sub %eax, %esp after probestack. So we adjust esp in probestack like
    MSVC x32's _chkstk.

### DIFF
--- a/src/probestack.rs
+++ b/src/probestack.rs
@@ -76,7 +76,7 @@ macro_rules! define_rust_probestack {
     };
 }
 
-#[cfg(target_os = "uefi")]
+#[cfg(all(target_os = "uefi", target_arch = "x86_64"))]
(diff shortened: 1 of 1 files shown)


jyao1:
This only impacts i686-unknown-uefi target.

The issue can be reproduced when a function has a big local variable.

PR #380, opened 2020-10-03
Division tweaks

AaronKutch:
This slims down `asymmetric.rs`, removes the old style signed division macro code, and constructs signed divisions independent of the `specialized_div_rem` module. I figured out that when this is done, LLVM autoinlines just how I want it (except that the code gen for RISCV is still horrendous, I guess my `leading_zeros` changes still have not made it into nightly). The fact that I don't have to manage the signed divisions in the `specialized_div_rem` module will make it much easier to convert some macros to generics later. In another PR, I will fix problems I have with the division testing code and improve the edge cases that it covers.

### DIFF
--- a/src/int/leading_zeros.rs
+++ b/src/int/leading_zeros.rs
@@ -4,6 +4,7 @@
 // Compilers will insert the check for zero in cases where it is needed.
 
 /// Returns the number of leading binary zeros in `x`.
+#[doc(hidden)]
 pub fn usize_leading_zeros_default(x: usize) -> usize {
     // The basic idea is to test if the higher bits of `x` are zero and bisect the number
     // of leading zeros. It is possible for all branches of the bisection to use the same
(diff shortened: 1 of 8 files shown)


src/int/leading_zeros.rs
```diff
@@ -4,6 +4,7 @@
 // Compilers will insert the check for zero in cases where it is needed.
 
 /// Returns the number of leading binary zeros in `x`.
+#[doc(hidden)]
 pub fn usize_leading_zeros_default(x: usize) -> usize {
```
bjorn3, line 8: Can this be `pub(crate)` instead of`#[doc(hidden)]`?

AaronKutch: It is being used by the testcrate (the CI does not have a RISCV target at the moment, so we want to make sure that they both work)

Issue #381, opened 2020-10-04
Build failure related to probestack with custom Windows target


epakskape:
I'm working on a project that involves creating a custom Windows target. When attempting to build my crate, I encounter the following error related to probestack:

```
active toolchain
----------------

nightly-x86_64-pc-windows-msvc (default)
rustc 1.49.0-nightly (25c8c53dd 2020-10-03)

D:\repro>cargo build -Z build-std=core --target .cargo\x86_64-sys-windows-msvc.json
   Compiling compiler_builtins v0.1.35
   Compiling repro v0.1.0 (D:\repro)
error: unknown directive
  |
note: instantiated into assembly here
 --> <inline asm>:2:13
  |
2 |             .pushsection .text.__rust_probestack
  |             ^

error: expected absolute expression
  |
note: instantiated into assembly here
 --> <inline asm>:4:20
  |
4 |             .type  __rust_probestack, @function
  |                    ^

error: unknown directive
  |
note: instantiated into assembly here
 --> <inline asm>:5:13
  |
5 |             .hidden __rust_probestack
  |             ^

error: unknown directive
   |
note: instantiated into assembly here
  --> <inline asm>:57:13
   |
57 |             .size __rust_probestack, . - __rust_probestack
   |             ^

error: unknown directive
   |
note: instantiated into assembly here
  --> <inline asm>:58:13
   |
58 |             .popsection
   |             ^

error: aborting due to 5 previous errors

error: could not compile `compiler_builtins`

To learn more, run the command again with --verbose.
```

The issue seems to be caused by an incorrect directive in probestack.rs which results in the probestack code being built:

```rust
// Windows already has builtins to do this.
#![cfg(not(windows))]
```

Which I believe should be:

```rust
#![cfg(not(target_os = "windows"))]
```

Making the above change allows the crate to build successfully.

I created a simple crate to help with reproducing this issue and verifying the fix. You should be able to follow these steps to reproduce:

```
git clone https://github.com/epakskape/bug_repro
cd bug_repro\compiler_builtins_custom_win_target\repro
cargo build -Z build-std=core --target .cargo\x86_64-sys-windows-msvc.json
```

[...]

PR #382, opened 2020-10-04
Fix the Windows target os check for probestack

epakskape:
Addresses #381

### DIFF
--- a/src/probestack.rs
+++ b/src/probestack.rs
@@ -43,7 +43,7 @@
 
 #![cfg(not(feature = "mangled-names"))]
 // Windows already has builtins to do this.
-#![cfg(not(windows))]
+#![cfg(not(target_os = "windows"))]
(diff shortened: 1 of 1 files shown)


alexcrichton:
AFAIK `windows` and `target_os = "windows"` are synonymous, is this perhaps a case where the custom target spec needs an updated?

epakskape:
The custom spec is here. As far as I know I filled out all of the required fields -- is there one that I missed?

https://github.com/epakskape/bug_repro/blob/master/rust/compiler_builtins_custom_win_target/repro/.cargo/x86_64-sys-windows-msvc.json

epakskape:
You were right. I dug into this a bit more, and it turns out the custom spec also needs to set "target_family" to "windows" for this specific cfg check to work as expected. I'll go ahead and close this PR as it's not needed.

PR #389, opened 2020-10-24
math: add {fmin,fmax}{f,} for thumb*-none-eabi*

### LINKED ISSUE #354: ARM Thumb targets need fmax*/fmin* intrinsics
`core::f32` and `core::f64` expose `max` and `min`, but these use the `fmax*`/`fmin*` intrinsics, which aren't exposed for the Thumb targets https://github.com/rust-lang/compiler-builtins/blob/cde22bc180391e75de1c189fe29f442ada86ccde/src/math.rs#L92-L99

https://rust.godbolt.org/z/B4tB6w

```rust
#![no_std]

pub fn maxf(a: f32, b: f32) -> f32 {
    a.max(b)
}

pub fn minf(a: f32, b: f32) -> f32 {
    a.min(b)
}

pub fn max(a: f64, b: f64) -> f64 {
    a.max(b)
}

pub fn min(a: f64, b: f64) -> f64 {
    a.min(b)
}
```

```asm
example::maxf:
        b       fmaxf

example::minf:
        b       fminf

example::max:
        b       fmax

example::min:
        b       fmin
```

jordens:
These are exposed in core::f32

close #354 
c.f. rust-lang/rust#62729
Patch from @whitequark (https://paste.debian.net/1168430/)

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -92,6 +92,10 @@ no_mangle! {
 // only for the thumb*-none-eabi* targets
 #[cfg(all(target_arch = "arm", target_os = "none"))]
 no_mangle! {
+    fn fmin(x: f64, y: f64) -> f64;
+    fn fminf(x: f32, y: f32) -> f32;
+    fn fmax(x: f64, y: f64) -> f64;
+    fn fmaxf(x: f32, y: f32) -> f32;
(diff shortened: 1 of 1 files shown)


alexcrichton:
👍

PR #392, opened 2020-11-03
Use REP MOVSB/STOSB when the ERMSB feature is present

josephlr:
Followup to #365, part of addressing #339 

The first commit reorganizes the `mem` functions to reduce the amount of platform specific code. Specifically, the following `#[inline(always)]` Rust functions:
  - `impls::copy_forward`
  - `impls::copy_backward`
  - `impls::set_bytes`

The second commit uses the functionality added in https://github.com/rust-lang/rust/pull/78396 to check if the `"ermsb"` feature is enabled at compile-time (as we can't detect the features at runtime). If so, we use `REP MOVSB/STOSB` instead of `REP MOVSQ/STOSQ`. This is [what Linux does](https://github.com/torvalds/linux/blob/1e6d1d96461eb350a98c1a0fe9fd93ea14a157e8/arch/x86/lib/memcpy_64.S#L35).

The final commit adds `memcpy`/`memset` tests that are not 8-byte aligned.

## Performance

See the "ERMSB" tab of [this spreadsheet](https://docs.google.com/spreadsheets/d/1H-ubR-xCJWomUYDI9D2JH19BNUD7R9kfkl_OHSv6vMk/edit#gid=1449062845). For Intel chips with the `"ermsb"` feature, we get a **25%-28%** performance improvement against the existing implementation, reaching parity with glibs's `memcpy` and `memset`.

Note this improvement only seems to help for small-ish (4 KiB) copies, 1 MiB copies show no effect from the change.

Also note that this feature doesn't exist on AMD CPUs, so we don't have to worry abound performance degradation on those chips.


alexcrichton:
This looks great to me, thanks for this!

Out of curiosity, what x86 targets do we have that end up using this?

josephlr:
> Out of curiosity, what x86 targets do we have that end up using this?

My use-case is custom `no_std` x86 targets, I tested by just using `target-cpu=native`. I don't think very many of our builtin-targets use this (as the `mem` functions are usually provided by the libc).

I think the SGX targets, which [do use these `mem` functions](https://github.com/rust-lang/compiler-builtins/blob/63c0091a6132a5b732ca425e816e67a66b004754/build.rs#L29), could set `+ermsb` [in their target `features`](https://github.com/rust-lang/rust/blob/56293097f7f877f1350a6cd00f79d03132f16515/compiler/rustc_target/src/spec/x86_64_fortanix_unknown_sgx.rs#L65), similar to what they already do for `+rdrnd` and `+rdseed`. ERMSB was introduced in Ivy Bridge, while SGX was introduced in Skylake.

PR #393, opened 2020-11-19
fix division on SPARC

### LINKED ISSUE #390: Build failure with LLVM ERROR on `sparc-unknown-linux-gnu`
The libc build with cargo's `build-std` fails due to:

```
2020-10-27T12:22:33.1169362Z LLVM ERROR: unable to allocate function return #6
2020-10-27T12:22:33.1212572Z error: could not compile `compiler_builtins`
```

(Here's the log: https://github.com/rust-lang/libc/runs/1314551399)

I recently updated `compiler_builtins` to 0.1.36 (https://github.com/rust-lang/rust/pull/78209) on rust-lang/rust and turned out that's the cause.
We can reproduce the failure with `cargo build -Z build-std=core,alloc -vv --target sparc-unknown-linux-gnu` on this repo as well.

AaronKutch:
fixes #390. I decided to make an entire configuration path for all SPARC architectures, because SPARC does not have any widening multiplications which makes `_delegate` insanely slow. I manually tested the new `u128_divide_sparc` function, but plan on testing it later with the new test system in PR #384. This PR needs to be merged first because 32-bit SPARC will not build without this. I will rebase #384 afterwards to test `u128_divide_sparc` properly.

### DIFF
--- a/src/int/specialized_div_rem/mod.rs
+++ b/src/int/specialized_div_rem/mod.rs
@@ -46,6 +46,7 @@ mod binary_long;
 
 #[macro_use]
 mod delegate;
+pub use self::delegate::u128_divide_sparc;
 
 #[macro_use]
 mod trifecta;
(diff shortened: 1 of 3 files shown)


AaronKutch:
Is the CI broken?

src/int/udiv.rs
```diff
@@ -68,3 +72,31 @@ intrinsics! {
         quo_rem.0
     }
 }
+
+#[cfg(any(target_arch = "sparc", target_arch = "sparc64"))]
+intrinsics! {
+    #[win64_128bit_abi_hack]
+    /// Returns `n / d`
+    pub extern "C" fn __udivti3(n: u128, d: u128) -> u128 {
+        u128_divide_sparc(n, d, &mut 0)
```
alexcrichton, line 81: Could this `#[cfg]` happen inside the previous `intrinsics!` block to avoid duplicating the definition?

src/int/specialized_div_rem/mod.rs
```diff
@@ -77,8 +74,22 @@ const USE_LZ: bool = if cfg!(target_feature = "thumb-mode") {
...
+const USE_LZ: bool = cfg!(target_feature = "b");
+
+// All other common targets Rust supports should have CLZ instructions
+#[cfg(not(any(
+    target_arch = "arm",
+    target_arch = "sparc",
+    target_arch = "sparc64",
+    target_arch = "riscv32",
+    target_arch = "riscv64"
+)))]
```
alexcrichton, line 92: Instead of conditional definitions of this constant could it use `||` and `cfg!`?

alexcrichton (commented):
Thanks! I'm trying to fix CI in https://github.com/rust-lang/compiler-builtins/pull/394

src/int/specialized_div_rem/mod.rs
AaronKutch: good idea

src/int/udiv.rs
AaronKutch: I changed it to use `#[cfg]` inside the previous block, but I don't know if it is beneficial because of all the duplicated `#[cfg]` that is required to completely disable `u128_div_rem`

alexcrichton:
👍

Issue #398, opened 2020-12-19
segmentation fault for `aarch64-unknown-none-softfloat` on 0.1.38


JohnTitor:
Found in https://github.com/rust-lang/rust/pull/79863#issuecomment-743935206:
```
rustc exited with signal: 11
error: could not compile `compiler_builtins`

Caused by:
[...]

PR #399, opened 2020-12-26
Remove `count_ones`

AaronKutch:
I can't remember why I added this in the first place since it isn't actually used. This fixes https://github.com/rust-lang/compiler-builtins/issues/398.

I just realized that no unused warnings are produced for unused functions in traits, even if I make the traits `pub(crate)`. Is there a way I can automatically check for unused functions in traits, so that I can remove them in my other PR?


alexcrichton:
This seems fine for now, thanks! I'm not sure of a way to automatically detect this though.

PR #401, opened 2021-01-21
Remove rotate_left from Int

bjorn3:
Cranelift doesn't implement rotate for small integers, so this is required to be able to compile the latest compiler-builtins with cg_clif.

### DIFF
--- a/src/int/mod.rs
+++ b/src/int/mod.rs
@@ -85,7 +85,6 @@ pub trait Int:
     fn wrapping_sub(self, other: Self) -> Self;
     fn wrapping_shl(self, other: u32) -> Self;
     fn wrapping_shr(self, other: u32) -> Self;
-    fn rotate_left(self, other: u32) -> Self;
     fn overflowing_add(self, other: Self) -> (Self, bool);
     fn aborting_div(self, other: Self) -> Self;
     fn aborting_rem(self, other: Self) -> Self;
@@ -207,10 +206,6 @@ macro_rules! int_impl_common {
 
         fn wrapping_shr(self, other: u32) -> Self {
             <Self>::wrapping_shr(self, other)
-        }
-
-        fn rotate_left(self, other: u32) -> Self {
-            <Self>::rotate_left(self, other)
         }
 
         fn overflowing_add(self, other: Self) -> (Self, bool) {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Would it be possible to fix this in Cranelift and/or cg_clif rather than changing this crate?

bjorn3:
At this point I think it is unlikely that the old x86 backend of Cranelift will get support for rotl on small integers. In addition `rotate_left` isn't used by compiler-builtins, but still codegened, thus slightly bloating compiler-builtins.

AaronKutch:
Wait, does Cranelift not support rotation for small integers at all? Or, is this because of some special bootstrapping requirement it has when building `compiler-builtins`? Does there need to be builtins analogous to `__rust_i128_shlo` but for rotations? I'm thinking the fix should be within Cranelift. Here is an arbitrary sub `usize` rotating function supporting arbitrary bitwidths that I made for a bitwidth-aware bigint library:
```
// assumes `bitwidth != 0` and `shift < bitwidth`
fn subdigit_rotl(bitwidth: usize, x: usize, shift: usize) -> usize {
    ((x >> (bitwidth - shift)) | (x << shift)) & (!0 >> ((usize::MAX.count_ones() as usize) - bitwidth))
}

fn main()  {
    // rotating an 8 bit integer inside a `usize`
    println!("{:08b}", subdigit_rotl(8, 0b10010110, 3));// 10110100
    println!("{:08b}", subdigit_rotl(8, 0b10010110, 4));// 01101001
    println!("{:08b}", subdigit_rotl(8, 0b10010110, 5));// 11010010
    // rotating a 16 bit integer inside a `usize`
    println!("{:016b}", subdigit_rotl(16, 0b1011011101111000, 5));// 1110111100010110
}

// rotate right is just `subdigit_rotl(bitwidth, x, bitwidth - shift)` conditioned on `shift != 0`

```

bjorn3:
> Wait, does Cranelift not support rotation for small integers at all?

Yeah, it isn't implemented yet at least for the old x86_64 backend. I switched the default to the new x86_64 backend, which does support it, a couple of days ago but I want to keep compatibility with the old backend for at least a few weeks just in case I find a big bug in the new backend. In any case this function is unused within compiler-builtins itself and as such shouldn't be codegened. It only bloats the object files.

Amanieu:
@bjorn3 Is this PR still relevant? Do we still need compatibility with the old x86_64 backend?

bjorn3:
I dropped support for the old backend.

PR #407, opened 2021-03-08
add support for building outlined aarch64 intrinsics

gburgessiv:
llvm/llvm-project@a4ac434c47434d80bca54bab96f295ed4e972cc6 saw the
addition of out-of-line aarch64 atomic intrinsics. LLVM will sometimes
emit these, so we need to ensure they're included in Rust's compiler-rt.

===

This probably isn't an issue that we'll see in ToT Rust until we pull in a new LLVM. Seems reasonable to me to hold off on merging this PR until then; I just had to do it anyway, so thought I'd provide the result in case it's helpful. :)

The approach here is admittedly ugly. If we want to stick more closely with the "all builds are done by `cfg.compile("libcompiler-rt.a");`," a potentially cleaner alternative would be to generate a temp .S file for each combination of these things. These would boil down to something like:

```
#define L_cas
#define SIZE 1
#define MODEL relax
#include "path/to/lse.S"
```

LLVM's cmake approach (AIUI) is "create a lot of symlinks to lse.S, and compile those with differing flags." Adopting that approach would still give us badness here, since it doesn't seem `cc` lets us apply file-specific flags.

I'm still doing a bit of extra testing to ensure this does the right thing, but aarch64 builds successfully for me with a local backport of the recent lse.S changes.


bjorn3:
The c feature is disabled by default. A rust port of these would be nice? That would probably require `global_asm!()` if these intrinsics use a custom calling convention. Otherwise `asm!()` could be used, though I think `global_asm!()` would still be nicer.

Amanieu (commented):
Can you rebase to fix CI?

Amanieu:
> The c feature is disabled by default.

Actually it is enabled by default in `src/bootstrap/compile.rs` (via the `compiler-builtins-c` feature in `alloc`/`std`/`test`).

joshtriplett:
This turns out to be an issue even without LLVM emitting references to these intrinsics. If rustc links in a C static library that was built to expect these outlined intrinsics, and it doesn't link in libgcc (which it doesn't on aarch64 musl), that'll result in a build error due to these missing symbols. It'd be helpful to have these symbols available even before we generate code using them ourselves.

joshtriplett:
Closed in favor of https://github.com/rust-lang/compiler-builtins/pull/416 .

PR #341, opened 2020-02-12
add mips/mips64 compiler-rt fallbacks so that libgcc is not required

vfreex:
This adds compiler-rt fallbacks for mips and mips64 arches.

Solves linking issues like https://github.com/rust-lang/rust/issues/57820.

Signed-off-by: Yuxiang Zhu <vfreex@gmail.com>

PR #408, opened 2021-03-24
Add compiler-rt fallbacks on mips64-musl

messense:
See https://github.com/rust-lang/compiler-builtins/pull/341#issuecomment-805516720

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -449,6 +449,8 @@ mod c {
                 ("__fixunstfsi", "fixunstfsi.c"),
                 ("__floatunsitf", "floatunsitf.c"),
                 ("__fe_getround", "fp_mode.c"),
+                ("__divtf3", "divtf3.c"),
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -450,6 +450,10 @@ mod c {
                 ("__floatunsitf", "floatunsitf.c"),
                 ("__fe_getround", "fp_mode.c"),
             ]);
+
+            if target_env == "musl" {
+                sources.extend(&[("__divtf3", "divtf3.c"), ("__trunctfdf2", "trunctfdf2.c")]);
```
Amanieu, line 455: Add these to the list above, there's no need to restrict these to musl only.

messense: Done

Amanieu:
I opened an LLVM bug for the CI failure: https://bugs.llvm.org/show_bug.cgi?id=49710

Amanieu:
Can you rebase? CI should be fixed now.

PR #411, opened 2021-04-03
Add `#[linkage = "weak"]` attribute to all `mem` instrinics.

MabezDev:
Following on from https://github.com/rust-lang/compiler-builtins/pull/379, in particular this comment https://github.com/rust-lang/compiler-builtins/pull/379#issuecomment-812807350. 

The rational for such a change is to provide a simple way to override the builtin memcpy routines, in particular `no_std` targets because the `mem` feature of `compiler-builtins` is always enabled for those targets and cannot be disabled.

### DIFF
--- a/src/mem/mod.rs
+++ b/src/mem/mod.rs
@@ -20,12 +20,14 @@ use core::ops::{BitOr, Shl};
 mod impls;
 
 #[cfg_attr(all(feature = "mem", not(feature = "mangled-names")), no_mangle)]
+#[cfg_attr(not(all(target_os = "windows", target_env = "gnu")), linkage = "weak")]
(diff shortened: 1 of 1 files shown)


Amanieu:
Weak linkage isn't properly supported on some targets (Windows, maybe apple too) and should be disabled on those.

MabezDev:
I grepped the `rustc_target` crate in the compiler repo and found that only windows gnu targets don't support weak linkage along with uefi targets:

```bash
grep -rn "allows_weak_linkage"
windows_gnu_base.rs:80:        allows_weak_linkage: false,
uefi_msvc_base.rs:44:        allows_weak_linkage: false,
```

The changes I have just pushed should cover those cases.

MabezDev:
Current CI error seems unrelated? Failing to pull a docker registry it seems.

PR #414, opened 2021-04-10
Mark global_asm! code with .att_syntax

Amanieu:
global_asm! will soon change to use Intel syntax by default.


bjorn3:
When `global_asm!` is switched to intel syntax by default can you ping me, so I can change cg_clif?

Amanieu:
I will do the necessary changes in cg_clif since it's now part of the rust repo. I will still ping you for review though.

bjorn3:
Thanks!

PR #416, opened 2021-04-30
Support aarch64 outlined atomics

joshtriplett:
On aarch64, GCC and LLVM will now sometimes emit references to
out-of-line intrinsics for atomic operations, used to handle atomic
operations portably between systems with and without the "large-system
extensions" (LSE) introduced in ARM v8.1.

Add these intrinsics to Rust's compiler-rt, both for code generated by
Rust and for libraries a user might link in.

Based on the PR at
https://github.com/rust-lang/compiler-builtins/pull/407 .


joshtriplett:
The CI error here looks unrelated; it's about uses of the `bench_black_box` feature.

Amanieu:
CI should be fixed by #417

Amanieu:
Can you rebase on the latest master?

joshtriplett:
Looks like the path to lse.S isn't correct, at least in CI. Investigating...

PR #421, opened 2021-05-26
Suppress some warnings

JohnTitor:
Found some warnings when checking a libc CI log.

- We use `#![allow(unused_imports)]` for most modules but arm doesn't have.
- `{memcpy,memmove,memset}_element_unordered_atomic` emits a `dead_code` warning when atomic is disabled (e.g. riscv32i-unknown-none-elf)

### DIFF
--- a/src/mem/mod.rs
+++ b/src/mem/mod.rs
@@ -69,6 +69,7 @@ pub unsafe extern "C" fn bcmp(s1: *const u8, s2: *const u8, n: usize) -> i32 {
 }
 
 // `bytes` must be a multiple of `mem::size_of::<T>()`
+#[cfg_attr(not(target_has_atomic_load_store = "8"), allow(dead_code))]
 fn memcpy_element_unordered_atomic<T: Copy>(dest: *mut T, src: *const T, bytes: usize) {
     unsafe {
         let n = exact_div(bytes, mem::size_of::<T>());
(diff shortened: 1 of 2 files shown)


src/mem/mod.rs
```diff
@@ -69,6 +69,7 @@ pub unsafe extern "C" fn bcmp(s1: *const u8, s2: *const u8, n: usize) -> i32 {
 }
 
 // `bytes` must be a multiple of `mem::size_of::<T>()`
+#[cfg_attr(target_has_atomic_load_store = "0", allow(dead_code))]
```
Amanieu, line 72: Are you sure this actually works? We never generate "0" as a value for `target_has_atomic_load_store`.

Also `memcpy_element_unordered_atomic` should almost always be supported. It is a bug with the riscv target if it doesn't expose the proper `target_has_atomic_load_store` attributes.

On RISC-V normal loads and stores are guaranteed to be atomic, even if the A extension is not supported. It's just that RMW operations are not available.

JohnTitor: Hmm, I thought this (https://github.com/rust-lang/rust/pull/66548) created `target_has_atomic_load_store = "0"`. Then does just `#[allow(dead_code)]` work here?

Amanieu: I see, the RISC-V LLVM target behaves differently from the ARM one in this regards. In any case, you should be able to just change the condition to `not(target_has_atomic_load_store = "8")`.

JohnTitor: Updated it, thanks!

PR #424, opened 2021-06-02
Add public-test-deps feature for better visibility control

### LINKED ISSUE #422: `u128_divide_sparc` get build into rust stdlib for x86_64-pc-windows-msvc target
Checked exported symbols for `std-[hash].dll` for current stable and nightly and found `_ZN17compiler_builtins3int19specialized_div_rem8delegate17u128_divide_sparc17hd4dfbbe4547e607fE` inside, i guess #393 is source of this.

I guess sparc related functions shouldn't get build into other os stdlib.

AaronKutch:
fixes #422. ~~Now, the only unwanted things in the public interface are some macros. This is unavoidable because `#[macro_export]` does not allow for fine grained control.~~ The public interface is finally fixed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,10 @@ no-lang-items = []
 # Only used in the compiler's build system
 rustc-dep-of-std = ['compiler-builtins', 'core']
 
+# This makes certain traits and function specializations public that
+# are not normally public but are required by the `testcrate`
+public-test-deps = []
+
 [[example]]
 name = "intrinsics"
 required-features = ["compiler-builtins"]
--- a/src/float/mod.rs
+++ b/src/float/mod.rs
@@ -11,9 +11,9 @@ pub mod mul;
 pub mod pow;
 pub mod sub;
 
+public_test_dep! {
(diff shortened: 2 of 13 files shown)


src/lib.rs
```diff
@@ -66,3 +66,23 @@ pub mod x86;
...
+
+/// Changes the visibility to `pub` if feature "public-test-deps" is set
+#[doc(hidden)]
+#[cfg(feature = "public-test-deps")]
+#[macro_export]
+macro_rules! public_test_dep {
+    {$(#[$($meta:meta)*])* pub(crate) $ident:ident $($tokens:tt)*} => {
+        $(#[$($meta)*])* pub $ident $($tokens)*
+    };
+}
```
bjorn3, line 88: If you place the macros before the `mod` items, you shouldn't need `#[macro_export]` and `use crate::public_test_dep`.

AaronKutch: Thanks, I managed to remove all macros from the public interface. I'm pretty sure I tried this before and it wouldn't work, I must have had some detail wrong.

Issue #434, opened 2021-09-21
request: add fmodf for x86_64-unknown-uefi


gshep:
Hello! I tried to use several math crates in a sample for x86_64-unknown-uefi and encountered the following issues:
1. libm - requires `__truncdfsf2`. There is already an issue for it = #327 
1. micromax - requires fmodf. I created an issue in the project's repo (https://github.com/tarcieri/micromath/issues/95)  but the author suggested me to request it here. The build error log can be found there.
1. https://github.com/nagisa/math.rs - requires both `fmodf` & `__trunsdfsf2`.

So I would like `fmodf` be added to the builtins. Thanks in advance!

bjorn3:
fmodf already exists in the libm crate: https://github.com/rust-lang/libm/blob/master/src/math/fmodf.rs It is only included by compiler-builtins on wasm32-unknown-none, baremetal arm and the fortanix sgx environment though: https://github.com/rust-lang/compiler-builtins/blob/e6fd1b272ff4cc34810e20126ffe17888a708f39/src/lib.rs#L45-L50

gshep:
Then I'd like it to be enabled for x86_64-unknown-uefi =)

Amanieu:
I'd be happy to accept a PR, I think you only need to add a line to that `cfg` block @bjorn3 linked to.

Issue #388, opened 2020-10-19
multiple rlib candidates for `compiler_builtins` found


jiayihu:
I'm having a similar issue to #334 but it's different I think. I have `build-std = ["core", "alloc"]` in  config.toml and I'm trying to compile a local patch of compiler-builtins, for the reasons stated in #387 . In Cargo.toml I have the following configuration:

```toml
wasm3 = { version = "0.1.3", default-features = false, features = ["use-32bit-slots"] }
compiler_builtins = { version = "0.1.36" }

[patch.crates-io]
wasm3 = { path = "../../wasm3-rs" }
compiler_builtins = { path = "../../compiler-builtins" }
```

But the compiler is giving the error:

```
error[E0465]: multiple rlib candidates for `compiler_builtins` found
  |
  = note: candidate #1: /Users/jiayihu/Desktop/Repo/fedra-thesis/fedra/target/thumbv7em-none-eabi/debug/deps/libcompiler_builtins-c00a7bd3257465d5.rlib
  = note: candidate #2: /Users/jiayihu/Desktop/Repo/fedra-thesis/fedra/target/thumbv7em-none-eabi/debug/deps/libcompiler_builtins-7cae6c5d923380b0.rlib
```

I believe one version is the one included by default by cargo because of `core` in https://github.com/rust-lang/cargo/blob/master/src/cargo/core/compiler/standard_lib.rs#L28, whereas the second one is my patch. How do I opt-out the cargo-inserted one?

yuanyuan2100:
Got the same error when I was complying for avr-atmega328p.

  ```
= note: candidate #1: /Rust/arduino/target/avr-atmega328p/release/deps/libcompiler_builtins-13aec7bae4adb9d2.rlib
  = note: candidate #2: /Rust/arduino/target/avr-atmega328p/release/deps/libcompiler_builtins-a98f3389f6f9d633.rlib
```

**Comment by haraldh** [comment]
maybe this helps:
```toml
[build]
rustflags = [
    "-C", "link-self-contained=no",
]
```

stlankes:
I have a similar problem and the compiler flag doesn't help.

tgross35:
This is a pretty old issue and it seems like there are solutions, so I will close this.

Issue #327, opened 2019-12-03
Missing __truncdfsf2 for aarch64-unknown-none-softfloat


rahealy:
I took a look at the C files referenced in this repo's README.md but don't feel that I'm qualified to make the port to rust.

I'd be very grateful if this function could be prioritized for implementation. Thank you.

vinc:
I have the same issue it seems on a project targeting `86_64-unknown-none` and `-mmx,-sse,+soft-float` when using `libm::atan2`:

```
  = note: rust-lld: error: undefined symbol: __truncdfsf2
          >>> referenced by libm.3ij53uav-cgu.12
          >>>               libm-94520a1e125a28f6.libm.3ij53uav-cgu.12.rcgu.o:(libm::math::atan::atan::h43feb293179e50c4) in archive /home/v/src/vinc/moros/target/x86_64-moros/release/deps/liblibm-94520a1e125a28f6.rlib
```

andre-richter:
There are two bare-metal aarch64 targets for rustc:

aarch64-unknown-none
aarch64-unknown-none-softfloat

The latter is intended for Kernels or Hypervisors where code that uses floating point would be concenptionally ill-suited in most cases.

It just happens that the current rust compiler infrastructure does fall back to fixed point implementations of float operations if avail, hence the unusual target name.

Hitting a case where a float intrinsic is missing with the softfloat  target is probably either a good hint that you accidentally included float code where you should not, or that the aarch64-unknown-none target is what you want instead.

dbeckwith:
Ran into the same linker error on `x86_64-unknown-none` with the `soft-float` feature enabled. https://github.com/rust-num/num-traits/issues/167 has the details. I'll look into the implementation that was going to be added in #262 and see if that works for me.

HTG-YT:
Just ran into this issue today whilst trying to target `x86_64-unknown-uefi`.

sajattack:
Same thing on mips2 with `+single-float` (mipsel-sony-psp tier 3 target).

sajattack:
Looks like it will be fixed soon? https://github.com/rust-lang/compiler-builtins/pull/448

ayrtonm:
These should be available on nightly now that rust-lang/rust#91986 is merged.

PR #435, opened 2021-09-29
expose some math to UEFI envs

gshep:
issue #434 

this was a bit tricky cause I had to use this workaround https://github.com/rust-lang/compiler-builtins/issues/388#issuecomment-907099273 for testing (also rename `__fltused` in my sources since it doesn't get mangled). Tested on a sample with math for UEFI and on a project for WASM; seems to be working =).

So the diff is simple: just expose several math-functions for x86_64/uefi environments. The remaining functions produces undefined reference to `truncdfsf2` (#327).


bjorn3:
I think it should be easy to port `truncdfsf2` to rust. It directly forwards to this function: https://github.com/rust-lang/llvm-project/blob/cba558df777a045b5657d56c29944e9e8fd3a776/compiler-rt/lib/builtins/fp_trunc_impl.inc#L41-L132

gshep:
there already was an attempt to do it - #262 , but unfortunately it was reverted. So maybe later =)

PR #436, opened 2021-10-08
Add xtensa to list of soft math targets.

MabezDev:
I am right in thinking these software impls only get used if there isn't hardware accelerated support for the current target? I.e some xtensa chips will be able to accelerate `fma` for example, adding this change wouldn't stop that from happening on chips that support those hardware instructions,  right?


Amanieu:
Can you rebase on the latest master to fix the CI failure (fixed in #437)?

PR #439, opened 2021-10-28
Adjust some build directives for wasm64

alexcrichton:
This is still an experimental target but this should get the wasm64
target to behave more like wasm32.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -23,9 +23,9 @@ fn main() {
         return;
     }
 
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! Would it be ok to request a version bump for this as well?

Amanieu:
Done

alexcrichton:
Thanks!

PR #448, opened 2021-12-12
Add `__truncdfsf2` intrinsic

ayrtonm:
This adds the truncdfsf2 intrinsic and a corresponding fuzz test case. I also kept the implementation of `trunc` generic to make it easy to add truncdfhf2 and truncsfhf2 if rust ever gets `f16` support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ features = ["c"]
 - [x] arm/softfloat-alias.list
 - [x] arm/subdf3vfp.S
 - [x] arm/subsf3vfp.S
-- [ ] arm/truncdfsf2vfp.S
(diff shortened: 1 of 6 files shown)


ayrtonm:
I added `__truncdfsf2vfp` to try to fix the failing thumbv7 tests, but I'm not sure if I need to add anything else to make sure LLVM generates the optimized version from compiler-rt. Also looks like there's some issues when checking for duplicate symbols on both arm and x86.

Amanieu:
You may need to add `#[maybe_use_optimized_c_shim]`, see the explanation in `macros.rs`.

PR #452, opened 2022-02-06
Wrap all intrinsics in the intrinsics! macro

Amanieu:
This ensures that each intrinsic ends up in a separate module, which in turn (because rustc treats compiler_builtins specially) will result in each intrinsic ending up in its own object file. This allows the linker to only pick up object files for intrinsics that are missing and avoids duplicate symbol definition errors.


petrochenkov:
@Amanieu 
Do you plan to also bump the `compiler-builtins` version in the main rust-lang/rust repo?

Amanieu:
Yes, I'm working on it.

PR #455, opened 2022-02-13
Handle Win64 builtins ABI change in LLVM 14

nikic:
As of https://reviews.llvm.org/D110413, these no longer use the unadjusted ABI (and use normal C ABI instead, passing i128 indirectly and returning it as `<2 x i64>`).

To support both LLVM 14 and older versions, rustc will expose a `"llvm14-builtins-abi"` target feature, based on which compiler-builtins can chose the appropriate ABI.

This is needed for https://github.com/rust-lang/rust/pull/93577.

### DIFF
--- a/src/float/conv.rs
+++ b/src/float/conv.rs
@@ -109,16 +109,6 @@ intrinsics! {
         }
     }
 
-    #[unadjusted_on_win64]
-    pub extern "C" fn __floattisf(i: i128) -> f32 {
-        int_to_float(i)
-    }
-
-    #[unadjusted_on_win64]
-    pub extern "C" fn __floattidf(i: i128) -> f64 {
-        int_to_float(i)
-    }
-
     #[arm_aeabi_alias = __aeabi_ui2f]
     pub extern "C" fn __floatunsisf(i: u32) -> f32 {
         int_to_float(i)
@@ -138,16 +128,6 @@ intrinsics! {
     #[maybe_use_optimized_c_shim]
     #[arm_aeabi_alias = __aeabi_ul2d]
     pub extern "C" fn __floatundidf(i: u64) -> f64 {
-        int_to_float(i)
-    }
-
-    #[unadjusted_on_win64]
-    pub extern "C" fn __floatuntisf(i: u128) -> f32 {
-        int_to_float(i)
-    }
-
-    #[unadjusted_on_win64]
-    pub extern "C" fn __floatuntidf(i: u128) -> f64 {
         int_to_float(i)
     }
 }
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Doesn't this need to be cfg gated somehow to retain support for older llvms?

nikic:
@Mark-Simulacrum Possibly. I assumed this doesn't matter for Windows, where distros linking against system LLVM are not a thing.

nikic:
If we want this gated on LLVM version, would it be acceptable to call `$RUSTC -vV` from build.rs and parse the LLVM version from there?

Mark-Simulacrum:
Hm. Probably not, unfortunately. The RUSTC available to build scripts may differ from that used to actually compile the code, particularly for std dependencies (since build scripts link to std themselves, we have to be a little fancy there).

We could expose an unstable target feature like "float-abi-llvm-14" or something that is used here, though, from rustc? Presumably if we cfg_attr on that, then we get the right behavior without needing build script hacks or similar, but also not exposing any new stable surface area from rustc. Once we bump min LLVM to 14, it could be dropped.

mati865:
> @Mark-Simulacrum Possibly. I assumed this doesn't matter for Windows, where distros linking against system LLVM are not a thing.

It's niche but does exist, at MSYS2 we link to LLVM 13 right now but plan to move to LLVM 14 as soon as it's released.
Another use case I can think of is cross compiling, Fedora does build multiple targets (including mingw-w64) for their system Rust package.

src/float/conv.rs
```diff
@@ -268,6 +233,87 @@ intrinsics! {
     pub extern "C" fn __fixunsdfdi(f: f64) -> u64 {
         float_to_int(f)
     }
+}
+
+// The ABI for the following intrinsics changed in LLVM 14. On Win64, they now
+// use Win64 ABI rather than unadjusted ABI. Pick the correct ABI based on the
+// llvm14-builtins-abi target feature.
+
+#[cfg(target_feature = "llvm14-builtins-abi")]
```
bjorn3, line 242: Should this target feature be behind a feature gate?

nikic: Apparently, we don't feature-gate `#[cfg(target_feature)]` (though we do gate `#[target_feature]`). `cfg(target_feature)` exposes all unstable target features on nightly builds.

Mark-Simulacrum:
Yeah, this seems to match my expectation.

One thought I do have: since out-of-tree users of compiler-builtins (https://crates.io/crates/compiler_builtins/reverse_dependencies, at least) are using this crate directly, there's an unfortunate situation where we're basically forced to leave this cfg gate in indefinitely unless we bump the MSRV for the crate to a new enough rustc -- I'm not sure what our policies there look like.

nikic:
@Mark-Simulacrum From the few I looked at, it looks like direct users of compiler_builtins are due to rustc-dep-of-std.

Amanieu:
Also compiler-builtins only supports nightly Rust anyways.

Amanieu:
Published in compiler-builtins 0.1.70.

PR #456, opened 2022-03-08
[watch_os] add support for Apple watchOS

vladimir-ea:
Hello - I am hoping to add compile targets for apple WatchOS and these are prerequisite changes for those new targets to work. 

Thanks!


vladimir-ea:
This is the rust PR: https://github.com/rust-lang/rust/pull/94736

PR #457, opened 2022-03-18
Provide an implementation of `strlen` to be used as a fallback

petrochenkov:
As suggested in https://github.com/rust-lang/rust/pull/94079#issuecomment-1044688919.


Amanieu:
It seems that wasm is missing a definition of `c_char`.

petrochenkov:
It wasn't just wasm, it was anything with `feature = "mem"` enabled, due to a misplaced import.
Fixed now.

PR #458, opened 2022-03-23
Build emutls.c on Android.

pcc:
Android uses emulated TLS so we need a runtime support function
from this source file.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -489,6 +489,16 @@ mod c {
             sources.remove(&["__aeabi_cdcmp", "__aeabi_cfcmp"]);
         }
 
+        // Android uses emulated TLS so we need a runtime support function.
+        if target_os == "android" {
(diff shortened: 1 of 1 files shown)


pcc:
I think this has some problems, I'm getting build failures on 32-bit ARM: 
```
warning: .../android-ndk-r24/toolchains/llvm/prebuilt/linux-x86_64/bin/../sysroot/usr/include/bits/signal_types.h:64:53: error: use of undeclared identifier 'LONG_BIT'
warning: typedef struct { unsigned long __bits[_KERNEL__NSIG/LONG_BIT]; } sigset64_t;
warning:  
```
Strange thing is that emutls.c is including `limits.h` which is where `LONG_BIT` is supposed to come from...

Amanieu:
Thanks! I published a new version of compiler-builtins.

PR #460, opened 2022-05-02
Remove apple_versioning.c inclusion

keith:
According to the README this file isn't used by rust, but it's currently
included which leads to these linker warnings in some cases:

```
ld: warning: linker symbol '$ld$hide$os10.5$___udivti3' hides a non-existent symbol '___udivti3'
ld: warning: linker symbol '$ld$hide$os10.4$___umoddi3' hides a non-existent symbol '___umoddi3'
ld: warning: linker symbol '$ld$hide$os10.5$___umoddi3' hides a non-existent symbol '___umoddi3'
ld: warning: linker symbol '$ld$hide$os10.4$___umodti3' hides a non-existent symbol '___umodti3'
ld: warning: linker symbol '$ld$hide$os10.5$___umodti3' hides a non-existent symbol '___umodti3'
```

This file exclusively contains macros which hides old symbols on Apple
OS versions where they don't exist.

https://github.com/rust-lang/llvm-project/blob/fc10370ef7d91babf512c10505f8f2176bc8519d/compiler-rt/lib/builtins/apple_versioning.c

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -193,7 +193,6 @@ mod c {
             ("__absvsi2", "absvsi2.c"),
             ("__addvdi3", "addvdi3.c"),
             ("__addvsi3", "addvsi3.c"),
-            ("apple_versioning", "apple_versioning.c"),
             ("__clzdi2", "clzdi2.c"),
             ("__clzsi2", "clzsi2.c"),
             ("__cmpdi2", "cmpdi2.c"),


Amanieu:
I think the proper fix here is to ensure that we include the symbols that are supposedly non-existent. We *should* already be providing these symbols in compiler-builtins though, so I'm not sure why this warning is triggering.

keith:
So the issue in my case where I'm seeing the warnings is that we're loading the entire compiler builtins archive, which loads these linker symbols, but the actual functions are unused in our case, so they print this warning. If you pass `-dead_strip` in the same invocation these warnings don't appear (but we don't want to do that in this development case for performance). Building LLVM @ HEAD right now it doesn't appear that the apple_versioning file is included in any platform's libclang_rt libraries, but I think it's gated on deployment target of the libraries through this logic https://github.com/llvm/llvm-project/blob/ad7c49bef774191304b804e60821c0cdb6a390cb/compiler-rt/cmake/Modules/CompilerRTDarwinUtils.cmake#L239-L261

This is the same for the shipping version of Xcode as well (aside from [driverkit](https://developer.apple.com/documentation/driverkit)). Is there a defined minimum version that this library should support? I guess theoretically new symbols could also be added there in the future as well.

Amanieu:
I think this is probably fine to merge. I don't expect it to cause any issues in practice.

keith:
Thanks!

PR #461, opened 2022-05-10
rv64 implement muldi3 intrinsic

### LINKED ISSUE #459: riscv64 infinite recursing during multiplication __muldi3
When I compile a multiplication on `rv64` without the `m` extension I'll end up in an infinite recursion as:
```
#2  0x0000000000021318 in core::num::{impl#8}::wrapping_mul (self=0, rhs=257)
    at /home/johannst/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/num/uint_macros.rs:1249
#3  compiler_builtins::int::{impl#21}::zero_widen_mul (self=0, rhs=257)
    at /home/johannst/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.71/src/int/mod.rs:337
#4  0x00000000000218d8 in compiler_builtins::int::mul::Mul::mul<u64> (self=0, rhs=257)
    at /home/johannst/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.71/src/int/mul.rs:16
#5  0x0000000000021ea0 in compiler_builtins::int::mul::__muldi3 (a=0, b=257)
    at /home/johannst/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.71/src/int/mul.rs:104
#6  0x000000000002177c in compiler_builtins::int::mul::__muldi3::__muldi3 (a=0, b=257)
    at /home/johannst/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.71/src/macros.rs:336

#7  0x0000000000021318 in core::num::{impl#8}::wrapping_mul (self=0, rhs=257)
    at /home/johannst/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/num/uint_macros.rs:1249
#8  compiler_builtins::int::{impl#21}::zero_widen_mul (self=0, rhs=257)
    at /home/johannst/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.71/src/int/mod.rs:337
#9  0x00000000000218d8 in compiler_builtins::int::mul::Mul::mul<u64> (self=0, rhs=257)
    at /home/johannst/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.71/src/int/mul.rs:16
#10 0x0000000000021ea0 in compiler_builtins::int::mul::__muldi3 (a=0, b=257)
    at /home/johannst/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.71/src/int/mul.rs:104
#11 0x000000000002177c in compiler_builtins::int::mul::__muldi3::__muldi3 (a=0, b=257)
    at /home/johannst/.cargo/registry/src/github.com-1ecc6299db9ec823/compiler_builtins-0.1.71/src/macros.rs:336
...
```
A potential solution would be to implement the `__muldi3` intrinsic for rv64 similar to `__mulsi3` for rv32.

To unblock myself I created following patch [rv64-muldi3 - b945767](https://github.com/johannst/compiler-builtins/commit/b945767d46fdc26a64ac8bc7f4dcd3bfb50ce2e2), if that would be an acceptable starting point I am happy to create a PR :)

johannst:
This fixes #459.

Implement the __muldi3 intrinsic to prevent infinite recursion during
multiplication on rv64 without the 'm' extension.

### DIFF
--- a/src/int/mul.rs
+++ b/src/int/mul.rs
@@ -100,6 +100,7 @@ impl_signed_mulo!(i128_overflowing_mul, i128, u128);
 intrinsics! {
     #[maybe_use_optimized_c_shim]
     #[arm_aeabi_alias = __aeabi_lmul]
+    #[cfg(not(target_arch = "riscv64"))]
     pub extern "C" fn __muldi3(a: u64, b: u64) -> u64 {
         a.mul(b)
     }
--- a/src/int/mul.rs
+++ b/src/int/mul.rs
@@ -100,7 +100,7 @@ impl_signed_mulo!(i128_overflowing_mul, i128, u128);
 intrinsics! {
     #[maybe_use_optimized_c_shim]
     #[arm_aeabi_alias = __aeabi_lmul]
-    #[cfg(not(target_arch = "riscv64"))]
+    #[cfg(any(not(any(target_arch = "riscv32", target_arch = "riscv64")), target_feature = "m"))]
     pub extern "C" fn __muldi3(a: u64, b: u64) -> u64 {
         a.mul(b)
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,8 +60,8 @@ pub mod arm;
 ))]
 pub mod arm_linux;
 
-#[cfg(any(target_arch = "riscv32"))]
-pub mod riscv32;
+#[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
+pub mod riscv;
 
 #[cfg(target_arch = "x86")]
 pub mod x86;
--- a/src/riscv.rs
+++ b/src/riscv.rs
@@ -1,7 +1,6 @@
 intrinsics! {
     // Implementation from gcc
     // https://raw.githubusercontent.com/gcc-mirror/gcc/master/libgcc/config/epiphany/mulsi3.c
-    #[cfg(target_arch = "riscv32")]
     pub extern "C" fn __mulsi3(a: u32, b: u32) -> u32 {
         let (mut a, mut b) = (a, b);
         let mut r = 0;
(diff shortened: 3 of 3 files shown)


Amanieu:
In the case of rv32im we would want to use the existing implementation which builds a 64-bit multiply on top of 32-bit multiplies. Could this be made conditional to just targets without `m`?

johannst:
For rv32 the rv64 specific  `__muldi3` intrinsic is not active.
But I agree for the rv64im case. I'll see if we can only enable this in case the `m` extension is not available.

Amanieu:
Even though LLVM doesn't actually use `__muldi3` for 64-bit multiply on rv32im, we should still make it available because it may be used when linking with C code generated by GCC. Same with `__mulsi3`.

johannst:
Okay thanks for the clarification.

Let me try to summarize to check if I understood you correctly
- rv32i: `riscv:__mulsi3`, `riscv:__muldi3`
- rv32im: `riscv:__mulsi3`, `int/mul:__muldi3`
- rv64i: `riscv:__mulsi3`, `riscv:__muldi3`
- rv64im: `riscv:__mulsi3`, `int/mul:__muldi3`

Please correct me if I got you wrong.

Amanieu:
That's correct.

johannst:
I adapted the conditions accordingly.
The one for `int/mul:__muldi3` became somewhat lengthy, if you see a better way I am happy to make it more readable.

PR #462, opened 2022-05-15
Fix division on AVRs

Patryk27:
For division and modulo, AVR uses a custom calling convention that does
not match compiler_builtins' expectations, leading to non-working code¹.

Ideally we'd just use hand-written naked functions (as, afair, ARM
does), but that's a lot of code to port², so hopefully we'll be able to
do it gradually later.

For the time being, I'd suggest not compiling problematic functions for
AVR target - this causes avr-gcc (which is a mandatory part of Rust+AVR
toolchain anyway) to link hand-written assembly from libgcc, which is
confirmed to work.

I've tested the code locally on simavr and the patch seems to be working
correctly :-)

¹ https://github.com/rust-lang/rust/issues/82242,
  https://github.com/rust-lang/rust/issues/83281
² https://github.com/gcc-mirror/gcc/blob/31048012db98f5ec9c2ba537bfd850374bdd771f/libgcc/config/avr/lib1funcs.S

Closes https://github.com/rust-lang/rust/issues/82242
Closes https://github.com/rust-lang/rust/issues/83281

### DIFF
--- a/src/int/sdiv.rs
+++ b/src/int/sdiv.rs
@@ -9,6 +9,7 @@ macro_rules! sdivmod {
         $($attr:tt),* // attributes
     ) => {
         intrinsics! {
+            #[avr_skip]
             $(
                 #[$attr]
             )*
@@ -50,6 +51,7 @@ macro_rules! sdiv {
         $($attr:tt),* // attributes
     ) => {
         intrinsics! {
+            #[avr_skip]
             $(
                 #[$attr]
             )*
@@ -85,6 +87,7 @@ macro_rules! smod {
         $($attr:tt),* // attributes
     ) => {
         intrinsics! {
+            #[avr_skip]
             $(
                 #[$attr]
             )*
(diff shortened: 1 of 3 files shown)


timwalls:
Hmm.  I thought this change would mean I wouldn't need the `compiler-builtins-mangled-names` workaround any more, but without it the latest nightly (`rustc 1.63.0-nightly (4cbaac699 2022-05-25)`) still produces broken code.

Am I misinterpreting what this fix achieves, or has it not made it into nightly yet (exactly when things actually get merged to nightly is somewhat opaque to me :-))?

Amanieu:
I just published a new version of compiler-builtins. It now needs a PR in rust-lang/rust to update the dependency to the latest version.

Patryk27:
Oh neat, I'll prepare the PR in rust-lang, then

timwalls:
> > has it not made it into nightly yet
> 
> Yes, this change hasn't made it into nightly yet -- hopefully in a few days! 🙂

Aha!  Well, then be assured it's eagerly awaited and greatly appreciated, thank you.  I shall go back to being patient :).

aykevl:
See: https://reviews.llvm.org/D138166

PR #465, opened 2022-05-24
Avoid int to ptr transmute by casting first

thomcc:
According to @RalfJung this is UB (or will soon be UB, or something), and this is the workaround. Which is harmless enough.


RalfJung:
Cc https://github.com/rust-lang/rust/pull/97321

thomcc:
> it is harmless enough but I'm not a big fan of this change

For the record, I agree. I think a better fix would be to make that code allowed.

RalfJung:
See https://github.com/rust-lang/unsafe-code-guidelines/issues/340 for the UCG thread; please raise your voice there :)

PR #466, opened 2022-05-24
Add avr_skip for __udivti3 & __umodti3

Patryk27:
I've forgotten to guard those two before and they also don't work at the moment (as confirmed by simavr).

A bit unfortunately, libgcc doesn't provide those functions, making i128/u128 division/modulo non-working on AVR (addition, subtraction and multiplication all seem to be working, as confirmed by random testing, so there's that, at least; hopefully no one plans to divide 128 bit numbers on AVRs 😛).


bjorn3:
Even if libgcc doesn't provide these, compiler-builtins can still provide them without issue, right?

PR #469, opened 2022-06-09
Build C code for `xous` operating system

xobs:
Xous is a new Tier 3 target:

```
$ rustc +nightly --print target-list | rg xous
riscv32imac-unknown-xous-elf
$
```

Xous has no C library at all, and relies on `compiler-builtins` to provide most floating-point math functions. These patches enable C code on this platform.

Note that `cc-rs` still needs to be updated to support `riscv32imac-unknown-xous-elf` -- for now any packages will need to set the `CC` environment variable to point to a valid C compiler.


Amanieu:
Let me know if you need me to publish a new version of the crate.

xobs:
Yes, please. A new version would be much appreciated!

Amanieu:
I published a new release.

PR #472, opened 2022-06-13
Avoid link-time dependency on core through `to_bits` and `from_bits`

tmiasko:
The compiler-builtins should not have any link time dependencies on any
other crates including core.

Replace uses of `to_bits` and `from_bits` from core with local `repr`
and `to_repr` methods implemented in terms of the `transmute` intrinsic.

This fixes a build issue that can occur in build configurations that
share generic monomorphizations between crates.

### DIFF
--- a/src/float/conv.rs
+++ b/src/float/conv.rs
@@ -74,32 +74,38 @@ mod int_to_float {
 intrinsics! {
     #[arm_aeabi_alias = __aeabi_ui2f]
     pub extern "C" fn __floatunsisf(i: u32) -> f32 {
-        f32::from_bits(int_to_float::u32_to_f32_bits(i))
+        use float::Float;
+        f32::from_repr(int_to_float::u32_to_f32_bits(i))
     }
 
     #[arm_aeabi_alias = __aeabi_ui2d]
     pub extern "C" fn __floatunsidf(i: u32) -> f64 {
-        f64::from_bits(int_to_float::u32_to_f64_bits(i))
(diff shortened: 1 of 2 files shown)


bjorn3:
These functions should always be inlined, right? They are marked as `#[inline]` and small enough that no reasonable compiler would not inline them when optimizations are enabled.

tmiasko:
Those functions themselves are indeed marked inline but not everything they transitively use is.

bjorn3:
What do you mean? `f*::to_bits` and `f*::from_bits` are both marked as `#[inline]` and only call `mem::transmute` which is an intrinsic and thus by definition inlined.

tmiasko:
`to_bits` calls `const_eval_select`, which calls `call_once`, which calls a closure, which calls the intrinsic.

bjorn3:
I see. On stable it is a simple transmute. On nightly it uses `const_eval_select`. Making the `rt_f32_to_u32` closure a regular function and marking it as `#[inline]` should help with that I think.

tmiasko:
Given the role of compiler builtins, I would assume that limiting the dependency on the core is desirable regardless, so this was my first choice in terms of the implementation strategy. Modifying the core instead would presumably work as well.

Amanieu:
We can't realistically avoid all dependencies on core, we just have to make sure that we only use inline functions. It is concerning however that `to_bits`/`from_bits` is not properly inlined in core and this should be fixed there.

PR #475, opened 2022-07-10
Enable win64_128bit_abi_hack for x86_64-unknown-uefi

nicholasbishop:
The [`x86_64-unknown-uefi`](https://github.com/rust-lang/rust/blob/master/compiler/rustc_target/src/spec/x86_64_unknown_uefi.rs) target is Windows-like, and requires the same altered ABI for some 128-bit integer intrinsics.

See also https://github.com/rust-lang/rust/issues/86494.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -209,13 +209,13 @@ macro_rules! intrinsics {
 
         $($rest:tt)*
     ) => (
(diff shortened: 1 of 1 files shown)


Amanieu:

nicholasbishop:
Thanks, rebase pushed

Issue #477, opened 2022-07-22
Please make aeabi division functions use weak linkage, because they are incorrect for ARMv4T


Lokathor:
The division functions included in this crate pop the link-register value directly into `pc` to return. This is correct for interworking code with v5T and later, but it is **incorrect** for v4T.

With v4T, you *must* use `bx` to switch code modes.

At this time I'm not aware of a way to `cfg` on the ARM version of a target. I think the best fix is to just declare the functions as weak linkage, and then v4T users can just override the functions on their end.

Amanieu:
You can use `cfg(target_feature = "v5te")` for this.

Lokathor:
Hmm. Interesting!

Would you also be willing to accept a PR for weak linkage on the division functions anyway? Because even if I use that target_feature to fix up these general functions with the correct return instructions (which I'll try to do later today) I would also like to be able to define my own versions that are stored in RAM rather than ROM because of the speed difference (it's like a 5x time difference in my case).

Amanieu:
I think that we should have weak linkage on all functions in compiler-builtins actually, subject to target support. Anything except apple/windows uses ELF where this should work fine.

PR #474, opened 2022-07-03
Align destination in x86_64's mem* instructions.

Demindiro:
While misaligned reads are generally fast, misaligned writes aren't and can have severe penalties.

I don't know if LLVM 9 is still supported. I've used Intel syntax anyways since it's more readable IMO.

Benchmark results on a Ryzen 2700X:
<details>
<summary><code>master</code></summary>
<pre>
test memset_rust_1048576              ... bench:      18,450 ns/iter (+/- 7,933) = 56833 MB/s
test memset_rust_1048576_offset       ... bench:      36,641 ns/iter (+/- 4,096) = 28617 MB/s
test memset_rust_4096                 ... bench:          93 ns/iter (+/- 3) = 44043 MB/s
test memset_rust_4096_offset          ... bench:         144 ns/iter (+/- 4) = 28444 MB/s
</pre>
<pre>
test memcpy_rust_1048576              ... bench:      56,233 ns/iter (+/- 8,037) = 18646 MB/s
test memcpy_rust_1048576_misalign     ... bench:      63,518 ns/iter (+/- 3,180) = 16508 MB/s
test memcpy_rust_1048576_offset       ... bench:      57,239 ns/iter (+/- 9,745) = 18319 MB/s
test memcpy_rust_4096                 ... bench:          87 ns/iter (+/- 12) = 47080 MB/s
test memcpy_rust_4096_misalign        ... bench:         161 ns/iter (+/- 10) = 25440 MB/s
test memcpy_rust_4096_offset          ... bench:         164 ns/iter (+/- 12) = 24975 MB/s
</pre>
<pre>
test memmove_rust_1048576             ... bench:      53,519 ns/iter (+/- 5,885) = 19592 MB/s
test memmove_rust_1048576_misalign    ... bench:     102,426 ns/iter (+/- 3,702) = 10237 MB/s
test memmove_rust_4096                ... bench:         234 ns/iter (+/- 31) = 17504 MB/s
test memmove_rust_4096_misalign       ... bench:         425 ns/iter (+/- 5) = 9637 MB/s
</pre>
</details>

<details>
<summary><code>x86_64-mem-align-dest</code></summary>
<pre>
test memset_rust_1048576              ... bench:      18,862 ns/iter (+/- 4,941) = 55591 MB/s
test memset_rust_1048576_offset       ... bench:      19,768 ns/iter (+/- 4,017) = 53044 MB/s
test memset_rust_4096                 ... bench:          93 ns/iter (+/- 8) = 44043 MB/s
[...]

PR #478, opened 2022-07-22
add weak linkage to the ARM AEABI division functions

Lokathor:
Closes https://github.com/rust-lang/compiler-builtins/issues/477

I know in the issue Amanieu said that *all* functions in this crate should probably get weak linkage when possible, but that seems overkill for my own needs and i didn't want to have to determine when the target will use ELF or not for a lot of functions i'd never heard of.

### DIFF
--- a/src/arm.rs
+++ b/src/arm.rs
@@ -22,6 +22,7 @@ intrinsics! {
     // custom calling convention which can't be implemented using a normal Rust function.
     #[naked]
     #[cfg(not(target_env = "msvc"))]
+    #[linkage = "weak"]
     pub unsafe extern "C" fn __aeabi_uidivmod() {
         core::arch::asm!(
             "push {{lr}}",
@@ -36,6 +37,7 @@ intrinsics! {
     }
 
     #[naked]
+    #[linkage = "weak"]
     pub unsafe extern "C" fn __aeabi_uldivmod() {
         core::arch::asm!(
             "push {{r4, lr}}",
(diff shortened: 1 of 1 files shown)


Lokathor:
Note: It was mentioned in the issue that these functions are also incorrect on v4T. I'll resolve that as a separate PR so that we're just talking about one thing at a time.

Lokathor:
I didn't touch those memory functions, so the test error seem unrelated.

Lokathor:
@Amanieu ping

Amanieu:
Sorry about the delay, I'm still looking into the powerpc issue. It seems like it might be an LLVM regression.

Amanieu:

Amanieu:
Could you make the weak linkage conditional on `!windows && !apple`? Basically only on ELF targets.

Lokathor:
do you detect for apple like this?
```rust
    #[cfg_attr(all(not(windows), not(target_vendor="apple")), linkage = "weak")]
```
or something else?

Amanieu:
Yes, that's correct.

Lokathor:
should be done i guess

Lokathor:
also, does this automatically get picked up in every nightly release once merged, or is there a separate PR required for that?

Amanieu:
I need to publish a new version of the crate (I want to get #474 merged first), and then a PR is needed in rust-lang/rust to update the crate version in Cargo.lock.

Amanieu:
I've published version 0.1.77.

PR #480, opened 2022-07-28
Use all of src/math for UEFI

Ayush1325:
The functions `sqrt` and `floor` are used by libtest.


Ayush1325:
Is the PowerPC test broken? seems it fails in all of the new PRs

Amanieu:

PR #486, opened 2022-08-03
Enable `unadjusted_on_win64` for UEFI in some cases

nicholasbishop:
The conversion functions from i128/u128 to f32/f64 have the `unadjusted_on_win64` attribute, but it is disabled starting with LLVM14. This seems to be the correct thing to do for Win64, but for some reason x86_64-unknown-uefi is different, despite generally using the same ABI as Win64.

### DIFF
--- a/src/float/conv.rs
+++ b/src/float/conv.rs
@@ -92,12 +92,12 @@ intrinsics! {
         f64::from_bits(int_to_float::u64_to_f64_bits(i))
     }
 
-    #[cfg_attr(not(target_feature = "llvm14-builtins-abi"), unadjusted_on_win64)]
+    #[cfg_attr(any(not(target_feature = "llvm14-builtins-abi"), target_os = "uefi"), unadjusted_on_win64)]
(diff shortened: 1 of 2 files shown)


nicholasbishop:
I'm really not at all sure why the ABI of these particular intrinsics would need to be different between Win64 and x86_64-unknown-uefi, but I have verified with a test UEFI program that this change fixes the conversions.

Amanieu:
@nikic Any idea what's going on with the ABI here? I thought UEFI was supposed to act just like Windows with regards to `i128` intrinsics.

nikic:
I think this happens because we declare uefi as a soft-float target: https://llvm.godbolt.org/z/3846ExoPz

Amanieu:
So is this a bug in LLVM? Or should this PR be merged?

nikic:
This is probably an LLVM bug, but at the same time we don't particularly care what the actual ABI here is, so I think it's fine to merge this.

nicholasbishop:
Thanks for checking on that and connecting it to soft-float :)

PR #495, opened 2022-09-27
Weak Linkage for "ARM Alias" functions

Lokathor:
This adds the weak linkage attribute used in the previous PR (https://github.com/rust-lang/compiler-builtins/pull/385)

```rust
#[cfg_attr(all(not(windows), not(target_vendor="apple")), linkage = "weak")]
```

to the "ARM alias" version of a function generated by the `intrinsics!` macro.

The previous PR didn't catch this case.


Lokathor:
@Amanieu I'm gonna ping you since there doesn't appear to be a bot that auto-assigns a reviewer.

Issue #491, opened 2022-09-03
Link-time dependency on core: use of `#[inline]` functions


fee1-dead:
See also https://github.com/rust-lang/rust/pull/100759#discussion_r960597855. This is undesirable as it causes a confusing linker error when modifying the library functions. #472 was opened to fixed this, but it was closed because it looked like adding `#[inline]` to fix this had been more appropriate. I encountered this issue again in the link above and I am creating this issue for further discussion as to whether this is an issue or not.

Amanieu:
`compiler-builtins` depends on a ton of inline functions in `core`, such as `wrapping_add`. I don't think it is reasonable for us to avoid using these functions entirely.

RalfJung:
I just spent almost an hour debugging strange linker errors until I realized they are caused by compiler_builtins. The current situation seems pretty fragile and it'd be great to find some proper solution here.

tgross35:
We now get an error "error: `compiler_builtins` cannot call functions through upstream monomorphizations; encountered invalid call from `foo` to `rustc_std_workspace_core::bar`", so I think this situation has been improved (even if the error sometimes lacks details).

Urhengulas:
Is it possible to configure compiler-builtins to be able to call libcore functions through upstream monomorphisation and therefore be able to not inline the core functions?

I understand this is bad for performance, but for our usecase this would be acceptable. Or is there another reason this is necessary apart from performance?

bjorn3:
No, there can't be a cyclic dependency between libcore and compiler-builtins. If there is ld.bfd will error out.

tgross35:
> I understand this is bad for performance, but for our usecase this would be acceptable. Or is there another reason this is necessary apart from performance?

What usecase is this?

PR #496, opened 2022-10-09
invoke the unreachable intrinsic, not the stable wrapper

RalfJung:
Works around https://github.com/rust-lang/compiler-builtins/issues/491: `hint::unreachable_unchecked` would like to trigger a panic in debug builds, but that would involve non-`inline` functions, so we cannot use it from compiler_builtins any more.


RalfJung:
Clippy fails, but that is not caused by my changes.

Amanieu:
Clippy is fixed by #498.

RalfJung:
Thanks - I rebased, hopefully CI passes now.

PR #502, opened 2022-11-12
Typo fix

nicholasbishop:


Cargo.toml
```diff
@@ -58,7 +58,7 @@ mem = []
 # compiler-rt implementations. Also used for testing
 mangled-names = []
 
-# Don't generate lang items for i128 intrisnics and such
+# Don't generate lang items for i128 intrinsics and such
 no-lang-items = []
```
bjorn3, line 62: Looks like this feature no longer has any effect. Maybe it should be removed?

nicholasbishop: Good catch :) I'll close this in favor of https://github.com/rust-lang/compiler-builtins/pull/503.

PR #506, opened 2022-11-16
Use a stub stdlib.h when compiling for UEFI targets

nicholasbishop:
int_util.c includes stdlib.h if `_WIN32` is defined. When compiling the UEFI targets with clang they are treated as Windows targets (e.g. if the Rust target is x86_64-unknown-uefi, the clang target is x86_64-unknown-windows-gnu). So stdlib.h gets included, even though we are compilling with `-ffreestanding` and don't want stdlib.h to be used. That file may not be present, or an incompatible version might be installed leading to typedef redefinition errors.

The contents of stdlib.h aren't actually needed for these targets anyway (due to `__STDC_HOSTED__` being 0), so create an empty stdlib.h in `build.rs` when `target_os == uefi` and add it to the include path.


Amanieu:
The freestanding part of `stdlib.h` is `stddef.h`, so it would be better to have a single `#include <stddef.h>` in the dummy `stdlib.h`. Although it doesn't matter here, this should be more robust against future changes of compiler-rt.

nicholasbishop:
Updated the stub header to add the include as suggested.

Issue #307, opened 2019-08-03
Crate declares incorrect Apache/MIT license


brson:
The source code in this crate is derivative of LLVM's compiler-rt, the license of which is declared correctly in LICENSE.TXT, but not in Cargo.toml, and is the [NCSA license](https://opensource.org/licenses/NCSA) (similar to BSD).

Cargo.toml needs to be updated to declare the NCSA license, and then the Rust tidy file in deps.rs to declare it as either an exception or an acceptable permissive license.

Note that LLVM is in the process of relicensing to add a runtime exception to its code (which Rust lacks), and this crate will not obviously benefit from that relicensing since it was derived prior to the relicense.

cc https://github.com/rust-lang/rust/issues/63232

brson:
This crate also includes a Rust port of MUSL's libm: https://github.com/rust-lang-nursery/libm, which also seems to incorrectly be described as Apache/MIT.

brson:
libm issue: https://github.com/rust-lang-nursery/libm/issues/215

brson:
Oh, actually, compiler-rt is NCSA/MIT, so this crate can declare just "MIT" or "MIT/NCSA" or "MIT or NCSA". It just can't add the option of alternately distributing under the apache license (I think).

mati865:
LLVM has been relicensed to Apache 2.0 with LLVM exception.

infinity0:
If that implies that compiler-rt has also been relicense to Apache-2.0-with-LLVM-exception, can Cargo.toml and LICENSE.TXT here be updated please? This uncertainty is confusing one of our copyright reviewers at Debian and slowing down the acceptance of this crate into Debian.

infinity0:
Also, the relicensing only applies to the source code of LLVM *at the version at which the relicensing was done*. If this crate still uses *old* code that is different from the relicensed version, then it still is licensed under NCSA. So Cargo.toml would still have to be updated.

infinity0:
Please ignore my previous comments, I think they are obsolete and I have deleted them.

As I understand it, the files in `src/**` are all rust translations that do not appear in the original `compiler-rt` work, therefore it is a derivative work, which has been authorised by the license of `compiler-rt` at the time of derivation. Since it is a derivative work, the new copyright holders are free to chose a license for it, according to the restrictions of the original work (in this case there are no restrictions, either in NCSA/MIT/Apache), it does not have to be a copy of the old license. So it is fine to relicense as Apache/MIT.

[...]

PR #511, opened 2022-12-13
Correct SPDX identifier to "MIT OR (Apache-2.0 AND NCSA)"

SvizelPritula:
The `Cargo.toml` file specifies the license as `MIT/Apache-2.0`. This library, however, inherited LLVM's `MIT/NCSA` licensing. LLVM has since switched to `MIT OR NCSA OR (Apache-2.0 WITH LLVM-exception)`, so it's probably possible to re-license this crate to `MIT/Apache-2.0`, however, the fact is that this hasn't happened yet, since the Readme still says that contributors license their contributions under `MIT/NCSA`. This implies that all contributors would need to agree to license their code under `Apache-2.0` first.

This would make the license of this repo `MIT/NCSA`, but the crate also directly bundles `libm`, which uses `MIT/Apache-2.0`. Combining the two, we get `MIT OR (Apache-2.0 AND NCSA)`.

This issue is being discussed in #307. This PR implements what I believe to be the correct solution to this problem. Please exercise caution, as I am not a lawyer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 authors = ["Jorge Aparicio <japaricious@gmail.com>"]
 name = "compiler_builtins"
 version = "0.1.85"
-license = "MIT/Apache-2.0"
+license = "MIT OR (Apache-2.0 AND NCSA)"
 readme = "README.md"
 repository = "https://github.com/rust-lang/compiler-builtins"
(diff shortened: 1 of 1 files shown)


Amanieu:
Assigning to @pnkfelix since this is license-related.

tgross35:
This was superseded by https://github.com/rust-lang/compiler-builtins/pull/717

PR #517, opened 2023-02-22
Extend the intrinsics exported for Xtensa no_std

MabezDev:
Builds on top of #436 to make more software intrinsics available to the no_std targets.

Closes https://github.com/esp-rs/rust/issues/155


MabezDev:
Turns out the math module wasn't actually enabled for Xtensa arch, which is now fixed.

PR #482, opened 2022-07-28
Add tgamma and tgammaf

ankane:
For https://github.com/rust-lang/rust/pull/99747

PR #518, opened 2023-03-04
Add lgamma_r and lgammaf_r

ankane:
For https://github.com/rust-lang/rust/issues/99842#issuecomment-1454900224

Similar to #482

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -87,6 +87,31 @@ no_mangle! {
 }
 
 #[cfg(any(
+    all(
+        target_family = "wasm",
+        target_os = "unknown",
+        not(target_env = "wasi")
+    ),
+    target_os = "xous",
+    all(target_arch = "x86_64", target_os = "uefi"),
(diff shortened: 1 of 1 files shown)


ankane:
Note the signature [is different](https://linux.die.net/man/3/lgamma_r) from `math.h` - not exactly sure if that should be taken into account here or downstream for https://github.com/rust-lang/rust/pull/99747.

ankane:
Now thinking it may need to be taken into account here since the signature in [cmath.rs](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/cmath.rs) will be:

```rust
pub fn lgamma_r(n: f64, s: &mut i32) -> f64;
```

Amanieu:
The signature here should match the C one. You'll need to manually generate the wrapper function instead of using the `no_mangle` macro.

ankane:
Thanks (again) @Amanieu. Updated to match the C signature.

ankane:
Thanks!

PR #521, opened 2023-03-22
Fix panic due to overflow in riscv.rs and int/shift.rs

taiki-e:
I ran into this while testing `riscv32i-unknown-none-elf` with `-Z build-std=core,alloc`.
```
panicked at 'attempt to add with overflow', /Users/taiki/.cargo/registry/src/index.crates.io-6f17d22bba15001f/compiler_builtins-0.1.87/src/riscv.rs:42:17
```

(Can be reproduced by cloning [semihosting repo](https://github.com/taiki-e/semihosting), comment out [this line](https://github.com/taiki-e/semihosting/blob/249ad13cd8faa4346a29afa48dfdb9c5b6a3f6d0/tools/no-std.sh#L257), and run `./tools/no-std riscv32i-unknown-none-elf`)

### DIFF
--- a/src/int/shift.rs
+++ b/src/int/shift.rs
@@ -12,7 +12,7 @@ trait Ashl: DInt {
         } else {
             Self::from_lo_hi(
                 self.lo().wrapping_shl(shl),
-                self.lo().logical_shr(n_h - shl) | self.hi().wrapping_shl(shl),
+                self.lo().logical_shr(n_h.wrapping_sub(shl)) | self.hi().wrapping_shl(shl),
             )
         }
     }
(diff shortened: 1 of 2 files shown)


taiki-e:
I found another panic due to overflow on the other targets (thumbv5te-none-eabi) as well. I will update the PR to include a fix for them later.

```
panicked at 'attempt to subtract with overflow', /Users/taiki/.cargo/registry/src/index.crates.io-6f17d22bba15001f/compiler_builtins-0.1.87/src/int/shift.rs:15:39
```

```
panicked at 'attempt to subtract with overflow', /Users/taiki/.cargo/registry/src/index.crates.io-6f17d22bba15001f/compiler_builtins-0.1.87/src/int/shift.rs:60:69
```

davidben:
A related issue we uncovered when we hit a variation on this in Arm: https://github.com/llvm/llvm-project/issues/62743

Issue #520, opened 2023-03-17
For msp430, `__ashldi3` takes a 16-bit integer as the second argument in LLVM's `compiler-rt`. Probably affects other builtins.


cr1901:
I've created a test case that panics on msp430 in debug/release mode:

```rust
fn do_shift(val: &u64, amt: &u8) -> u64 {
    // compiler_builtins::int::shift::__ashldi3(*val, *amt as u32)
    *val << *amt
}

fn shift_test() {
    let my_u64: u64 = 1;
    let my_shift: u8 = 0;

    if black_box(do_shift(black_box(&my_u64), black_box(&my_shift))) != 1 {
        panic!()
    }
}
```

If I call `compiler_builtins::int::shift::__ashldi3(*val, *amt as u32)` directly, `shift_test` does not panic. Relevant assembly code:

```asm
.Ltmp6:
	.loc	1 46 18 is_stmt 1
	mov	6(r12), r15
	mov	4(r12), r14
	mov	2(r12), r13
	mov	0(r12), r12
	mov.b	0(r11), r11
.Ltmp7:
	.loc	1 38 5
	mov	r11, 0(r1)
	clr	2(r1)
	call	#_ZN17compiler_builtins3int5shift9__ashldi317h573dd44357b45ab3E
```

However, if I let LLVM lower the shift to a builtin, `shift_test` tends to panic:

```asm
.Ltmp7:
	.loc	1 46 18 is_stmt 1
	mov.b	0(r12), r12
.Ltmp8:
	.loc	1 39 5
	and	#63, r12
	mov	r12, 0(r1)
.Ltmp9:
	.loc	1 46 18
	mov	0(r15), r12
	mov	2(r15), r13
	mov	4(r15), r14
	mov	6(r15), r15
.Ltmp10:
	.loc	1 39 5
	call	#__ashldi3
```

[...]

PR #522, opened 2023-03-26
Ensure shift instrinsic arguments match width of compiler-rt's.

cr1901:
This fixes [#520](https://github.com/rust-lang/compiler-builtins/issues/520). I tested this locally and confirmed this fixes my issue, along with the additional changes listed in the below sections _just for testing_.

I realized while working on this patch that as far as `compiler-builtins` is concerned, only the shift instrinsics are affected by 16-bit vs 32-bit mismatch. This dates back to a [patch](https://github.com/llvm/llvm-project/commit/4d41df64828195aa24cddc5d34d3f446ca7bb6d1) by @aykevl in 2020.

It is very possible that the msp430 backend generates code w/ incorrect width for other instrinsics changed in the above patch. However, I don't know a good way to sift these out w/ tests, Rust or otherwise. Perhaps https://github.com/rust-lang/rust/issues/107612 could be useful here? `mspdebug` has a simulator; it's [how](https://github.com/rust-lang/rust/issues/107612#issuecomment-1472377638) I was able to get a test case for #520.

## `compiler-builtins` patches

```diff
diff --git a/src/lib.rs b/src/lib.rs
index 71f249c..dabe076 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,6 +14,7 @@
 #![no_builtins]
 #![no_std]
 #![allow(unused_features)]
+#![allow(unexpected_cfgs)]
 // We use `u128` in a whole bunch of places which we currently agree with the
 // compiler on ABIs and such, so we should be "good enough" for now and changes
 // to the `u128` ABI will be reflected here.
```

Without this, for some reason Rust bails with `-D warnings`, due to https://github.com/rust-lang/rust/issues/96472 when I'm using `compiler-builtins` as a path dep (but not when using a crates.io dep... huh...).

## `rust` patches

```diff
diff --git a/Cargo.lock b/Cargo.lock
index 449f0c73588..c8a69ff7d5a 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -868,9 +868,7 @@ dependencies = [
 
 [[package]]
 name = "compiler_builtins"
-version = "0.1.87"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f867ce54c09855ccd135ad4a50c777182a0c7af5ff20a8f537617bd648b10d50"
+version = "0.1.90"
 dependencies = [
  "cc",
  "rustc-std-workspace-core",
diff --git a/Cargo.toml b/Cargo.toml
index 15cbb2659c9..8fa9b040bfa 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,6 +104,12 @@ gimli.debug = 0
 miniz_oxide.debug = 0
 object.debug = 0
 
+[profile.release.package.core]
+opt-level = 's'
+
+[profile.release.package.alloc]
+opt-level = 's'
+
 [patch.crates-io]
 # See comments in `src/tools/rustc-workspace-hack/README.md` for what's going on
 # here
@@ -114,6 +120,7 @@ rustc-workspace-hack = { path = 'src/tools/rustc-workspace-hack' }
 rustc-std-workspace-core = { path = 'library/rustc-std-workspace-core' }
 rustc-std-workspace-alloc = { path = 'library/rustc-std-workspace-alloc' }
 rustc-std-workspace-std = { path = 'library/rustc-std-workspace-std' }
+compiler_builtins = { version="0.1.87", path = '../compiler-builtins' }
 
 [patch."https://github.com/rust-lang/rust-clippy"]
 clippy_lints = { path = "src/tools/clippy/clippy_lints" }
diff --git a/library/backtrace b/library/backtrace
--- a/library/backtrace
+++ b/library/backtrace
@@ -1 +1 @@
-Subproject commit 07872f28cd8a65c3c7428811548dc85f1f2fb05b
+Subproject commit 07872f28cd8a65c3c7428811548dc85f1f2fb05b-dirty
```

Use local `compiler-builtins` with my patch, work around some _unrelated_ LLVM assertion failures at `opt-level=3`.

### DIFF
--- a/src/int/shift.rs
+++ b/src/int/shift.rs
@@ -78,8 +78,8 @@ intrinsics! {
     #[avr_skip]
     #[maybe_use_optimized_c_shim]
     #[arm_aeabi_alias = __aeabi_llsl]
-    pub extern "C" fn __ashldi3(a: u64, b: u32) -> u64 {
-        a.ashl(b)
+    pub extern "C" fn __ashldi3(a: u64, b: core::ffi::c_uint) -> u64 {
+        a.ashl(b as u32)
     }
 
     #[avr_skip]
@@ -96,8 +96,8 @@ intrinsics! {
     #[avr_skip]
     #[maybe_use_optimized_c_shim]
     #[arm_aeabi_alias = __aeabi_lasr]
-    pub extern "C" fn __ashrdi3(a: i64, b: u32) -> i64 {
-        a.ashr(b)
+    pub extern "C" fn __ashrdi3(a: i64, b: core::ffi::c_uint) -> i64 {
+        a.ashr(b as u32)
     }
 
     #[avr_skip]
@@ -114,8 +114,8 @@ intrinsics! {
     #[avr_skip]
     #[maybe_use_optimized_c_shim]
     #[arm_aeabi_alias = __aeabi_llsr]
-    pub extern "C" fn __lshrdi3(a: u64, b: u32) -> u64 {
-        a.lshr(b)
+    pub extern "C" fn __lshrdi3(a: u64, b: core::ffi::c_uint) -> u64 {
+        a.lshr(b as u32)
     }
 
     #[avr_skip]


cr1901:
>https://github.com/rust-lang/compiler-builtins/actions/runs/4523089026/jobs/7965979554?pr=522#step:8:941

Interesting... well, `floatuntisf` [does use](https://github.com/rust-lang/compiler-builtins/blob/5511f3087255236680eefb862458ed2f90e11bb5/src/float/conv.rs#L52-L60) shifts, but without having access to the assembly, I can't tell why these conversions weren't a noop on 32-bit platforms and above. Ahhh well, not solving that tonight, will take a look when I have time.

Amanieu:
Can you rebase now that #524 is merged, which should fix CI?

cr1901:
@Amanieu Done, and CI is green :D!

src/int/shift.rs
```diff
@@ -78,8 +78,8 @@ intrinsics! {
     #[avr_skip]
     #[maybe_use_optimized_c_shim]
     #[arm_aeabi_alias = __aeabi_llsl]
-    pub extern "C" fn __ashldi3(a: u64, b: u32) -> u64 {
-        a.ashl(b)
+    pub extern "C" fn __ashldi3(a: u64, b: core::ffi::c_uint) -> u64 {
+        a.ashl(b as u32)
```
aykevl, line 82: Drive by comment:

While this appears to follow the calling convention that LLVM uses, it does _not_ follow the calling convention that avr-gcc uses: https://godbolt.org/z/azq4749hd
`b` is only a single byte on avr-gcc.
(I also see now that the C implementation in compiler-rt has the same bug, but only realized it now).

cr1901: AIUI, these code paths are skipped when compiling for `avr`, precisely because there's a pass which gets rid of these types of shifts in LLVM. So I made no accommodations for its nonstandard calling conventions in this PR.

aykevl: > because there's a pass which gets rid of these types of shifts in LLVM

I know, because I wrote that pass ;)
But it only works for 32-bit values.`__ashldi3` is a 64-bit shift function, which is emitted as a regular compiler-rt call. Proof: https://godbolt.org/z/zqWfE6ndq

Anyway, it doesn't really matter as LLVM uses `i16` and not `i8` like avr-gcc does. This PR would only be buggy for avr-gcc, _right now_ it is safe with LLVM (but this could in theory change as `i8` is more efficient).

cr1901: Minddump for future-me follows:

>b is only a single byte on avr-gcc.

* According to [gcc docs](https://gcc.gnu.org/onlinedocs/gccint/Integer-library-routines.html), `__ashldi3`'s `b` argument is supposed to take an `int`, which must be at least 16 bits.  But indeed, your GCC code snippet is only sending a byte to  `__ashldi3`, while the LLVM version sends 16 bits (the `clr r17` line is clearing the top-most bits).
* On the other hand, `gcc`'s docs also says `__ashldi3` is supposed to return a `long`, and `__ashlti3` is the function that is supposed to return a `long long`. 
* On the other _other hand, `compiler-rt` doesn't seem to support the `t` versions of functions, and tends to hardcode a 64/32-bit int in places gcc's builtins do not<sup>1</sup>. In addition to `compiler-rt` uses `unsigned` in args where gcc's builtins are `signed`.

In other words, it seems like _both_ `gcc` `compiler-rt` builtins tend to do their own things and aren't directly compatible with each other. `avr` using a byte in for the `gcc` `__ashldi3` `b` arg, but 2 bytes for `compiler-rt`'s `__ashldi3` `b` arg is just one example of this. It's not going to be fun :P. Unifying `compiler-rt`'s and `gcc`'s builtins in terms of `compiler_builtins` might be necessary due to the `gcc` backend though.

1. The shift functions' `b` args notwithstanding :P.

cr1901: >But it only works for 32-bit values. __ashldi3 is a 64-bit shift function, which is emitted as a regular compiler-rt call.

How does this not break for Rust code, where all the shift functions are [skipped](https://github.com/rust-lang/compiler-builtins/blob/master/src/int/shift.rs#L72-L124) for AVR (`avr_skip`)?

Issue #525, opened 2023-05-18
The "mangled-names" feature removes intrinsics from compiler_builtins that aren't in compiler_rt


danakj:
We're [trying](https://bugs.chromium.org/p/chromium/issues/detail?id=1445978) to link Rust into a C++ target and we don't want to create ODR violations between compiler_rt and compiler_builtins.

To avoid this, we set the `mangled-names` feature on compiler_builtins, so that the intrinsics do not get demangled. This avoids all collisions!

However Rust's compiler_builtins provides some intrinsics that are not present in clang's compiler_rt, for 128-bit integers. Such as: `__udivti3` and `__umodti3` which are referenced by `rust\library\std\src\sys\windows\thread_parking.rs:169`.

An error showing this is here: https://logs.chromium.org/logs/chromium/buildbucket/cr-buildbucket/8780748803129506673/+/u/compile__with_patch_/stdout#L7628_7

The `mangled-names` feature [claims](https://github.com/rust-lang/compiler-builtins/blob/4677881d50f2e2f44ae7720ce19366a78fdf0e77/Cargo.toml#L57-L58) that it is for using Rust with compiler_rt, which is good. But it still needs to provide intrinsics that a C++ compiler_rt does not have.

How do we use `mangled-names` but keep 128-bit integer intrinsics that Rust needs and C/C++ does not provide?

bjorn3:
`mangled-names` is meant only for testing purposes. If you use it, compiler-builtins is effectively unused as LLVM won't know to use the mangled names when emitting calls into it (and LLVM doesn't support changing the names afaik). Duplicate definitions in compiler_rt and compiler-builtins shouldn't matter. Rust's build system builds compiler-builtins with like 1000 codegen units and we place ever intrinsic in it's own module. This combined effectively causes every intrinsic to end up in a separate object file inside the libcompiler_builtins.rlib archive. By default linkers only pull in archive members if any of the symbols defined in them are actually used. This means that if an intrinsic is defined by both compiler-builtins and compiler_rt, the linker will pull in the right object for either compiler-builtins or compiler_rt, but not both, thus preventing a ODR violation.

[...]

PR #526, opened 2023-05-19
Add the weak-intrinsics feature

danakj:
When enabled, the weak-intrinsics feature will cause all intrinsics functions to be marked with weak linkage (i.e. `#[linkage = "weak"]) so that they can be replaced at link time by a stronger symbol.

This can be set to use C++ intrinsics from the compiler-rt library, as it will avoid Rust's implementation replacing the compiler-rt implementation as long as the compiler-rt symbols are linked as strong symbols. Typically this requires the compiler-rt library to be explicitly specified in the link command.

Addresses https://github.com/rust-lang/compiler-builtins/issues/525.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -25,6 +25,12 @@ macro_rules! public_test_dep {
 /// platforms need and elsewhere in this library it just looks like normal Rust
 /// code.
 ///
+/// When the weak-intrinsics feature is enabled, all intrinsics functions are
+/// marked with #[linkage = "weak"] so that they can be replaced by another
+/// implementation at link time. This is particularly useful for mixed Rust/C++
+/// binaries that want to use the C++ intrinsics, otherwise linking against the
+/// Rust stdlib will replace those from the compiler-rt library.
+///
 /// This macro is structured to be invoked with a bunch of functions that looks
 /// like:
 ///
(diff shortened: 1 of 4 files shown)


src/macros.rs
```diff
@@ -283,6 +291,38 @@ macro_rules! intrinsics {
         intrinsics!($($rest)*);
     );
 
+    // Explicit weak linkage gets dropped when weak-intrinsics is on since it
+    // will be added unconditionally to all intrinsics and would conflict
+    // otherwise.
+    (
+        #[linkage = "weak"]
```
danakj, line 298: This is not doing what I want it to do, and I am still getting double linkage errors. Will debug, surely I don't understand something about macro_rules

danakj: Right, I understand and will need to make some changes to how this works in order to avoid adding conditionals to every linkage="weak" elsewhere.

danakj: Got it now.

Without weak-intrinsics:
00000000 W __aeabi_memclr8   // Is explicitly weak
00000000 T __udivsi3  // Is not.

With weak-intrinsics:
00000000 W __aeabi_memclr8
00000000 W __udivsi3

Amanieu:
Thanks! If this works well, we may even enable it by default in the future.

PR #528, opened 2023-06-17
add nightly toolchain toml

knickish:
Making this a separate PR as I'm not sure if it's desirable or not, but the code requires nightly to be built anyway.


knickish:
Closing, as this looks like it breaks infra currently

PR #531, opened 2023-06-25
port updated version of llvm float div

knickish:
This pr updates the float division code to match more recent versions of LLVM's implementations from compiler-rt, adding proper handling of subnormal values. I tested this by fuzzing all f32 values and an equivalent number of f64s against hard float operations on my computer (AMD zen3 cpu) in addition to updating the existing tests in the repo. 

There seem to be very slight differences against the `arm` target, not sure if there's a better approach than just ignoring subnormals on that arch.

### DIFF
--- a/src/float/div.rs
+++ b/src/float/div.rs
@@ -12,11 +12,17 @@ where
     i32: CastInto<F::Int>,
     F::Int: CastInto<i32>,
     F::Int: HInt,
+    <F as Float>::Int: core::ops::Mul,
 {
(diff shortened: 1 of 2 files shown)


Amanieu:
Thanks!

Regarding ARM, I know the the ARM SIMD instructions always flush subnormals to zero, so that might be what is happening here. The normal scalar FP instructions properly support subnormals though. I think it's fine to leave it under a `#[cfg]` with this explanation as a comment.

knickish:
Okay, added a comment next the `cfg` to explain.

PR #532, opened 2023-06-26
Port outline-atomics to rust

jyn514:
This has a very long history, summarized in https://github.com/rust-lang/rust/issues/109064. This port is a very minimal subset of `aarch64/lse.S` from LLVM's compiler-rt. In particular, it is missing the following:

1. Any form of runtime dispatch between LL/SC and LSE.
Determining which version of the intrinsics to use requires one of the following:
    1. `getauxval` from glibc. It's unclear whether `compiler_builtins` is
allowed to depend on libc at all, and musl's support for getauxval interacts poorly with the static linking shenanigans rustc does by default. Don't enshrine the requirement "de-facto" by making it required for outline-atomics.
    2. kernel support. Linux and FreeBSD have limited support, but it
requires an extremely recent kernel version and doesn't work at all under QEMU (https://github.com/rust-lang/rust/issues/109064#issuecomment-1494939904).

    Instead, we hard-code LL/SC intrinsics. Users who want LSE support should use the LLVM compiler-rt (if you're building from source in rust-lang/rust, make sure you have `src/llvm-project` checked out locally. the goal is to soon add a new `optimized-compiler-builtins` option so this is easier to discover).

2. The global `___aarch64_have_lse_atomics` CTOR, required to do runtime dispatch without runtime overhead that defeats the point of the LSE intrinsics. @thomcc has this to say about global CTORs:
    > static ctors are problems because we are pretty eager about dead code elim
    > in general if you have a module that isnt directly reference we will probably not have its static ctors
    > also, while llvm has a super robust way to have a static ctor (theres s special "appending global" to use for c++), we     dont use that and just have people make a #[used] static in a special section
    > 1. the robust way kinda requires rust knowing that the argument is a static ctor (maybe a #[rustc_static_ctor] attribute). it also would be... finnicky, since on windows we actually care beyond being a static ctor, that we run as part in a specific group of ctors, which means a very specific section (one for TLS and the other for, uh, i dont remember)
    > 2. we still actually have to codegen the cgu that isn't referenced. but maybe we could remember that it has that attribute and use that

    So while this is possible in theory, it's decidedly non-trivial, and needs invasive changes to rustc itself. In any case, it doesn't matter until we decide the story around libc.

Note that 16-bytes CAS is not used at all in the standard library that I can find, so unlike the rest of these ops it's only tested by the unit tests, not by any real usage.

### DIFF
--- a/src/aarch64.rs
+++ b/src/aarch64.rs
@@ -22,6 +22,7 @@
 //!
 //! The original `lse.S` has some truly horrifying code that expects to be compiled multiple times with different constants.
 //! We do something similar, but with macro arguments.
+#![cfg_attr(feature = "c", allow(unused_macros))] // avoid putting the macros into a submodule
 
 // We don't do runtime dispatch so we don't have to worry about the `__aarch64_have_lse_atomics` global ctor.
 
--- a/src/aarch64.rs
+++ b/src/aarch64.rs
@@ -131,7 +131,7 @@ macro_rules! compare_and_swap {
         intrinsics! {
             #[maybe_use_optimized_c_shim]
             #[naked]
-            pub extern "C" fn $name (
+            pub unsafe extern "C" fn $name (
                 expected: int_ty!($bytes), desired: int_ty!($bytes), ptr: *mut int_ty!($bytes)
             ) -> int_ty!($bytes) {
                 // We can't use `AtomicI8::compare_and_swap`; we *are* compare_and_swap.
@@ -162,7 +162,7 @@ macro_rules! compare_and_swap_i128 {
         intrinsics! {
             #[maybe_use_optimized_c_shim]
             #[naked]
-            pub extern "C" fn $name (
+            pub unsafe extern "C" fn $name (
                 expected: i128, desired: i128, ptr: *mut i128
             ) -> i128 {
                 unsafe { core::arch::asm! {
@@ -192,7 +192,7 @@ macro_rules! swap {
         intrinsics! {
             #[maybe_use_optimized_c_shim]
             #[naked]
-            pub extern "C" fn $name (
+            pub unsafe extern "C" fn $name (
                 left: int_ty!($bytes), right_ptr: *mut int_ty!($bytes)
             ) -> int_ty!($bytes) {
                 unsafe { core::arch::asm! {
@@ -218,7 +218,7 @@ macro_rules! fetch_op {
         intrinsics! {
             #[maybe_use_optimized_c_shim]
             #[naked]
-            pub extern "C" fn $name (
+            pub unsafe extern "C" fn $name (
                 val: int_ty!($bytes), ptr: *mut int_ty!($bytes)
             ) -> int_ty!($bytes) {
                 unsafe { core::arch::asm! {
--- a/src/aarch64.rs
+++ b/src/aarch64.rs
@@ -12,31 +12,21 @@
 //! Ported from `aarch64/lse.S` in LLVM's compiler-rt.
 //!
 //! Generate functions for each of the following symbols:
+//!  __aarch64_casM_ORDER
 //!  __aarch64_swpN_ORDER
 //!  __aarch64_ldaddN_ORDER
 //!  __aarch64_ldclrN_ORDER
 //!  __aarch64_ldeorN_ORDER
 //!  __aarch64_ldsetN_ORDER
-//! for N = {1, 2, 4, 8}, M = {1, 2, 4, 8}, ORDER = { relax, acq, rel, acq_rel }
-//!
-//! TODO: M = 16
+//! for N = {1, 2, 4, 8}, M = {1, 2, 4, 8, 16}, ORDER = { relax, acq, rel, acq_rel }
 //!
 //! The original `lse.S` has some truly horrifying code that expects to be compiled multiple times with different constants.
 //! We do something similar, but with macro arguments.
 
-/// We don't do runtime dispatch so we don't have to worry about the global ctor.
-/// Apparently MacOS uses a different number of underscores in the symbol name (???)
-// #[cfg(target_vendor = "apple")]
-// macro_rules! have_lse {
-//     () => { ___aarch64_have_lse_atomics }
-// }
-
-// #[cfg(not(target_vendor = "apple"))]
-// macro_rules! have_lse {
-//     () => { __aarch64_have_lse_atomics }
-// }
+// We don't do runtime dispatch so we don't have to worry about the `__aarch64_have_lse_atomics` global ctor.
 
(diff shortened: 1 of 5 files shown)


src/aarch64.rs
```diff
@@ -0,0 +1,221 @@
...
+            #[maybe_use_optimized_c_shim]
+            #[naked]
+            pub extern "C" fn $name (
+                expected: int_ty!($bytes), desired: int_ty!($bytes), ptr: *mut int_ty!($bytes)
+            ) -> int_ty!($bytes) {
+                // We can't use `AtomicI8::compare_and_swap`; we *are* compare_and_swap.
+                unsafe { core::arch::asm! {
+                    // UXT s(tmp0), s(0)
+                    concat!(uxt!($bytes), " ", reg!($bytes, 16), ", ", reg!($bytes, 0)),
+                    "0:",
```
thomcc, line 122: IIRC the `0` and `1` anonymous label names cause issues for some assemblers and so we have to start with `2`. That said, this might only be a problem on x86.

bjorn3:
> Instead, we hard-code LL/SC intrinsics. Users who want LSE support should use the LLVM compiler-rt (if you're building from source in rust-lang/rust, make sure you have src/llvm-project checked out locally. the goal is to soon add a new optimized-compiler-builtins option so this is easier to discover).

compiler-builtins takes priority over compiler-rt amd libgcc as it comes first in the linker order. On desktop OSes where these intrinsics actually matter you already get either libgcc or compiler-rt as dependency by gcc/clang when they act as linker.

bjorn3:
Can you please make this respect the no-asm feature. Cg_clif needs this as it doesn't support inline asm on AArch64 yet.

thomcc:
(Note that I think I wrote the quoted message when I was very tired several months ago, so I'm not sure it's as, err, polished as might be desirable. I think it's *mostly* correct, but I don't really remember what I was talking about with regards to the windows stuff)

taiki-e:
> Determining which version of the intrinsics to use requires one of the following:

FYI, the latest release of the Linux kernel (6.4) has added a way to read auxv without depending on either libc or mrs trap (https://github.com/torvalds/linux/commit/ddc65971bb677aa9f6a4c21f76d3133e106f88eb). It is more complicated than just calling getauxval since it requires using asm to call syscall and then also parsing auxv, but if versions of the kernel that supports it becomes widely used in the future, we might want to use that way.

It is also possible to get auxv without depending on libc by [calling syscall using asm](https://github.com/taiki-e/portable-atomic/commit/ac8ddbf5fe29d746bd024211cc805f84a5882203) even on FreeBSD. However, IIRC, FreeBSD does not guarantee the stability of raw syscall as much as Linux does ([It may actually be stable enough, though](https://lists.llvm.org/pipermail/llvm-dev/2019-June/133393.html)).

(I think I can talk about other operating systems if needed (such as those listed [here](https://github.com/taiki-e/portable-atomic/tree/main/src/imp/atomic128#run-time-feature-detection), or NetBSD, which is [not yet listed](https://github.com/taiki-e/portable-atomic/pull/66)).)

> musl doesn't even support getauxval.

No, musl supports getauxvel [since 1.1.0](https://git.musl-libc.org/cgit/musl/tree/WHATSNEW?h=v1.1.0#n1197). AFAIK, the problem with musl seems to be that getauxval is not well recognized when using static linking (maybe due to linking [modifier](https://github.com/rust-lang/libc/commit/fd331f65f214ea75b6210b415b5fd8650be15c73) or order). (See also [this comment](https://github.com/taiki-e/portable-atomic/blob/dfbad889609d4872c4bc32f0cee34bde104190a3/src/imp/atomic128/detect/auxv.rs#L6-L31) for the availability of getauxval in various libc.)

src/aarch64.rs
Amanieu: It's only an issue on x86.

src/aarch64.rs
```diff
@@ -0,0 +1,221 @@
...
+//!  __aarch64_ldsetN_ORDER
+//! for N = {1, 2, 4, 8}, M = {1, 2, 4, 8}, ORDER = { relax, acq, rel, acq_rel }
+//!
+//! TODO: M = 16
+//!
+//! The original `lse.S` has some truly horrifying code that expects to be compiled multiple times with different constants.
+//! We do something similar, but with macro arguments.
+
+/// We don't do runtime dispatch so we don't have to worry about the global ctor.
+/// Apparently MacOS uses a different number of underscores in the symbol name (???)
```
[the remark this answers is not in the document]
jyn514: ah gotcha - i'll remove this whole commented out secdtion then.

src/macros.rs
```diff
@@ -419,7 +419,7 @@ macro_rules! intrinsics {
     (
         #[naked]
         $(#[$($attr:tt)*])*
-        pub unsafe extern $abi:tt fn $name:ident( $($argname:ident:  $ty:ty),* ) $(-> $ret:ty)? {
```
Amanieu, line 422: All these functions should be unsafe since they take raw pointer arguments.

src/lib.rs
```diff
@@ -57,6 +57,13 @@ pub mod mem;
 #[cfg(target_arch = "arm")]
 pub mod arm;
 
+#[cfg(all(
+    target_arch = "aarch64",
+    not(feature = "no-asm"),
+    not(feature = "optimized-c")
```
jyn514, line 63: note that this means we won't generate the call to the extern symbol in `maybe_use_optimized_c_shim` when `optimized-c` is disabled: https://github.com/rust-lang/compiler-builtins/blob/07cf3b4f109de98910a443c3cacce779eba54cf0/src/macros.rs#L213-L222
i'm not sure how much that matters - it seemed to work ok in core? that whole section of code is very confusing to me, it looks like it's doing an infinitely recursive call but clearly it works *somehow*

Amanieu: It's not infinitely recursive  because the outer function is not `no_mangle`. I agree that it's quite confusing though.

Amanieu: I think `not(feature = "optimized-c")` should be removed here, the tests rely on these functions being present (it's just that with `optimized-c` they redirect to the compiler-rt version).

src/lib.rs
```diff
@@ -57,6 +57,9 @@ pub mod mem;
 #[cfg(target_arch = "arm")]
 pub mod arm;
 
+#[cfg(all(target_arch = "aarch64", not(feature = "no-asm"),))]
+pub mod aarch64;
```
jyn514, line 61: @nilstrieb found in https://github.com/rust-lang-ci/rust/actions/runs/5433440965/jobs/9881115904#step:24:8768 that this is wrong - it also needs to gate on `target_os = "linux"`

PR #536, opened 2023-07-06
Use the target_feature "zbb" instead of "b" for RISC-V

AaronKutch:
When I first wrote this there was a unified "b" bitmanip extension, but recently it was frozen under multiple smaller extensions, of which Zbb is the correct one (https://github.com/riscv/riscv-bitmanip).
"zbb" is what shows on today's `rustc --target=riscv64gc-unknown-linux-gnu --print target-features`


AaronKutch:
CI failing because MIPS is unavailable

Amanieu:
CI should be fixed by #537, can you rebase?

PR #538, opened 2023-07-07
Add lgamma_r and lgammaf_r to MSVC

ankane:
Extends #518 to MSVC to address https://github.com/rust-lang/rust/pull/99747#issuecomment-1606211071


ankane:
It looks like the CI failure is unrelated

ankane:
Thanks @Amanieu!

PR #540, opened 2023-07-10
Work around rust-lang/rust#113533

Amanieu:
This was causing build failures on AArch64 due to name resolution ambiguity.


Amanieu:
Closing in favor of a fix in rustc: https://github.com/rust-lang/rust/pull/113557

PR #545, opened 2023-07-30
Add lgamma_r and lgammaf_r to MSVC

ankane:
Missed this was needed for #538 

Ref: https://github.com/rust-lang/rust/pull/99747#issuecomment-1657104983


ankane:
Thanks @Amanieu!

PR #546, opened 2023-07-31
Add lgamma_r and lgammaf_r to MinGW

ankane:
Sorry for the many PRs. Looks like it's missing from MinGW (`*-pc-windows-gnu` and `*-pc-windows-gnullvm`) as well (https://github.com/rust-lang/rust/pull/99747#issuecomment-1659227726).

[Another source](https://www.gnu.org/software/gnulib/manual/html_node/lgamma_005fr.html)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -51,7 +51,7 @@ pub mod int;
     all(target_arch = "mips", target_os = "none"),
     target_os = "xous",
     all(target_vendor = "fortanix", target_env = "sgx"),
-    target_env = "msvc"
+    target_os = "windows"
 ))]
 pub mod math;
 pub mod mem;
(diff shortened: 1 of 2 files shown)


workingjubilee:
Can't we redesign this cfg so that it is available by default and then rearrange it to used an optimized form on some platforms instead?

ankane:
I'm not exactly sure how it would work (would need to prevent duplicate symbol errors), but maybe others have ideas

workingjubilee:
Mmm, something something weak linking?

Amanieu:
We have support for generating weak symbols (via -`Z build-std-features=compiler-builtins-weak-intrinsics`), but it seems to be broken on mingw: https://github.com/rust-lang/rust/pull/113822

ankane:
Thanks @Amanieu!

mati865:
Weak symbols on Windows with ld.bfd are a can of worms.
Generally you can expect them to rather work as long as you are combining static archives with weak symbols into one final binary/library. Trying to do the same with shared libraries might work for simplest cases but generally think of it as broken.
You may have more luck with `__declspec(selectany)` but it will use more COMDATs which Windows limits to 2^16.

LLD is has working weak symbols (in both static and shared cases) but we cannot use it because of another Binutils bug import libraries created by LLD break symbol table with Binutils <2.40.

PR #547, opened 2023-08-07
update to edition 2018, allow internal_features

klensy:
Crate was on 2015 edition, so update to 2018. Can be changed to 2021 (probably not. libm is submodule on 2018 edition, how that worked before?).

Also adds `#![allow(internal_features)]` recently added in https://github.com/rust-lang/rust/pull/108955, without it build stops on recent nightly.

### DIFF
--- a/examples/intrinsics.rs
+++ b/examples/intrinsics.rs
@@ -5,6 +5,7 @@
 
 #![allow(unused_features)]
 #![allow(stable_features)] // bench_black_box feature is stable, leaving for backcompat
+#![allow(internal_features)]
 #![cfg_attr(thumb, no_main)]
 #![deny(dead_code)]
 #![feature(bench_black_box)]
--- a/src/int/specialized_div_rem/mod.rs
+++ b/src/int/specialized_div_rem/mod.rs
@@ -306,5 +306,6 @@ impl_binary_long!(
     u32_normalization_shift,
     32,
     u32,
-    i32
+    i32,
+    allow(dead_code)
 );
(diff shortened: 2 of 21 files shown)


src/int/specialized_div_rem/mod.rs
```diff
@@ -300,6 +300,8 @@ impl_asymmetric!(
 );
 
 // 32 bits is the smallest division used by `compiler-builtins`, so we end with binary long division
+// FIXME: remove dead_code
+#[allow(dead_code)]
```
klensy, line 304: This fixme for https://github.com/rust-lang/compiler-builtins/actions/runs/5786737992/job/15682137705#step:8:330
```
   --> src/int/specialized_div_rem/mod.rs:304:5
    |
304 |     u32_div_rem,
    |     ^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
    = note: this warning originates in the macro `impl_binary_long` (in Nightly builds, run with -Z macro-backtrace for more info)
```
how this should be gated?

Amanieu: Attributes don't work on macros. You need to put the `#[allow]` on the actual function inside the macro definition.

klensy: Yes, i see warning now, but why it prevents CI fail in that case?

examples/intrinsics.rs
```diff
@@ -5,6 +5,7 @@
 
 #![allow(unused_features)]
 #![allow(stable_features)] // bench_black_box feature is stable, leaving for backcompat
+#![allow(internal_features)]
```
klensy, line 8: Btw, bootstrap compiler won't say anything about unknown lint?

Amanieu: I believe the bootstrap compiler allows warnings, but this may have changed since I last checked.

klensy: Builds with v0.1.100.

PR #548, opened 2023-08-10
build: compile memory intrinsics on xous

xobs:
Like SGX, Xous does not have any libc to link against. As a result, memory intrinsics need to be available as part of `compiler_builtins`

This will enable us to build for Xous as part of `-Zbuild-std` as referenced in https://github.com/rust-lang/rust/pull/104101#discussion_r1288530982

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -30,6 +30,7 @@ fn main() {
         || target.contains("-none")
         || target.contains("nvptx")
         || target.contains("uefi")
+        || target.contains("xous")
     {
         println!("cargo:rustc-cfg=feature=\"mem\"");
     }


build.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(unused_extern_crates)]
```
Amanieu, line 1: Why is this needed?

xobs: It was needed in order to get it to compile with the particular version of Rust that I was building with. It might be that it was an older or newer version where it didn't like the `extern crate cc` declaration. I'll remove it and force-push.

PR #557, opened 2023-11-21
avoid the core_intrinsics feature (for a transition)

RalfJung:
This should let us do the transition in https://github.com/rust-lang/rust/pull/118123.

### DIFF
--- a/src/arm_linux.rs
+++ b/src/arm_linux.rs
@@ -1,5 +1,6 @@
-use core::intrinsics;
 use core::mem;
+
+use crate::intrinsics;
 
 // Kernel-provided user-mode helper functions:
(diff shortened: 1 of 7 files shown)


Nilstrieb:
maybe its easier to just special case core_intrinsics in rustc instead of renaming

RalfJung:
That is an option, but hard-coding a lib feature name in the compiler did not seem great...

Nilstrieb:
🤷 it seems easier than all of this nightmare stuff, and like, I don't think it will change or cause significant trouble to hardcode it, intrinsics are very entangled anyways

RalfJung:
You are probably right, I'm just too much of a perfectionist. :see_no_evil:

RalfJung:
All right, trying to do this without a rename now.

PR #558, opened 2023-11-26
fix: Add (even more) `#[avr_skip]` for floats

Patryk27:
Tale as old as the world - there's an ABI mismatch: https://github.com/rust-lang/compiler-builtins/pull/527

Fortunately, newest GCCs (from v11, it seems) actually provide most of those intrinsics (even for f64!), so that's pretty cool.

(the only intrinsics not provided by GCC are `__powisf2` & `__powidf2`, but our codegen for AVR doesn't emit those anyway.)

Fixes https://github.com/rust-lang/rust/issues/118079.

### DIFF
--- a/src/float/div.rs
+++ b/src/float/div.rs
@@ -902,11 +902,13 @@ where
 }
 
 intrinsics! {
+    #[avr_skip]
     #[arm_aeabi_alias = __aeabi_fdiv]
     pub extern "C" fn __divsf3(a: f32, b: f32) -> f32 {
         div32(a, b)
     }
 
+    #[avr_skip]
     #[arm_aeabi_alias = __aeabi_ddiv]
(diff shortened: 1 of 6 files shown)


Patryk27:
Btw, technically we only need to comment-out the f64 definitions here, since f32 seem to work correctly (which is why I haven't caught any invalid operations over https://github.com/rust-lang/compiler-builtins/pull/527), but I've commented out both, because:

- GCC provides its own intrinsics for both anyway,
- f32 ABI seems to have aligned by chance - it's not enforced in here, so it feels kinda fragile to leave it.

Patryk27:
By the way, @Amanieu would you mind releasing the next version of compiler-builtins? I'd bump it in rustc straight away 🙂

Amanieu:
Done!

Issue #561, opened 2023-12-06
`__extenddftf2` is found to be missing at runtime when using the WASI SDK to build Rust crates' C code


adamnovak:
The WASI SDK folks said that https://github.com/WebAssembly/wasi-sdk/issues/361 is probably a Rust issue.

I'm trying to build a Rust project that uses `rusqlite` (which compiles C code), for WebAssembly, for the `wasm32-wasi` Rust target.

When I do my build with:
```
RUSTC_LOG="rustc_codegen_ssa::back::link=info" CC_wasm32_wasi="$(pwd)/wasi-sdk-20.0/bin/clang" LIBSQLITE3_FLAGS="-DSQLITE_THREADSAFE=0" cargo build --release --target=wasm32-wasi
```

I end up with the Rust linker linking against `/Users/anovak/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/wasm32-wasi/lib/libcompiler_builtins-a0294efb55e060eb.rlib`, which does not define `__extenddftf2`. But the WASI SDK compiler version 20 (the current release) builds C code assuming it will link against *its* `wasi-sdk-20.0/lib/clang/16/lib/wasi/libclang_rt.builtins-wasm32.a` file, which *does* define `__extenddftf2`.

The WASI SDK folks suggested that the real problem is that the Rust `libcompiler_builtins` *should* define that symbol, and is out of date and needs updating. But I installed the `wasm32-wasi` target with `rustup` within the past 48 hours, so presumably I have the current version.

Can that function be added?

My full list of missing intrinsic-looking functions from my build seems to be:

```
 - func[0] sig=18 <__extenddftf2> <- env.__extenddftf2
 - func[1] sig=19 <__trunctfdf2> <- env.__trunctfdf2
 - func[2] sig=20 <__multf3> <- env.__multf3
 - func[3] sig=20 <__addtf3> <- env.__addtf3
 - func[4] sig=21 <__gttf2> <- env.__gttf2
 - func[5] sig=21 <__getf2> <- env.__getf2
 - func[6] sig=20 <__divtf3> <- env.__divtf3
 - func[7] sig=21 <__lttf2> <- env.__lttf2
 - func[8] sig=22 <__fixtfsi> <- env.__fixtfsi
 - func[9] sig=1 <__floatsitf> <- env.__floatsitf
 - func[10] sig=20 <__subtf3> <- env.__subtf3
[...]

Issue #167, opened 2017-06-24
Duplicate __builtin_clz symbols on MSVC Windows


alexcrichton:
This is due to the fact that [their definition](https://github.com/llvm-mirror/compiler-rt/blob/bc311bbb04f823ee9d58261111a4455a0354c358/lib/builtins/int_lib.h#L92-L127) is not marked `static`. Filing an issue to track this, gonna add a FIXME for now.

alexcrichton:
I opened an [upstream diff](https://reviews.llvm.org/D34599) for this.

AaronKutch:
What's the status on this?

PR #566, opened 2024-01-08
build: Allow building C compiler-rt fallbacks for wasm

trevyn:
The comment disabling wasm is almost 5 years old, and is outdated:

```
// * wasm - clang for wasm is somewhat hard to come by and it's
//   unlikely that the C is really that much better than our own Rust.
```

Clang for wasm is now much easier to come by, and C fallbacks are needed for e.g. https://github.com/rust-lang/compiler-builtins/issues/561 and https://github.com/rusqlite/rusqlite/pull/1010

For the `__builtin_sadd_overflow` thing, see https://github.com/llvm/llvm-project/pull/69305 and https://github.com/rust-lang/compiler-builtins/issues/167

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -90,8 +90,8 @@ jobs:
     - run: rustup component add llvm-tools-preview
     - name: Download compiler-rt reference sources
       run: |
-        curl -L -o code.tar.gz https://github.com/rust-lang/llvm-project/archive/rustc/12.0-2021-04-15.tar.gz
-        tar xzf code.tar.gz --strip-components 1 llvm-project-rustc-12.0-2021-04-15/compiler-rt
+        curl -L -o code.tar.gz https://github.com/rust-lang/llvm-project/archive/rustc/13.0-2021-08-08.tar.gz
+        tar xzf code.tar.gz --strip-components 1 llvm-project-rustc-13.0-2021-08-08/compiler-rt
         echo RUST_COMPILER_RT_ROOT=./compiler-rt >> $GITHUB_ENV
       shell: bash
 
--- a/build.rs
+++ b/build.rs
@@ -56,15 +56,10 @@ fn main() {
     if !cfg!(feature = "mangled-names") && cfg!(feature = "c") {
         // Don't use a C compiler for these targets:
         //
-        // * wasm - clang for wasm is somewhat hard to come by and it's
-        //   unlikely that the C is really that much better than our own Rust.
         // * nvptx - everything is bitcode, not compatible with mixed C/Rust
         // * riscv - the rust-lang/rust distribution container doesn't have a C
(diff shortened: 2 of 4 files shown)


bjorn3:
wasm32-unknown-unknown is not ABI compatible with C and currently doesn't depend on a C compiler in any way.

trevyn:
The ABI surface actually used by compiler-rt is compatible with Rust, and per https://github.com/rust-lang/rust/issues/35437, the scope of compiler-builtins is "[each and every compiler-rt intrinsic](https://github.com/rust-lang/compiler-rt/tree/8598065bd965d9713bfafb6c1e766d63a7b17b89/lib/builtins)".

bjorn3:
You need this for linking against sqlite, right? Sqlite is written in C and thus almost certainly abi incompatible with wasm32-unknown-unknown. compiler-builtins should already support everything rust only wasm modules need, and if not, the missing intrinsics should be ported to rust.

I'm not a reviewer of this repo, but IMHO compiler-builtins should remain rust only on wasm32-unknown-unknown until it is actually possible to use C code as user without risking an abi incompatibility and thus we would need the extra intrinsics.

Amanieu:
I'm not too familiar with the wasm situation. cc @alexcrichton

alexcrichton:
There's some more discussion of related intrinsics over at https://github.com/WebAssembly/wasi-sdk/issues/361 as well (linked from https://github.com/rust-lang/compiler-builtins/issues/561). I think there's two questions here:

* Should the `compiler-builtins` crate be reponsible for bringing along intrinsics that Rust itself doesn't use and only C uses? (which I believe is the case here since AFAIK Rust doesn't use `__builtin_sadd_overflow`)
* Should `compiler-builtins` build the C code of the `compiler-rt` library?

For the first question that's a bit beyond myself personally, although lots has changed in the 8 years since the issue you linked was posted and AFAIK there was no "official" decision one way or another when I was working on Rust, but things may have changed in the meantime.

For the second question it's unfortunately a bit tied up in the first as well. If compiler-builtins provides only Rust-used intrinsics then there's no need for C because I think compiler-builtins already supplies everything. There might still be performance differences but it seems that's not the question here.

Otherwise if Rust's compiler-builtins does want to provide intrinsics it doesn't use, but instead C might use (as I believe is the case here), then I'm at least personally less worried than @bjorn3 about the ABI issues here. While it's true wasm32-unknown-unknown is not exactly matching clang it's not like every signature is irrationally different. It should be possible to manually confirm that the signatures here all match and since these aren't general ABIs, just fixed ones, it should work out ok. Additionally as @trevyn points out the comment here is indeed outdated, so there shouldn't be any issue compiling C intrinsics if requested, although they should probably still be opt-in since folks compiling for wasm32-unknown-unknown often don't have a clang lying around.

PR #572, opened 2024-01-31
Remove 80-bit builtins entirely

nikic:
It turns out that these also don't build on x86 + MSVC. Rather than fixing up the condition, I'm just deleting them entirely. As far as I know, Rust does not support 80-bit floats and has no plan to support them, so we shouldn't need these builtins at all.


nikic:
After this patch there is still a build failure in floatdixf.c. Will fix it tomorrow.

PR #575, opened 2024-02-10
Sync x86 chkstk intrinsics with LLVM

kleisauke:
Incorporates the following commits:
https://github.com/llvm/llvm-project/commit/885d7b759b5c166c07c07f4c58c6e0ba110fb0c2
https://github.com/llvm/llvm-project/commit/1f9eff100ce8faea1284d68b779d844c6e019b77
https://github.com/llvm/llvm-project/commit/7a5cba8bea8f774d48db1b0426bcc102edd2b69f

### DIFF
--- a/src/x86.rs
+++ b/src/x86.rs
@@ -6,7 +6,6 @@ use core::intrinsics;
 // calling convention which can't be implemented using a normal Rust function
 
 // NOTE These functions are never mangled as they are not tested against compiler-rt
-// and mangling ___chkstk would break the `jmp ___chkstk` instruction in __alloca
 
 intrinsics! {
     #[naked]
@@ -15,50 +14,8 @@ intrinsics! {
         target_env = "gnu",
         not(feature = "no-asm")
     ))]
-    pub unsafe extern "C" fn ___chkstk_ms() {
-        core::arch::asm!(
-            "push   %ecx",
-            "push   %eax",
-            "cmp    $0x1000,%eax",
-            "lea    12(%esp),%ecx",
-            "jb     1f",
-            "2:",
-            "sub    $0x1000,%ecx",
-            "test   %ecx,(%ecx)",
-            "sub    $0x1000,%eax",
-            "cmp    $0x1000,%eax",
-            "ja     2b",
-            "1:",
-            "sub    %eax,%ecx",
-            "test   %ecx,(%ecx)",
-            "pop    %eax",
-            "pop    %ecx",
-            "ret",
-            options(noreturn, att_syntax)
-        );
-    }
-
-    // FIXME: __alloca should be an alias to __chkstk
-    #[naked]
-    #[cfg(all(
-        windows,
-        target_env = "gnu",
-        not(feature = "no-asm")
-    ))]
-    pub unsafe extern "C" fn __alloca() {
-        core::arch::asm!(
-            "jmp ___chkstk", // Jump to ___chkstk since fallthrough may be unreliable"
-            options(noreturn, att_syntax)
-        );
-    }
-
-    #[naked]
-    #[cfg(all(
-        windows,
-        target_env = "gnu",
-        not(feature = "no-asm")
-    ))]
-    pub unsafe extern "C" fn ___chkstk() {
+    pub unsafe extern "C" fn _alloca() {
+        // _chkstk and _alloca are the same function
         core::arch::asm!(
             "push   %ecx",
             "cmp    $0x1000,%eax",
--- a/README.md
+++ b/README.md
@@ -163,7 +163,6 @@ features = ["c"]
 - [ ] i386/ashldi3.S
 - [ ] i386/ashrdi3.S
 - [x] i386/chkstk.S
-- [x] i386/chkstk2.S
 - [ ] i386/divdi3.S
 - [ ] i386/lshrdi3.S
 - [ ] i386/moddi3.S
@@ -192,7 +191,6 @@ features = ["c"]
 - [x] umoddi3.c
 - [x] umodsi3.c
 - [x] x86_64/chkstk.S
-- [x] x86_64/chkstk2.S
 
 These builtins are needed to support 128-bit integers, which are in the process of being added to Rust.
 
--- a/src/x86_64.rs
+++ b/src/x86_64.rs
@@ -6,7 +6,6 @@ use core::intrinsics;
 // calling convention which can't be implemented using a normal Rust function
 
 // NOTE These functions are never mangled as they are not tested against compiler-rt
-// and mangling ___chkstk would break the `jmp ___chkstk` instruction in __alloca
 
 intrinsics! {
     #[naked]
@@ -32,49 +31,6 @@ intrinsics! {
             "test   %rcx,(%rcx)",
             "pop    %rax",
             "pop    %rcx",
-            "ret",
-            options(noreturn, att_syntax)
-        );
-    }
-
-    #[naked]
-    #[cfg(all(
-        any(all(windows, target_env = "gnu"), target_os = "uefi"),
-        not(feature = "no-asm")
-    ))]
-    pub unsafe extern "C" fn __alloca() {
-        core::arch::asm!(
-            "mov    %rcx,%rax", // x64 _alloca is a normal function with parameter in rcx
-            "jmp    ___chkstk", // Jump to ___chkstk since fallthrough may be unreliable"
-            options(noreturn, att_syntax)
-        );
-    }
-
-    #[naked]
-    #[cfg(all(
-        any(all(windows, target_env = "gnu"), target_os = "uefi"),
-        not(feature = "no-asm")
-    ))]
-    pub unsafe extern "C" fn ___chkstk() {
-        core::arch::asm!(
-            "push   %rcx",
-            "cmp    $0x1000,%rax",
-            "lea    16(%rsp),%rcx", // rsp before calling this routine -> rcx
-            "jb     1f",
-            "2:",
-            "sub    $0x1000,%rcx",
-            "test   %rcx,(%rcx)",
-            "sub    $0x1000,%rax",
-            "cmp    $0x1000,%rax",
-            "ja     2b",
-            "1:",
-            "sub    %rax,%rcx",
-            "test   %rcx,(%rcx)",
-            "lea    8(%rsp),%rax",  // load pointer to the return address into rax
-            "mov    %rcx,%rsp",     // install the new top of stack pointer into rsp
-            "mov    -8(%rax),%rcx", // restore rcx
-            "push   (%rax)",        // push return address onto the stack
-            "sub    %rsp,%rax",     // restore the original value in rax
             "ret",
             options(noreturn, att_syntax)
         );


kleisauke:
For reference, this resolves this linker failure on `i686-pc-windows-gnullvm` with LLVM 18.1.0-rc2:
```
ld.lld: error: undefined symbol: ___chkstk
>>> referenced by librsvg_c_api.a(compiler_builtins-3a76121a5ec728e8.compiler_builtins.3baf6562b232e6be-cgu.0.rcgu.o):(.weak.___alloca.default.__ZN17compiler_builtins3int6addsub7UAddSub4uadd17hff4d6b09d2fec294E)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

/cc @mati865 FYI.

mati865:
Thank, looks good at the quick glance but I don't understand why it'd fail with rc2 (unless it's rc2 with llvm/llvm-project@7a5cba8 backported).

kleisauke:
Commit https://github.com/llvm/llvm-project/commit/248aeac1ad2cf4f583490dd1312a5b448d2bb8cc was cherry-picked/backported as https://github.com/llvm/llvm-project/commit/7a5cba8bea8f774d48db1b0426bcc102edd2b69f in LLVM 18.1.0-rc2. 

However, it's likely that LLVM 18.1.0-rc1 would also encounter the same linker failure on i386, since commit https://github.com/llvm/llvm-project/commit/885d7b759b5c166c07c07f4c58c6e0ba110fb0c2 removed the misspelled form of the MSVC `_chkstk` function, with three underscores instead of two.

mati865:
Oh, GitHub for some reason didn't show this commit belongs to rc2 tag and I assumed it'll be part of rc3. Makes sense now.

mati865 (approved):

mati865:
CI should be fixed in today's nightly:  https://github.com/rust-lang/rust/issues/121552

Can you somehow trigger the build?

kleisauke:
Great! I just triggered the build via:
```console
$ git rebase --force-rebase upstream/master
```

kleisauke:
... it looks like this is blocked on https://github.com/rust-lang/compiler-builtins/pull/582.

Amanieu:

src/x86.rs
```diff
@@ -6,59 +6,16 @@ use core::intrinsics;
...
-    }
-
-    // FIXME: __alloca should be an alias to __chkstk
     #[naked]
     #[cfg(all(
         windows,
         target_env = "gnu",
         not(feature = "no-asm")
     ))]
     pub unsafe extern "C" fn __alloca() {
```
Amanieu, line 17: Are you sure that this is correct? rustc's name mangling will add another underscore in front of this function, so the final symbol will be `___alloca`.

kleisauke: Good point, this should of course be `_alloca`. 🤦‍♂️

Commit d32d5eeeb565c3c6f4288976e4bc07d82d9694fd fixes this.

jeremyd2019:
This seems to have broken MSYS2's build of i686-pc-windows-gnu (with gcc):
```
    = note: D:/M/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/13.2.0/../../../../i686-w64-mingw32/bin/ld.exe: D:/M/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/13.2.0/libgcc.a(_chkstk.o):(.text+0x0): multiple definition of `_alloca'; C:\_\B\src\MINGW32\build\i686-pc-windows-gnu\stage0-sysroot\lib\rustlib\i686-pc-windows-gnu\lib/libstd-394a669d59370585.dll.a(std_394a669d59370585_dll_d001695.o):(.text+0x0): first defined here
```

We needed to backport this change to make i686-pc-windows-gnullvm (with clang) work, because of an error linking to ___chkstk, but the latest update renaming `_alloca` has caused this error with gcc.

kleisauke:
@jeremyd2019 Ah, bummer. PR https://github.com/rust-lang/compiler-builtins/pull/452 ensures that each intrinsic is compiled into its own separate object file, which would avoid duplicate symbol definition errors. However, IIUC, this approach would only work if `_alloca` is also placed in a separate object file within `libgcc.a`, which does not appear to be the case:
```console
$ ar t i686-w64-mingw32/lib/gcc/i686-w64-mingw32/13.2.0/libgcc.a | grep -E "chkstk|alloca"
_chkstk.o
_chkstk_ms.o
```

Perhaps adding `_alloca` to `LIB1ASMFUNCS` here would fix this?:
https://github.com/gcc-mirror/gcc/blob/releases/gcc-13.2.0/libgcc/config/i386/t-chkstk#L2
(untested)

Otherwise, I think marking `_alloca` as a weak symbol (with the `#[weak]` attribute) would fix this as well.

jeremyd2019:
> @jeremyd2019 Ah, bummer. PR #452 ensures that each intrinsic is compiled into its own separate object file, which would avoid duplicate symbol definition errors. However, IIUC, this approach would only work if `_alloca` is also placed in a separate object file within `libgcc.a`, which does not appear to be the case

I just applied the changes from this PR on top of whatever version of compiler_builtins was vendored into rust 1.77.1.  I can try to backport that PR as well, if it hasn't already been included in that version.

> Perhaps adding `_alloca` to `LIB1ASMFUNCS` here would fix this?: https://github.com/gcc-mirror/gcc/blob/releases/gcc-13.2.0/libgcc/config/i386/t-chkstk#L2 (untested)

Hmm, I think there are a lot of GCC versions out there, and requiring a new one (or a patched one) may not be viable.

> Otherwise, I think marking `_alloca` as a weak symbol (with the `#[weak]` attribute) would fix this as well.

Maybe.  Any time I see "weak" symbols it scares me now, we have had a *lot* of bugs/issues with weak symbols in gcc/binutils on Windows.  I *think* this case would be fine though, the issues had to do with shared libraries IIRC.  If the PR you referenced is already present, I can try slapping `#[weak]` on `_alloca` and seeing what happens.  (I don't know much rust, so I'm trusting your syntax :grin:)

As a procedure question, should I have opened a new issue on this rather than commenting on the PR?  I usually like to comment on the PR in cases like this so I know that all the relevant people (including the author) are subscribed to it.

jeremyd2019:
I guess I missed something:
```
error: cannot find attribute `weak` in this scope
  --> D:\M\mingw-w64-rust\src\rustc-1.77.1-src\vendor\compiler_builtins\src\x86.rs:12:7
   |
12 |     #[weak]
   |       ^^^^
```

bjorn3:
It would have to be `#[linkage = "weak"]`.

jeremyd2019:
It was inside an `intrinsics!` block, but `#[linkage = "weak"]` seems to be going further.

jeremyd2019:
Nope:
```
  = note: D:/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/13.2.0/../../../../i686-w64-mingw32/bin/ld.exe: D:/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/13.2.0/libgcc.a(_chkstk.o):(.text+0x0): multiple definition of `_alloca'; D:\M\mingw-w64-rust\src\MINGW32\build\i686-pc-windows-gnu\stage0-sysroot\lib\rustlib\i686-pc-windows-gnu\lib/libstd-394a669d59370585.dll.a(std_394a669d59370585_dll_d001695.o):(.text+0x0): first defined here
          collect2.exe: error: ld returned 1 exit status
```

jeremyd2019:
At least for us, everything worked while the function was called `__alloca` rather than `_alloca`.  Did renaming it cause some other previously-broken scenario to start working?

kleisauke:
> UPDATE: it does appear to already be present, at least `_alloca` is wrapped in `intrinsics!`

Yes exactly, sorry for the confusion.

> I think there are a lot of GCC versions out there, and requiring a new one (or a patched one) may not be viable.

Indeed, marking `_alloca` as a weak symbol is probably better.

> ```
> error: cannot find attribute `weak` in this scope
>   --> D:\M\mingw-w64-rust\src\rustc-1.77.1-src\vendor\compiler_builtins\src\x86.rs:12:7
>    |
> 12 |     #[weak]
>    |       ^^^^
> ```

I think adding `#[weak]` above `#[naked]` would make it work (i.e. on line 11 instead).

> If it was never called because it had the "wrong" name, perhaps it is not needed at all? It was removed from x86_64.rs after all.

According to commit https://github.com/llvm/llvm-project/commit/885d7b759b5c166c07c07f4c58c6e0ba110fb0c2 only the symbols named `__alloca` on i386 and `___chkstk_ms` on x86_64 are utilized on MinGW; the remaining functions are not used and can lead to confusion. On i386, this is the raw symbol name - if we consider a C-level function name with the additional implicit leading underscore, it would be called `_alloca`.

If the `#[weak]` trick doesn't work and we can safely assume the `i686-pc-windows-gnu` target will always link against libgcc, then we could guard the `_alloca` symbol only for `target_abi = "llvm"` (i.e. making it only available for the `i686-pc-windows-gnullvm` target).

jeremyd2019:
> > ```
> > error: cannot find attribute `weak` in this scope
> >   --> D:\M\mingw-w64-rust\src\rustc-1.77.1-src\vendor\compiler_builtins\src\x86.rs:12:7
> >    |
> > 12 |     #[weak]
> >    |       ^^^^
> > ```
> 
> I think adding `#[weak]` above `#[naked]` would make it work (i.e. on line 11 instead).

I will try that, but don't hold out much hope of it working because there is in fact a dll import library involved.  As I recall, Windows PE format does not have weak shared library symbols as ie ELF does, so this cannot really be expected to work.  Perhaps @mstorsjo could elaborate?

> If the `#[weak]` trick doesn't work and we can safely assume the `i686-pc-windows-gnu` target will always link against libgcc, then we could guard the `_alloca` symbol only for `target_abi = "llvm"` (i.e. making it only available for the `i686-pc-windows-gnullvm` target).

I don't know about that assumption.  I only deal with rust where there is a C toolchain for the target in question already present.  I assume rust supports situations where that's not the case (otherwise, what's the point of this crate? why not always just link against the intrinsics from the C toolchain?)

jeremyd2019:
> I think adding `#[weak]` above `#[naked]` would make it work (i.e. on line 11 instead).

This compiles, but still fails linking with the same multiple definition of `_alloca` error as before.

mstorsjo:
> I will try that, but don't hold out much hope of it working because there is in fact a dll import library involved. As I recall, Windows PE format does not have weak shared library symbols as ie ELF does, so this cannot really be expected to work. Perhaps @mstorsjo could elaborate?

I didn't try to read up the whole discussion above to see which semantic of weak symbols this refers to.

For static linking, LLVM/lld (and to some extend also GCC/binutils, although it's buggy in some cases) does support weak symbols in COFF - you can have a weak undefined, which resolves as a null pointer if there's no definition of the symbol, or you can have multiple weak definitions and zero/one non-weak definitions.

This works for symbols within one module, but for symbols imported from a DLL, it gets resolved at link time, and if referenced, the symbol must exist at runtime when the module is loaded too, there's no weak behaviour left at runtime. (There's a mechanism for delay loading, but that just shifts the runtime cost of loading a DLL, it does not allow the symbol to be missing. And I guess this isn't the behaviour you're discussing above anyway.)

jeremyd2019:
That's how I understood weak symbols to work, that it meant at reference time.  This discussion came about because of the idea that marking the symbol (one of the symbols) as weak would fix a multiple definition error at link time.  That seems to be a different semantic of weak symbols that I was not familiar with.

jeremyd2019:
> If the `#[weak]` trick doesn't work and we can safely assume the `i686-pc-windows-gnu` target will always link against libgcc, then we could guard the `_alloca` symbol only for `target_abi = "llvm"` (i.e. making it only available for the `i686-pc-windows-gnullvm` target).

This probably wouldn't help in our particular case, we are *still* patching rust to build the {i686,x86_64}-pc-windows-gnu targets but actually be -gnullvm (not sure if I explained that well).  We cannot switch to *not* lie about the target yet, due to microsoft/windows-rs#2960

PR #577, opened 2024-02-19
Always have math functions but with `weak` linking attribute if we can

Amjad50:
This is a replacement for https://github.com/rust-lang/libm/pull/290

This fixes crashes during compilations for targets that don't have math symbols by default.

So, we will provide them libm symbols, but mark it as `weak` (if its supported), so that the linker will choose the system builtin functions, since those are sometimes more optimized.
If the linker couldn't find those, it will go with `libm` implementation.

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -16,6 +16,7 @@ macro_rules! no_mangle {
     }
 }
 
+#[cfg(all(not(windows), not(target_vendor = "apple")))]
 no_mangle! {
     fn acos(x: f64) -> f64;
     fn asin(x: f64) -> f64;
--- a/src/math.rs
+++ b/src/math.rs
@@ -7,6 +7,7 @@ macro_rules! no_mangle {
     ($(fn $fun:ident($($iid:ident : $ity:ty),+) -> $oty:ty;)+) => {
         intrinsics! {
             $(
+                #[cfg_attr(all(not(windows), not(target_vendor = "apple")), weak)]
                 pub extern "C" fn $fun($($iid: $ity),+) -> $oty {
                     self::libm::$fun($($iid),+)
                 }
@@ -15,17 +16,6 @@ macro_rules! no_mangle {
     }
 }
 
-#[cfg(any(
-    all(
-        target_family = "wasm",
-        target_os = "unknown",
-        not(target_env = "wasi")
-    ),
-    target_os = "xous",
-    target_os = "uefi",
-    all(target_arch = "xtensa", target_os = "none"),
-    all(target_vendor = "fortanix", target_env = "sgx")
-))]
 no_mangle! {
     fn acos(x: f64) -> f64;
     fn asin(x: f64) -> f64;
@@ -41,10 +31,6 @@ no_mangle! {
     fn log10f(x: f32) -> f32;
     fn log(x: f64) -> f64;
     fn logf(x: f32) -> f32;
-    fn fmin(x: f64, y: f64) -> f64;
-    fn fminf(x: f32, y: f32) -> f32;
-    fn fmax(x: f64, y: f64) -> f64;
-    fn fmaxf(x: f32, y: f32) -> f32;
     fn round(x: f64) -> f64;
     fn roundf(x: f32) -> f32;
     fn rint(x: f64) -> f64;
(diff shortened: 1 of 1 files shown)


Amjad50:
For some reason, windows building fails, and the error is confusing, it can't find some symbols that are unrelated to what have changed.

Amjad50:
Can you check this @Amanieu? Thanks

Amanieu:
We're currently waiting for fixes on the rustc side before work on compiler-builtins can continue:  https://github.com/rust-lang/rust/issues/121552

Amjad50:
Ah, thanks for the info.

Amanieu:

Amjad50:
Thanks, CI is fixed now

src/math.rs
```diff
@@ -85,65 +70,40 @@ no_mangle! {
...
 #[cfg(any(
-    all(target_arch = "arm", target_os = "none"),
-    all(target_arch = "riscv32", not(target_feature = "f"), target_os = "none"),
-    all(target_arch = "x86_64", target_os = "none"),
-    all(target_arch = "mips", target_os = "none"),
+    target_arch = "arm",
+    all(target_arch = "riscv32", not(target_feature = "f")),
+    target_arch = "x86_64",
+    target_arch = "mips",
 ))]
```
Amanieu, line 107: I would just remove this cfg entirely and merge it into the block above.

src/math.rs
```diff
@@ -7,6 +7,7 @@ macro_rules! no_mangle {
     ($(fn $fun:ident($($iid:ident : $ity:ty),+) -> $oty:ty;)+) => {
         intrinsics! {
             $(
+                #[cfg_attr(all(not(windows), not(target_vendor="apple")), weak)]
```
Amanieu, line 10:
```suggestion
                #[cfg_attr(all(not(windows), not(target_vendor = "apple")), weak)]
```

src/math.rs
```diff
@@ -85,65 +70,40 @@ no_mangle! {
...
-        not(target_env = "wasi")
-    ),
-    target_os = "xous",
-    target_os = "uefi",
-    all(target_arch = "xtensa", target_os = "none"),
-    all(target_vendor = "fortanix", target_env = "sgx"),
-    target_os = "windows"
-))]
 intrinsics! {
+    #[cfg_attr(all(not(windows), not(target_vendor="apple")), weak)]
```
Amanieu, line 86:
```suggestion
    #[cfg_attr(all(not(windows), not(target_vendor = "apple")), weak)]
```

src/math.rs
```diff
@@ -85,65 +70,40 @@ no_mangle! {
...
     pub extern "C" fn lgamma_r(x: f64, s: &mut i32) -> f64 {
         let r = self::libm::lgamma_r(x);
         *s = r.1;
         r.0
     }
 
+    #[cfg_attr(all(not(windows), not(target_vendor="apple")), weak)]
```
Amanieu, line 93:
```suggestion
    #[cfg_attr(all(not(windows), not(target_vendor = "apple")), weak)]
```

src/math.rs
```diff
@@ -15,17 +16,7 @@ macro_rules! no_mangle {
...
-        target_family = "wasm",
-        target_os = "unknown",
-        not(target_env = "wasi")
-    ),
-    target_os = "xous",
-    target_os = "uefi",
-    all(target_arch = "xtensa", target_os = "none"),
-    all(target_vendor = "fortanix", target_env = "sgx")
-))]
+#[cfg(not(target_os = "windows"))]
```
Amanieu, line 19:
```suggestion
#[cfg_attr(all(not(windows), not(target_vendor = "apple")), weak)]
```

Amjad50: We do have this already inside `no_mangle`, so probably just remove the `cfg` completely is a better choice

Amanieu: No, we shouldn't provide the math intrinsics at all on windows/apple targets since weak linkage doesn't work properly. On those target we should always be getting these symbols from libc. The only exception is `lgamma_r` and `lgammaf_r` which are missing on MSVC targets.

Amjad50: then you probably mean to use `cfg` and not `cfg_attr`

PR #578, opened 2024-02-25
Add a bunch of `rustc_nounwind` for `c_unwind`

DianQK:
This PR fixes encountered dependency core issue in https://github.com/rust-lang/rust/pull/116088.

After enabling `c_unwind`, consider the following code.

```rust
#![feature(c_unwind)]

#[no_mangle]
extern "C" fn foo() {
    unsafe {
        bar();
    }
}

extern "Rust" {
    fn bar();
}
```

Due to the lack of inlining or `rustc_nounwind` for `bar`, the `core::panicking::panic_cannot_unwind` function is called.

```asm
foo:
  push rax
  call qword ptr [rip + bar@GOTPCREL]
  pop rax
  ret
  call qword ptr [rip + core::panicking::panic_cannot_unwind@GOTPCREL]
```

Godbolt: https://rust.godbolt.org/z/Ea4rhfWcr.

In this PR, I'm checking if there are any undefined symbols belonging to `core` in the release build. I believe checking the debug build is meaningless. First, we won't distribute the debug build of `compiler-builtins`, and second, it has a lot of dependencies on `core`.

Since we're already checking for undefined symbols in `core`, I think checking the `intrinsics` example might not be meaningful anymore.

r? @Amanieu

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -42,9 +42,9 @@ if [ "$NM" = "" ]; then
 fi
 
 if [ -d /target ]; then
-    path=/target/${1}/debug/deps/libcompiler_builtins-*.rlib
+    path=/target/${1}/release/deps/libcompiler_builtins-*.rlib
 else
-    path=target/${1}/debug/deps/libcompiler_builtins-*.rlib
+    path=target/${1}/release/deps/libcompiler_builtins-*.rlib
 fi
 
 # Look out for duplicated symbols when we include the compiler-rt (C) implementation
@@ -93,15 +93,15 @@ if [ -z "$DEBUG_LTO_BUILD_DOESNT_WORK" ]; then
   RUSTFLAGS="-C debug-assertions=no" \
     CARGO_INCREMENTAL=0 \
     CARGO_PROFILE_DEV_LTO=true \
-    $cargo rustc --features "$INTRINSICS_FEATURES" --target $1 --example intrinsics
+    $cargo rustc --features "$INTRINSICS_FEATURES" --target $1 --example intrinsics --release
 fi
 CARGO_PROFILE_RELEASE_LTO=true \
   $cargo rustc --features "$INTRINSICS_FEATURES" --target $1 --example intrinsics --release
 
-# Ensure no references to a panicking function
+# Ensure no references to any symbols from core
 for rlib in $(echo $path); do
     set +ex
-    $NM -u $rlib 2>&1 | grep panicking
+    $NM -u $rlib 2>&1 | grep core
 
     if test $? = 0; then
         exit 1
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -80,10 +80,13 @@ rm -f $path
 
 # Verify that we haven't drop any intrinsic/symbol
 build_intrinsics="$cargo build --target $1 -v --example intrinsics"
-RUSTFLAGS="-C debug-assertions=no" $build_intrinsics
-RUSTFLAGS="-C debug-assertions=no" $build_intrinsics --release
-RUSTFLAGS="-C debug-assertions=no" $build_intrinsics --features c
-RUSTFLAGS="-C debug-assertions=no" $build_intrinsics --features c --release
+# Match the behavior with 
+# https://github.com/rust-lang/rust/blob/e49442bf9815a67e68f9d9f2f80560ec1d754b31/Cargo.toml#L66-L86.
+export RUSTFLAGS="-C debug-assertions=no -C overflow-checks=no -C codegen-units=10000"
+$build_intrinsics
+$build_intrinsics --release
+$build_intrinsics --features c
+$build_intrinsics --features c --release
 
 # Verify that there are no undefined symbols to `panic` within our
 # implementations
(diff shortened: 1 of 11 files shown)


bjorn3:
Is mir inlining enough here or does it depend on LLVM's inliner?

DianQK:
> Is mir inlining enough here or does it depend on LLVM's inliner?

I think any one of them should work.

DianQK:
> Is mir inlining enough here or does it depend on LLVM's inliner?

Or are you referring to using `-Zinline-mir-threshold=n`? Or perhaps I missed some configuration information?

DianQK:
Hmm, based on https://github.com/rust-lang/compiler-builtins/actions/runs/8036861350/job/21951067773?pr=578, it seems like besides panic, we also depend on others.

bjorn3:
My question is if the inlining done by the MIR inliner is enough to fix this issue or if it depends on the codegen backend inlining too. AFAIK it is possible for the MIR inliner bail out on `#[inline(always)]` in some cases even when LLVM's inliner would still be able to inline.

DianQK:
> My question is if the inlining done by the MIR inliner is enough to fix this issue or if it depends on the codegen backend inlining too. AFAIK it is possible for the MIR inliner bail out on `#[inline(always)]` in some cases even when LLVM's inliner would still be able to inline.

I might understand now. We want it to work on Cranelift and GCC as well. Let me see if there is a way to experiment with it.

RalfJung:
In case we do not actually want to inline some of these functions, marking them `#[rustc_nounwind]` should also help.

DianQK:
> In case we do not actually want to inline some of these functions, marking them `#[rustc_nounwind]` should also help.

Thanks. That's what I'm looking for. I thought it was written as `extern "rustc-nounwind" fn foo`.

Amanieu:
If I understand correctly, this attribute needs to be applied to all `fn` defined in compiler-builtins? At this point wouldn't it be easier to do this in the compiler by checking `self.tcx.is_compiler_builtins(LOCAL_CRATE)`? This would solve the problem in a single place instead of having attributes peppered around the compiler_builtins code.

ci/run.sh
```diff
@@ -93,15 +96,18 @@ if [ -z "$DEBUG_LTO_BUILD_DOESNT_WORK" ]; then
   RUSTFLAGS="-C debug-assertions=no" \
     CARGO_INCREMENTAL=0 \
     CARGO_PROFILE_DEV_LTO=true \
-    $cargo rustc --features "$INTRINSICS_FEATURES" --target $1 --example intrinsics
+    $cargo rustc --features "$INTRINSICS_FEATURES" --target $1 --example intrinsics --release
```
Amanieu, line 99: This should not have `--release`: this test specifically checks that *debug* builds (`-Copt-level=0`) do not reference any symbols in core. This is needed for people using `-Z build-std` without `--release`.

RalfJung:
> If I understand correctly, this attribute needs to be applied to all fn defined in compiler-builtins? At this point wouldn't it be easier to do this in the compiler by checking self.tcx.is_compiler_builtins(LOCAL_CRATE)? This would solve the problem in a single place instead of having attributes peppered around the compiler_builtins code.

I was thinking of a crate-level version of `#[rustc_nounwind]`. But I also didn't know this crate already was so magic that `tcx` knows about it.

nbdd0121:
> If I understand correctly, this attribute needs to be applied to all `fn` defined in compiler-builtins? At this point wouldn't it be easier to do this in the compiler by checking `self.tcx.is_compiler_builtins(LOCAL_CRATE)`? This would solve the problem in a single place instead of having attributes peppered around the compiler_builtins code.

Good point. I'll have a try with this approach.

RalfJung:
Sounds good! Make sure to only apply it to definitions though, not declarations.

DianQK:
> > If I understand correctly, this attribute needs to be applied to all fn defined in compiler-builtins? At this point wouldn't it be easier to do this in the compiler by checking self.tcx.is_compiler_builtins(LOCAL_CRATE)? This would solve the problem in a single place instead of having attributes peppered around the compiler_builtins code.
> 
> I was thinking of a crate-level version of `#[rustc_nounwind]`. But I also didn't know this crate already was so magic that `tcx` knows about it.

I've tried this and it doesn't seem to work.

DianQK:
> Sounds good! Make sure to only apply it to definitions though, not declarations.

I have found that if I add to trait it causes problems.

ci/run.sh
DianQK: So I understand that we might expect that only `compiler-builtins` is optimized.

RalfJung:
>  >   I was thinking of a crate-level version of #[rustc_nounwind]. But I also didn't know this crate already was so magic that tcx knows about it.
>
> I've tried this and it doesn't seem to work.

Yes, it's not implemented in the compiler (yet). I was proposing to implement it. (Sorry if that wa snot clear).

But @nbdd0121 is on the case already: https://github.com/rust-lang/rust/pull/121605.

DianQK:
> > > I was thinking of a crate-level version of #[rustc_nounwind]. But I also didn't know this crate already was so magic that tcx knows about it.
> > 
> > 
> > I've tried this and it doesn't seem to work.
> 
> Yes, it's not implemented in the compiler (yet). I was proposing to implement it. (Sorry if that was not clear).
> 
> But @nbdd0121 is on the case already: [rust-lang/rust#121605](https://github.com/rust-lang/rust/pull/121605).

We can consider implementing it when we encounter it again.

DianQK:
Prefer https://github.com/rust-lang/rust/pull/121605.

PR #587, opened 2024-04-07
Add missing functions for `f16` and `f128`

### LINKED ISSUE #567: Unable to find long double symbols on `aarch64-apple-darwin`
I am working on https://github.com/rust-lang/rust/pull/114607 and unfortunately can't build on my system due to linker errors:

```
  = note: ld: warning: ignoring duplicate libraries: '-lSystem'
          ld: Undefined symbols:
            ___addtf3, referenced from:
                std::num::test_num::h83194c034ce7125a in std-a7f7c4c812353510.std.3e4ab7aa3f61e492-cgu.12.rcgu.o
            ___divtf3, referenced from:
                std::num::test_num::h83194c034ce7125a in std-a7f7c4c812353510.std.3e4ab7aa3f61e492-cgu.12.rcgu.o
            ___multf3, referenced from:
                std::num::test_num::h83194c034ce7125a in std-a7f7c4c812353510.std.3e4ab7aa3f61e492-cgu.12.rcgu.o
            ___subtf3, referenced from:
                std::num::test_num::h83194c034ce7125a in std-a7f7c4c812353510.std.3e4ab7aa3f61e492-cgu.12.rcgu.o
            ___trunctfdf2, referenced from:
                _$LT$$RF$T$u20$as$u20$core..fmt..Debug$GT$::fmt::h8dd8a3b2b4bc54ac in std-a7f7c4c812353510.std.3e4ab7aa3f61e492-cgu.06.rcgu.o
                core::fmt::float::_$LT$impl$u20$core..fmt..Debug$u20$for$u20$f128$GT$::fmt::h0686fe3ef7164d45 in std-a7f7c4c812353510.std.3e4ab7aa3f61e492-cgu.10.rcgu.o
            ___unordtf2, referenced from:
                std::num::test_num::h83194c034ce7125a in std-a7f7c4c812353510.std.3e4ab7aa3f61e492-cgu.12.rcgu.o
                std::num::test_num::h83194c034ce7125a in std-a7f7c4c812353510.std.3e4ab7aa3f61e492-cgu.12.rcgu.o
                std::num::test_num::h83194c034ce7125a in std-a7f7c4c812353510.std.3e4ab7aa3f61e492-cgu.12.rcgu.o
                std::num::test_num::h83194c034ce7125a in std-a7f7c4c812353510.std.3e4ab7aa3f61e492-cgu.12.rcgu.o
                std::num::test_num::h83194c034ce7125a in std-a7f7c4c812353510.std.3e4ab7aa3f61e492-cgu.12.rcgu.o
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
---
          ld: Undefined symbols:
            ___extenddftf2, referenced from:
                __RNvNtCsgSq1suy33ib_12clippy_utils6consts19lit_to_mir_constant in libclippy_utils-52261686d0193f09.rlib[10](clippy_utils-52261686d0193f09.clippy_utils.c495e003dc0e053d-cgu.07.rcgu.o)
            ___trunctfdf2, referenced from:
                __RINvXs1_NtCsgSq1suy33ib_12clippy_utils6constsNtB6_8ConstantNtNtCsdSdPP26wXhD_4core4hash4Hash4hashNtCsgzWzvnNrs6k_10rustc_hash8FxHasherEB8_ in libclippy_utils-52261686d0193f09.rlib[10](clippy_utils-52261686d0193f09.clippy_utils.c495e003dc0e053d-cgu.07.rcgu.o)
```

It seems like a fix similar to https://github.com/rust-lang/libc/pull/1034 may work, I can submit a PR if this seems correct. @Amanieu

tgross35:
Fixes #567

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,9 +232,9 @@ These builtins are needed to support 128-bit integers.
 
 These builtins are needed to support `f16` and `f128`, which are in the process of being added to Rust.
 
-- [ ] addtf3.c
-- [ ] comparetf2.c
-- [ ] divtf3.c
+- [x] addtf3.c
+- [x] comparetf2.c
+- [x] divtf3.c
 - [x] extenddftf2.c
 - [x] extendhfsf2.c
 - [x] extendhftf2.c
@@ -249,13 +249,13 @@ These builtins are needed to support `f16` and `f128`, which are in the process 
 - [ ] floatsitf.c
 - [ ] floatunditf.c
 - [ ] floatunsitf.c
-- [ ] multf3.c
+- [x] multf3.c
 - [ ] powitf2.c
 - [ ] ppc/fixtfdi.c
 - [ ] ppc/fixunstfdi.c
 - [ ] ppc/floatditf.c
 - [ ] ppc/floatunditf.c
-- [ ] subtf3.c
+- [x] subtf3.c
 - [x] truncdfhf2.c
 - [x] truncsfhf2.c
 - [x] trunctfdf2.c
--- a/src/float/add.rs
+++ b/src/float/add.rs
@@ -1,5 +1,5 @@
 use crate::float::Float;
-use crate::int::{CastInto, Int};
+use crate::int::{CastInto, Int, MinInt};
 
 /// Returns `a + b`
 fn add<F: Float>(a: F, b: F) -> F
@@ -57,9 +57,9 @@ where
         }
 
         // zero + anything = anything
-        if a_abs == Int::ZERO {
+        if a_abs == MinInt::ZERO {
             // but we need to get the sign right for zero + zero
-            if b_abs == Int::ZERO {
+            if b_abs == MinInt::ZERO {
                 return F::from_repr(a.repr() & b.repr());
             } else {
                 return b;
@@ -67,7 +67,7 @@ where
         }
 
         // anything + zero = anything
-        if b_abs == Int::ZERO {
+        if b_abs == MinInt::ZERO {
             return a;
         }
     }
@@ -113,10 +113,10 @@ where
     // Shift the significand of b by the difference in exponents, with a sticky
     // bottom bit to get rounding correct.
     let align = a_exponent.wrapping_sub(b_exponent).cast();
-    if align != Int::ZERO {
+    if align != MinInt::ZERO {
         if align < bits {
             let sticky =
-                F::Int::from_bool(b_significand << bits.wrapping_sub(align).cast() != Int::ZERO);
+                F::Int::from_bool(b_significand << bits.wrapping_sub(align).cast() != MinInt::ZERO);
             b_significand = (b_significand >> align.cast()) | sticky;
         } else {
             b_significand = one; // sticky; b is known to be non-zero.
@@ -125,8 +125,8 @@ where
     if subtraction {
         a_significand = a_significand.wrapping_sub(b_significand);
         // If a == -b, return +zero.
-        if a_significand == Int::ZERO {
-            return F::from_repr(Int::ZERO);
+        if a_significand == MinInt::ZERO {
+            return F::from_repr(MinInt::ZERO);
         }
 
         // If partial cancellation occured, we need to left-shift the result
@@ -143,8 +143,8 @@ where
 
         // If the addition carried up, we need to right-shift the result and
         // adjust the exponent:
-        if a_significand & implicit_bit << 4 != Int::ZERO {
-            let sticky = F::Int::from_bool(a_significand & one != Int::ZERO);
+        if a_significand & implicit_bit << 4 != MinInt::ZERO {
+            let sticky = F::Int::from_bool(a_significand & one != MinInt::ZERO);
             a_significand = a_significand >> 1 | sticky;
             a_exponent += 1;
         }
@@ -160,7 +160,7 @@ where
         // need to shift the significand.
         let shift = (1 - a_exponent).cast();
         let sticky =
-            F::Int::from_bool((a_significand << bits.wrapping_sub(shift).cast()) != Int::ZERO);
+            F::Int::from_bool((a_significand << bits.wrapping_sub(shift).cast()) != MinInt::ZERO);
         a_significand = a_significand >> shift.cast() | sticky;
         a_exponent = 0;
     }
--- a/src/float/add.rs
+++ b/src/float/add.rs
@@ -203,6 +203,10 @@ intrinsics! {
         add(a, b)
     }
 
+    pub extern "C" fn __addtf3(a: f128, b: f128) -> f128 {
+        add(a, b)
+    }
+
     #[cfg(target_arch = "arm")]
     pub extern "C" fn __addsf3vfp(a: f32, b: f32) -> f32 {
         a + b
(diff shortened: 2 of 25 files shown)


README.md
```diff
@@ -213,16 +213,22 @@ These builtins are needed to support 128-bit integers, which are in the process
 - [x] udivti3.c
 - [x] umodti3.c
 
+These builtins are needed to support 16-bit floating point numbers, which are
```
RalfJung, line 216: And 128-bit?

tgross35: Yes, there Is nothing for 16-bit 🙃

tgross35:
@Amanieu this will be the corresponding update to builtins once https://github.com/rust-lang/rust/pull/122470 merges. How do I need to gate things here since it will only work with a recent nightly?

Amanieu:
compiler-builtins is effectively an internal implementation detail of rustc, you're not meant to use it directly. So it therefore only supports the latest rustc anyways.

However you may have problems with bootstrap since that uses the latest beta which won't have support for the `f16`/`f128` types. I believe `cfg(bootstrap)` should work for dependencies of the standard library though.

README.md
```diff
@@ -264,7 +269,6 @@ These builtins involve floating-point types ("`f128`", "`f80`" and complex numbe
 - ~~ppc/gcc_qmul.c~~
 - ~~ppc/gcc_qsub.c~~
 - ~~ppc/multc3.c~~
-- ~~subtf3.c~~
 - ~~trunctfdf2.c~~
 - ~~trunctfsf2.c~~
```
Amanieu, line 273: Are you sure these aren't needed? I would expect all the `tf` builtins to be needed for `f128`.

tgross35:
Awesome, thanks for the information. Luckily compiler support did get merged right before the beta branch so I have been able to avoid `cfg(not(bootstrap))`ing in most places.

README.md
tgross35: They are, I just wanted to get arithmetic passing before working on those.

tgross35:
What is the best way to test this locally? For everything I have tried, the macros at https://github.com/rust-lang/compiler-builtins/blob/72408497f6646f4dd2e903e00a66083d1f548c5c/testcrate/tests/lse.rs#L94-L100 can't be located and running `./ci/run-docker.sh` doesn't find the right linker. Then the tests at https://github.com/tgross35/compiler-builtins/blob/7d347bb9e348e070a87b4123a29c3fdd796d4dee/testcrate/tests/addsub.rs#L112 fail because the symbols to compare against aren't available on my machine (which is aarch64, maybe I just need to gate this?).

Looking forward to being able to get some output from CI but I'm still waiting on nightly to catch up 🙂

Amanieu:
Normally `run-docker.sh` just works, but I guess nobody has tried running it on aarch64. It works fine for me when run on an x86_64 host.

If you're feeling adventurous, you can try fixing the tests to work on an aarch64 host.

Regarding the LSE issues, I think that may be because you are on an Apple system where these symbols don't exist. That test should be gated on `target_os = "linux"`.

tgross35:
The linker just wasn't available on ubuntu 18, but I had to add a way to not use host rustc. Opened https://github.com/rust-lang/compiler-builtins/pull/591 to add a way to use a different image

src/float/add.rs
```diff
@@ -201,6 +201,10 @@ intrinsics! {
         add(a, b)
     }
 
+    pub extern "C" fn __addtf3(a: f128, b: f128) -> f128 {
+        add(a, b)
+    }
```
bjorn3, line 206: This will unconditionally codegen f16 and f128 usage, breaking cg_clif and cg_gcc. Maybe add a cargo feature to disable these new intrinsics?

tgross35:
What is the reason tests don't seem to get run on these platforms? (aarch, i586, thumb, wasm)

image

tgross35:
My widening function is failing at "0xffffffffffffffffffffffffffffffff * 0xffffffffffffffffffffffffffffffff = 0xfffffffffffffffffffffffffffffffe00000000000000000000000000000001 got 0xfffffffffffffffffffffffffffffffd00000000000000000000000000000001", https://github.com/rust-lang/compiler-builtins/actions/runs/8717331100/job/23912367209?pr=587#step:9:370. This seems to be identical to what LLVM's `wideMultiply` does, https://clang.godbolt.org/z/Y4vMxd6se, but from Python it seems like the upper half should end with an `e` rather than a `d`:

```python
>>> hex(0xffffffffffffffffffffffffffffffff * 0xffffffffffffffffffffffffffffffff)
'0xfffffffffffffffffffffffffffffffe00000000000000000000000000000001'
```

Relevant code if anyone has any ideas https://github.com/rust-lang/compiler-builtins/blob/3cc63552b5eb4ee4791bd6b739c61e4f2bfc60bb/src/int/big.rs#L233-L284

marthadev:
> My widening function is failing at "0xffffffffffffffffffffffffffffffff * 0xffffffffffffffffffffffffffffffff = 0xfffffffffffffffffffffffffffffffe00000000000000000000000000000001 got 0xfffffffffffffffffffffffffffffffd00000000000000000000000000000001", https://github.com/rust-lang/compiler-builtins/actions/runs/8717331100/job/23912367209?pr=587#step:9:370. This seems to be identical to what LLVM's `wideMultiply` does, https://clang.godbolt.org/z/Y4vMxd6se ([original source](https://github.com/llvm/llvm-project/blob/49b209d0d1833a339e66735e1288c1805224603e/compiler-rt/lib/builtins/fp_lib.h#L145-L185)), but from Python it seems like the upper half should end with an `e` rather than a `d`:

AFAICT the code is dropping a carry with the `wrapping_add`

> https://github.com/rust-lang/compiler-builtins/blob/3cc63552b5eb4ee4791bd6b739c61e4f2bfc60bb/src/int/big.rs#L269-L276

Should probably be something like:

```
    let (lo, carry) = r0.overflowing_add(r1 << 64);
    let hi = (r1 >> 64)
        + (sum1 >> 96)
        + (sum2 >> 64)
        + (sum3 >> 32)
        + sum4
        + (sum5 << 32)
        + (sum6 << 64)
        + u128::from(carry);
```

tgross35:
Thanks @marthadev, I just today realized that we have a nightly `carrying_add` too to make this even easier. Any idea if what LLVM has is also incorrect then, or am I missing something in that comparison?

marthadev:
> I just today realized that we have a nightly `carrying_add`

TIL, it does indeed look nice

> Any idea if what LLVM has is also incorrect then, or am I missing something in that comparison?

The LLVM implementation seems to have the same problem, which honestly surprises me given that the code is [10 years old](https://github.com/llvm/llvm-project/commit/ea1f8fb775598bfcd7595d49cbc2366db8922ecd). I have no clue though how widely that function is being used, but having a logic bug in integer arithmetic seems pretty bad, so I am glad you found it.

tgross35:
I opened an issue at https://github.com/llvm/llvm-project/issues/90009 for the LLVM side to take a look at, thanks for confirming.

tgross35:
Everything except division has been cleaned up and split off to https://github.com/rust-lang/compiler-builtins/pull/606

tgross35:
Superceded by a few PRs, many have merged already.

PR #590, opened 2024-04-10
Add Apple visionOS support

QuentinPerez:


Amanieu:
Can we change all of these to check target_vendor instead? For the ARM-specific ones just using target_vendor should be fine as well.

PR #591, opened 2024-04-13
Add a way to test more targets on non-linux machines

tgross35:
Allow using the `rust-lang/rust:nightly` docker image to run tests in cases where the host rust and cargo cannot be used, such as non-linux hosts.

Additionally, prefix image tags and the target directory with `builtins-` to reduce the chance of an accidental conflict, and fix shellcheck warnings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 *.rs.bk
 Cargo.lock
 target
+compiler-rt
+*.tar.gz
--- a/README.md
+++ b/README.md
@@ -78,6 +78,26 @@ features = ["c"]
 [8]: http://en.cppreference.com/w/cpp/language/implicit_conversion
 [9]: https://doc.rust-lang.org/std/primitive.i32.html
 
+## Testing
+
+The easiest way to test locally is using Docker. This can be done by running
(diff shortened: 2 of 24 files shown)


tgross35:
https://github.com/rust-lang/compiler-builtins/actions/runs/8672263156/job/23782359199?pr=591#step:8:2321 I have no clue why powerpc would have started failing because of this

tgross35:
I guess the error was expected (?) based on previous runs https://github.com/rust-lang/compiler-builtins/actions/runs/8672321314/job/23782498611?pr=592#step:8:2165, and I just accidentally enabled it. Since powerpc is the only architecture that fails, I made it an xfail but left the tests enabled.

.gitignore
```diff
@@ -1,3 +1,6 @@
 *.rs.bk
 Cargo.lock
 target
+target-docker
+compiler-rt
+*.tar.gz
```
Amanieu, line 6: I don't think these changes should be committed. Maybe only the compiler-rt one?

tgross35: Ah, `target-docker` was from before I changed it to just use a docker volume. Removed this.

tar.gz is just there so if you download the C libraries like in CI it won't get picked up. Is this okay to keep?

PR #592, opened 2024-04-13
Add a fix for aarch64-darwin and enable it in CI

tgross35:
This includes two commits:

###  Change aarch64_linux module and lse tests to have the same gating

Trying to run testcrate on non-linux aarch64 currently hits a compilation error. Make this test linux-only, to be consistent with the `aarch64_linux` module that it depends on.

Additionally, enable the `aarch64_linux` module for `target_arch = "arm64ec"` to be the same as these tests.

### Add CI testing for AArch64 Darwin

The Apple ARM silicon has been around for a while now and hopefully will become Rust Tier 1 at some point. Add it to CI since it is distinct enough from aarch64-linux and x86_86-darwin that there may be differences.

### DIFF
--- a/testcrate/tests/lse.rs
+++ b/testcrate/tests/lse.rs
@@ -1,8 +1,5 @@
 #![feature(decl_macro)] // so we can use pub(super)
-#![cfg(all(
-    any(target_arch = "aarch64", target_arch = "arm64ec"),
-    not(feature = "no-asm")
-))]
+#![cfg(all(target_arch = "aarch64", target_os = "linux", not(feature = "no-asm")))]
 
 /// Translate a byte size to a Rust type.
 macro int_ty {
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,6 +9,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
+        - target: aarch64-apple-darwin
+          os: macos-14
+          rust: nightly
         - target: aarch64-unknown-linux-gnu
           os: ubuntu-latest
           rust: nightly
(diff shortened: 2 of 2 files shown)


tgross35:
@Amanieu this PR is pretty small, would you be able to merge it? It would just make it easier for me to develop/test locally without carrying an extra patch around.

src/lib.rs
```diff
@@ -62,7 +62,11 @@ pub mod arm;
 #[cfg(any(target_arch = "aarch64", target_arch = "arm64ec"))]
 pub mod aarch64;
 
-#[cfg(all(target_arch = "aarch64", target_os = "linux", not(feature = "no-asm"),))]
+#[cfg(all(
+    any(target_arch = "aarch64", target_arch = "arm64ec"),
```
Amanieu, line 66: ARM64EC only exists on Windows.

testcrate/tests/lse.rs
```diff
@@ -1,6 +1,7 @@
 #![feature(decl_macro)] // so we can use pub(super)
 #![cfg(all(
     any(target_arch = "aarch64", target_arch = "arm64ec"),
```
Amanieu, line 3: You can also remove arm64ec here since it's windows-only.

src/lib.rs
tgross35: Removed this

testcrate/tests/lse.rs
tgross35: Removed it here too

src/lib.rs
tgross35: Actually, would it be better to leave arm64ec but change the cfg to `not(target_os = "macos")` so it still gets tested on Windows? I don't know where this is needed but it must have been passing okay on the windows platforms.

tgross35:
Thanks!

PR #593, opened 2024-04-14
Add builtins for `f16`/`f128` float conversions

beetrees:
Adds builtins for float conversions to/from `f16`/`f128`, which will solve rust-lang/rust#123885. I also updated the `README.md` to reflect that `f16`/`f128` are being added to Rust.

@tgross35 this should save you some work on #587.

I'm not sure if the new functions need the `#[avr_skip]` attribute like the existing extend/truncate builtins do, but thought it was better to add it and accidentally cause linker errors if the builtin is needed and not supplied by `libgcc` rather than not having the attribute and accidentally causing subtle ABI bugs if a special ABI is required on AVR. @Patryk27 who added the original `#[avr_skip]`s.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,10 @@ c = ["cc"]
 # which use inline assembly and fall back to pure Rust versions (if avalible).
 no-asm = []
 
+# Workaround for codegen backends which haven't yet implemented `f16` and
+# `f128` support. Disabled any intrinsics which use those types.
+no-f16-f128 = []
+
 # Flag this library as the unstable compiler-builtins lib
 compiler-builtins = []
 
--- a/src/float/trunc.rs
+++ b/src/float/trunc.rs
@@ -52,8 +52,10 @@ where
         // destination format.  We can convert by simply right-shifting with
         // rounding and adjusting the exponent.
         abs_result = (a_abs >> sign_bits_delta).cast();
-        let tmp = src_exp_bias.wrapping_sub(dst_exp_bias) << R::SIGNIFICAND_BITS;
-        abs_result = abs_result.wrapping_sub(tmp.cast());
+        // Cast before shifting to prevent overflow.
+        let bias_diff: R::Int = src_exp_bias.wrapping_sub(dst_exp_bias).cast();
+        let tmp = bias_diff << R::SIGNIFICAND_BITS;
+        abs_result = abs_result.wrapping_sub(tmp);
 
         let round_bits = a_abs & round_mask;
         if round_bits > halfway {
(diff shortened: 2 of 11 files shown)


Patryk27:
Yeah, I'm not sure on AVR's ABI here, so probably better to keep those `#[avr_skip]` 🙂

tgross35:
Should this also be setting `COMPILER_RT_HAS_FLOAT16` for the C build, in case that winds up being used as a fallback in the future?

tgross35:
Thanks, that part looks great to me. Looks like tests might be missing? Like https://github.com/rust-lang/compiler-builtins/blob/2978bdd1c56f4ed40a7dd3b896f7a93b227ce94b/testcrate/tests/misc.rs#L114-L121 / https://github.com/rust-lang/compiler-builtins/blob/2978bdd1c56f4ed40a7dd3b896f7a93b227ce94b/testcrate/tests/misc.rs#L201-L214

tgross35:
> I've instead added tests comparing `compiler_builtins` float conversions to those in [`rustc_apfloat`](https://crates.io/crates/rustc_apfloat)

Great idea, I can probably do the same on #587. Everything here looks reasonable to my understanding, probably ready for a look from @Amanieu.

testcrate/tests/conv.rs
```diff
@@ -130,3 +140,118 @@ fn float_to_int() {
...
+    use compiler_builtins::float::extend::__extendsfdf2;
+
+    extend!(f32, f64, __extendsfdf2);
+    conv!(f32, f64, __extendsfdf2, Single, Double);
+    #[cfg(not(feature = "no-f16-f128"))]
+    {
+        use compiler_builtins::float::extend::{
+            __extenddftf2, __extendhfsf2, __extendhftf2, __extendsftf2, __gnu_h2f_ieee,
+        };
+        // FIXME: Also do extend!() for `f16` and `f128` when builtins are in nightly
```
tgross35, line 193: Minor nit, could you change these couple fixmes to `FIXME(f16_f128)`? To be consistent with fixmes related to these features in the compiler.

src/float/trunc.rs
```diff
@@ -52,8 +52,10 @@ where
         // destination format.  We can convert by simply right-shifting with
         // rounding and adjusting the exponent.
         abs_result = (a_abs >> sign_bits_delta).cast();
-        let tmp = src_exp_bias.wrapping_sub(dst_exp_bias) << R::SIGNIFICAND_BITS;
-        abs_result = abs_result.wrapping_sub(tmp.cast());
+        // Cast before shifting to prevent overflow.
```
Amanieu, line 55: I'm having trouble understand how exactly this helps prevent overflows. Aren't you casting to a smaller size (u16) here instead of operating as a u32?

Amanieu: Ah, I missed that. LGTM.

Cargo.toml
```diff
@@ -49,6 +49,10 @@ c = ["cc"]
 # which use inline assembly and fall back to pure Rust versions (if avalible).
 no-asm = []
 
+# Workaround for codegen backends which haven't yet implemented `f16` and
+# `f128` support. Disabled any intrinsics which use those types.
+no-f16-f128 = []
```
RalfJung, line 54: Features should usually be additive and not take things away -- so, can't we have an "f16-f128" feature that *enables* these intrinsics, and have them off by default?

tgross35: I think that was just following the `no-asm` feature above, but they could probably both be flipped to be positive

bjorn3: This "adds" support for codegen backends that don't support f16/f128. If a `f16-f128` feature is used instead, liballoc would likely enable it when depending on compiler-builtins, which makes it impossible to disable the f16/f128 intrinsics using `--features` when invoking cargo, instead requiring the standard library sources to be patched.

RalfJung: ... what?^^ I am confused.

We are [currently](https://github.com/rust-lang/rust/pull/125016) (once that PR lands) setting this feature in `std` because of said codegen backends. But that means building `core` alone will still fail with these codegen backends.

> liballoc would likely enable it when depending on compiler-builtins

That sounds wrong to me. Instead, bootstrap should enable the feature when it deems them supported by the environment.

bjorn3: > We are https://github.com/rust-lang/rust/pull/125016 (once that PR lands) setting this feature in std because of said codegen backends.

The build system of these codegen backends should set it, not libstd. For example for no-asm, this is what cg_clif does when building the standard library:

https://github.com/rust-lang/rustc_codegen_cranelift/blob/6db27529ddbb6d94e0c990e527aa8341c3d85a79/build_system/build_sysroot.rs#L279

RalfJung: I can accept the answer "this is a hack, compiler_builtins is special and has a single consumer and so we can get away with negative features".

PR #594, opened 2024-04-15
Fix math libraries not being linked on some platforms

Amjad50:
This is a continuation/fix of 018616e. In that commit, we made it add the math functions to all platforms (except apple-targets and windows), and use `weak` linking, so that it can be used if the system doesn't have those functions.

Didn't notice `mod math` was behind another set of `cfg`, so removed it as well here.

There is an issue with clippy, so disabled it on the file as a whole

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -45,9 +45,6 @@ mod macros;
 pub mod float;
 pub mod int;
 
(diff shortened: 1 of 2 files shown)


Amanieu:
Rather than messing with clippy attributes, just disable it in CI for now.

Though this should be fixed properly by updating the `libm` crate.

Amjad50:
Updated libm

PR #595, opened 2024-04-16
Add CI testing for AArch64 Darwin

tgross35:
The Apple ARM silicon has been around for a while now and hopefully will become Rust Tier 1 at some point. Add it to CI since it is distinct enough from aarch64-linux and x86_86-darwin that there may be differences.


tgross35:
Hits the problem from #592 so just merge this into that PR

PR #598, opened 2024-04-29
Apply `weak` attributes to all intrinsics

Amjad50:
Removed the `weak-intrinsics` feature, so that all functions will have the `weak` linkage attribute.

But this caused issues for `memset` (maybe other mem functions are related), so added a special feature for those `mem-weak-intrinsics`.

`mem` functions won't have the `weak` attribute unless this feature is enabled.

This also fixed the bug in https://github.com/rust-lang/rust/issues/124042.

Before this commit, generated code will be
```rust
pub extern "C" fn <name>(...) -> ... {
        // code...
}
pub mod <name> {
    #[linkage = "weak"]
    #[no_mangle]
    pub extern "C" fn <name>(...) -> ... {
        super::<name>(...)
    }
}
```

The issue is that there is 2 `weak` linkage, the first one is not required. Along refactoring `weak` attributes, this was fixed.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -147,7 +147,6 @@ macro_rules! intrinsics {
         $($rest:tt)*
     ) => (
         #[cfg($name = "optimized-c")]
-        #[cfg_attr(all(not(windows), not(target_vendor = "apple")), linkage = "weak")]
         pub $(unsafe $($empty)? )? extern $abi fn $name( $($argname: $ty),* ) $(-> $ret)? {
             extern $abi {
                 fn $name($($argname: $ty),*) $(-> $ret)?;
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -25,11 +25,12 @@ macro_rules! public_test_dep {
 /// platforms need and elsewhere in this library it just looks like normal Rust
 /// code.
 ///
-/// When the weak-intrinsics feature is enabled, all intrinsics functions are
-/// marked with #[linkage = "weak"] so that they can be replaced by another
-/// implementation at link time. This is particularly useful for mixed Rust/C++
-/// binaries that want to use the C++ intrinsics, otherwise linking against the
-/// Rust stdlib will replace those from the compiler-rt library.
+/// All intrinsics functions are marked with #[linkage = "weak"] when
+/// `not(windows) and not(target_vendor = "apple")`.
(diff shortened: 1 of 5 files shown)


src/macros.rs
```diff
@@ -25,11 +25,14 @@ macro_rules! public_test_dep {
...
 ///
-/// When the weak-intrinsics feature is enabled, all intrinsics functions are
-/// marked with #[linkage = "weak"] so that they can be replaced by another
-/// implementation at link time. This is particularly useful for mixed Rust/C++
-/// binaries that want to use the C++ intrinsics, otherwise linking against the
-/// Rust stdlib will replace those from the compiler-rt library.
+/// All intrinsics functions are marked with #[linkage = "weak"] when
+/// `not(windows) and not(target_vendor = "apple")` except `mem` intrinsics, which
+/// will only have the `#[linkage = "weak"]` attribute when the `mem-weak-intrinsics`
+/// feature is enabled.
```
Amanieu, line 31: `mem-weak-intrinsics` is not needed since the root cause of the original issue was tracked down to https://github.com/Kingcom/armips/pull/242.

Amjad50: yup, removed it. Looks like I didn't commit the changes to `Cargo.toml` before

src/macros.rs
```diff
@@ -211,7 +147,7 @@ macro_rules! intrinsics {
         $($rest:tt)*
     ) => (
         #[cfg($name = "optimized-c")]
-        #[cfg_attr(feature = "weak-intrinsics", linkage = "weak")]
+        #[cfg_attr(all(not(windows), not(target_vendor = "apple")), linkage = "weak")]
```
Amanieu, line 150: This is not needed here: weak linkage is only needed for functions with `#[no_mangle]`.

Amanieu (commented):
This may also be a good opportunity to cleanup the `mod` wrappers around the `no_mangle` version of each intrinsic. Specifically:
- There is no need for these modules to be `pub`. Same with the function inside them. `#[no_mangle]` will ensure the symbol is exported.
- The whole module should be `#[cfg(not(feature = "mangled-names")]`. The functions themselves can then be `#[no_mangle]` unconditionally.

Amanieu:
Done

PR #603, opened 2024-05-06
Deny warnings in CI and add check-cfg directives

tgross35:
There are currently a lot of warnings printed in CI, a lot of which come from the new check config. Update CI to deny warnings.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -375,15 +375,6 @@ mod c {
             ]);
         }
 
-        // When compiling in rustbuild (the rust-lang/rust repo) this library
-        // also needs to satisfy intrinsics that jemalloc or C in general may
-        // need, so include a few more that aren't typically needed by
-        // LLVM/Rust.
-        #[allow(unexpected_cfgs)]
-        if cfg!(feature = "rustbuild") {
-            sources.extend(&[("__ffsdi2", "ffsdi2.c")]);
-        }
-
         // On iOS and 32-bit OSX these are all just empty intrinsics, no need to
         // include them.
         if target_os != "ios"
--- a/build.rs
+++ b/build.rs
@@ -2,6 +2,7 @@ use std::{collections::BTreeMap, env, sync::atomic::Ordering};
 
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
+    configure_check_cfg();
 
     let target = env::var("TARGET").unwrap();
     let cwd = env::current_dir().unwrap();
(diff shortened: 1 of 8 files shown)


build.rs
```diff
@@ -307,6 +378,7 @@ mod c {
         // also needs to satisfy intrinsics that jemalloc or C in general may
         // need, so include a few more that aren't typically needed by
         // LLVM/Rust.
+        #[allow(unexpected_cfgs)]
         if cfg!(feature = "rustbuild") {
             sources.extend(&[("__ffsdi2", "ffsdi2.c")]);
         }
```
Urgau, line 384: Could probably be removed. I don't think it's used anywhere in rust-lang/rust.
```suggestion
```

build.rs
```diff
@@ -145,6 +151,71 @@ fn generate_aarch64_outlined_atomics() {
...
+        println!(
+            "cargo::rustc-check-cfg=cfg({}, values(\"optimized-c\"))",
+            fn_name
+        );
+    }
+
+    // FIXME: this feature definitely exists in
+    // `rustc --print target-features --target sparc64-unknown-linux-gnu`. Why does check-cfg
+    // flag it?
+    println!("cargo::rustc-check-cfg=cfg(target_feature, values(\"vis3\"))");
```
Urgau, line 212: Probably because `rustc` (the front-end) doesn't know about any of the sparc target features. The full list of known target features is defined [here](https://github.com/rust-lang/rust/blob/d287f3e4eeaf680e8fe875f1ec75cca68f357d30/compiler/rustc_target/src/target_features.rs) and there isn't any entry for the Sparc architecture so any attempt to use them with the front-end will result in a warning.

It may still work since we pass them anyway to codegen but that's not guaranteed at all.

build.rs
```diff
@@ -1,14 +1,16 @@
...
+    configure_check_cfg();
 
     let target = env::var("TARGET").unwrap();
     let cwd = env::current_dir().unwrap();
 
     println!("cargo:compiler-rt={}", cwd.join("compiler-rt").display());
 
     // Activate libm's unstable features to make full use of Nightly.
+    println!("cargo::rustc-check-cfg=cfg(feature, values(\"unstable\"))");
     println!("cargo:rustc-cfg=feature=\"unstable\"");
```
Urgau, line 14: Should probably be replaced by a real custom config and not a faux Cargo feature, ie `unstable` and not `feature = "unstable"`.

build.rs
tgross35: I looked and couldn't find it being set in rust-lang/rust. Removed

build.rs
tgross35: Ah, makes sense. I just updated the note.

build.rs
tgross35: This comes from libm so I could probably do it here https://github.com/rust-lang/libm/pull/296

Thanks for the quick review

Issue #604, opened 2024-05-06
aarch64 windows build failure


theKidOfArcrania:
There is a bug when trying to build for windows arm64 target. See in [line 145](https://github.com/rust-lang/compiler-builtins/blob/a991664446c5d766b58e7f45a589831e8c3abf6e/build.rs#L145) concatenates a path of env `$OUT_DIR` to `"/outlined_atomics.rs"`. The problem arises in windows because `$OUT_DIR` may resolve to an unc path in windows, and those do not support having forward slashes. Using something like `Path.join(...)` should fix this bug.

Amanieu:
I don't have a setup for building arm64 windows. Could you submit a PR with the fix once you have tested it?

PR #609, opened 2024-05-13
Fix paths for Windows arm64 build

theKidOfArcrania:
Fixes issue #604 by using `Path::join` to denote a subpath (as opposed to using forward slash).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -83,6 +83,9 @@ jobs:
         - target: x86_64-pc-windows-gnu
           os: windows-latest
           rust: nightly-x86_64-gnu
+        - target: aarch64-pc-windows-gnu
+          os: windows-latest
+          rust: nightly
     steps:
     - name: Print runner information
       run: uname -a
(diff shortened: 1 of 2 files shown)


theKidOfArcrania:
Wanted to see if we can get an aarch64 windows build as part of the CI/CD, but it seems like this is something that is not supported properly? Or is there a way and I'm just not familiar with it?

Amanieu:
Github doesn't have any ARM64 Windows runners, so it's fine to just exclude this from CI.

theKidOfArcrania:
Alright, this bug *should* be fixed, I've tried building windows aarch64 by passing an unc path to `--target-dir` and it no longer crashes. Though I've taken a comb through the code, and we might have deeper issues if let's say the `RUST_COMPILER_RT_ROOT` is an unc path, since there are a lot of paths relative to that directory that contain forward slashes. It may also be harder to fix those issues also.

Normally this shouldn't be the case, but in case if this comes up, there may be larger fixes needed.

tgross35:
You could do a cross compile build or at least check on an x64 windows runner to verify the config without running tests

PR #612, opened 2024-05-15
Add tests for verbatim paths on windows builds

theKidOfArcrania:
This adds test building for windows targets if `--target-dir` gets set to a verbatim path. This is due to the fact that verbatim or verbatim unc paths have an additional requirement of having all path separators to be backslashes as opposed to be permissive of forward and back slashes.

This was made in response to #609 / #604.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ jobs:
         - target: aarch64-pc-windows-msvc
           os: windows-latest
           rust: nightly
-          test_unc: 1
+          test_verbatim: 1
           no_std: 1
         - target: arm-unknown-linux-gnueabi
           os: ubuntu-latest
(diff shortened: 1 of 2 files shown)


ChrisDenton:
Terminology note: rust calls these ["verbatim"](https://doc.rust-lang.org/std/path/enum.Prefix.html) paths. A UNC path is technically one in the form:

 `\\servername\sharename\`

A bit more about this here: https://users.rust-lang.org/t/understanding-windows-paths/58583

theKidOfArcrania:
> Terminology note: rust calls these ["verbatim"](https://doc.rust-lang.org/std/path/enum.Prefix.html) paths. A UNC path is technically one in the form:
> 
> `\\servername\sharename\`
> 
> A bit more about this here: https://users.rust-lang.org/t/understanding-windows-paths/58583

noted

theKidOfArcrania:
hmm... it looks the gnu toolchain does not support verbatim paths?

ci/run.sh
```diff
@@ -32,6 +32,19 @@ else
     $run --features no-f16-f128 --release
 fi
 
+if [ "${TEST_VERBATIM:-}" = "1" ]; then
+    verb_path=$(cmd.exe //C echo \\\\?\\%cd%\\testcrate\\target2)
+    run="cargo build --manifest-path testcrate/Cargo.toml --target $target --target-dir $verb_path"
+    $run
+    $run --release
+    $run --features c
```
Amanieu, line 40: Only `$run --features c` is really needed here. Let's avoid wasting too much CI time on an edge case.

.github/workflows/main.yml
```diff
@@ -78,15 +83,19 @@ jobs:
...
           rust: nightly
+          test_verbatim: 1
         - target: i686-pc-windows-gnu
           os: windows-latest
           rust: nightly-i686-gnu
+          test_verbatim: 1
         - target: x86_64-pc-windows-gnu
           os: windows-latest
           rust: nightly-x86_64-gnu
+          test_verbatim: 1
```
Amanieu, line 98: It's fine to not test this on -gnu targets, fundamentally it seems to be a limitation of the underlying toolchain.

theKidOfArcrania:
Ok I think all the changes have been made and should be good to merge?

PR #613, opened 2024-05-16
Add `f128` float to integer conversion functions

tgross35:
Add the following:

- `__fixtfsi`
- `__fixtfdi`
- `__fixtfti`
- `__fixunstfsi`
- `__fixunstfdi`
- `__fixunstfti`

### DIFF
--- a/testcrate/build.rs
+++ b/testcrate/build.rs
@@ -1,7 +1,15 @@
-use std::env;
+use std::{collections::HashSet, env};
+
+/// Features to enable
+#[derive(Debug, PartialEq, Eq, Hash)]
+enum Feature {
+    NoSysF128,
+    NoSysF128IntConvert,
+}
 
 fn main() {
     let target = env::var("TARGET").unwrap();
+    let mut features = HashSet::new();
 
     // These platforms do not have f128 symbols available in their system libraries, so
     // skip related tests.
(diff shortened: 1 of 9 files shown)


tgross35:
I think this one should be all set now

testcrate/build.rs
```diff
@@ -21,7 +29,24 @@ fn main() {
...
         || target.starts_with("powerpc64-")
     {
-        println!("cargo:warning=using apfloat fallback for f128");
-        println!("cargo:rustc-cfg=feature=\"no-sys-f128\"");
+        features.insert(Feature::NoSysF128);
+        features.insert(Feature::NoSysF128IntConvert);
+    }
+
+    if target.starts_with("i586") {
+        // i586 seems to not have ``__fixunstfti`, but does have everything else
```
Amanieu, line 37: It's strange that this is only missing for i586 but not i686 since they should both be linking to the same libgcc. Are the docker images using different package versions for these 2 targets?

tgross35: It just didn't show up because i686 was already disabled a few lines up due to an ABI issue https://github.com/rust-lang/compiler-builtins/pull/613/files#diff-05ae0211bd995d6870067db299c32944cf2b743226442889bb07d48232afc2c5R22-R24. Interesting, I guess that specific issue must not be a problem on i586.

I explicitly listed i586 here to be clear.

PR #614, opened 2024-05-16
Add `__powitf2` symbol for `f128` integer exponentiation

tgross35:


tgross35:
Blocked until we have the function in libs, https://github.com/rust-lang/rust/pull/125188

Qadeem411 (approved):

tgross35:
Lol @Qadeem411 I have to imagine you didn’t mean to approve this

tgross35:
WIndows targets appear to fail compiling this crate because powi uses the f128 multiply and divide in its implementation, and these are not available. I don't know why only Windows has the issue, but other platforms without these available natively (e.g. aarch darwin) do not, must be slightly different linkage rules. How should I handle this @Amanieu?

Amanieu:
I think this just happens to work before libgcc provides these symbols, but that's not available on -msvc targets. You'll need to implement a fallback for that target for testing, at least temporarily.

tgross35:
It looks like I can make this work by recreating `__multf3` and `__divtf3` in testcrate's `lib.rs`, this will just be blocked until I implement `divtf3`.

tgross35:
`__divtf3` hasn't made it to nightly yet, and I can't seem to figure out a way to provide it here to avoid the Windows linking. Hopefully that won't be too far off, but I just disabled this new symbol on MSVC in the meantime.

This should be ready for a look when you get the chance @Amanieu

Amanieu (approved):
LGTM. Feel free to merge it either now (and enable it with MSVC later) or after a boostrap bump.

tgross35:
Thanks! https://github.com/rust-lang/rust/pull/130792 hasn't been picked up yet so I might as well just update that to include this.

PR #615, opened 2024-05-18
Rework the test crate to separate individual tests

tgross35:
Currently, tests of the same kind are grouped together across all types into a single function. This makes it difficult to understand exactly what failed in CI.

Change test macros to create separate functions for separate types so failures are more fine grained. Most existing functions become modules, then their individual calls to `fuzz` are split to functions.

### DIFF
--- a/testcrate/tests/addsub.rs
+++ b/testcrate/tests/addsub.rs
@@ -4,139 +4,149 @@
 
 use testcrate::*;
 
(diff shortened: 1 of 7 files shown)


tgross35:
This conflicts with the other PRs I have open, but I can rebase in either direction depending on which lands first.

PR #623, opened 2024-05-27
Remove __multc3

XrXr:
This C file doesn't seem to be compiled correctly, as it creates a reference to the x86 specific `__builtin_copysignq` on ARM. This intrinsic is unused and unavailable on Windows because it's for complex numbers Rust doesn't support, so it should be fine to remove.

The inclusion of the file seems to be the cause of some downstream link errors: https://github.com/rust-lang/rust/issues/125619

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -533,10 +533,6 @@ mod c {
                 ("__fe_getround", "fp_mode.c"),
                 ("__fe_raise_inexact", "fp_mode.c"),
             ]);
-
-            if target_os != "windows" {
-                sources.extend(&[("__multc3", "multc3.c")]);
-            }
         }
 
         if target_arch == "mips" || target_arch == "riscv32" || target_arch == "riscv64" {


Amanieu:
This is actually a bug in upstream LLVM that should be fixed there:

```c
#if __has_builtin(__builtin_copysignf128)
#define crt_copysignf128(x, y) __builtin_copysignf128((x), (y))
#elif __has_builtin(__builtin_copysignq) || (defined(__GNUC__) && __GNUC__ >= 7)
#define crt_copysignf128(x, y) __builtin_copysignq((x), (y))
#endif
```

The condition is incorrect: on GCC 7+, we should be using either `__builtin_copysignq` or `__builtin_copysignf128` depending on the target. From the glibc source I find that x86, ia64 and powerpc use `__builtin_copysignq` while all other targets use `__builtin_copysignf128`.

GCC 6 and before simply don't properly support f128.

XrXr:
Thanks, I was hoping we could fix this just by removing it. I'm surprised that this one slipped through the cracks since any attempt to actually use `__multc3` on aarach64 should trigger the link error. I guess upstream LLVM doesn't test the `!__has_builtin(__builtin_copysignf128)` case, and rust releases build this crate without `__builtin_copysignf128`.

Amanieu:
The issue only happens when compiler-rt is built with GCC, which happens in our build.rs. This is why it was not caught earlier: compiler-rt is normally only built as part of Clang toolchains and clang will always return true for `__has_builtin(__builtin_copysignf128)`.

The best way to fix this would be to send a PR to llvm/llvm-project, and then backport the fix to rust's LLVM fork (rust-lang/llvm-project).

XrXr:
I'm having trouble recreating the bad object with GCC to make a patch for `compiler-rt`. I have GCC 13.2, and `__has_builtin(__builtin_copysignf128)` has been available  since GCC 7.1 ([1] [2]). When I make a crate that depends on `compiler-builtins` with the `c` feature, it links and runs fine. I'm able to single step into my local copy of `__multc3` and everything.
I suspect rust binary releases are built with an old GCC. If they are, a patch for `compiler-rt` might need to  include an implementation of `copysignf128` without using the builtin. But also, using a newer GCC might solve this issue without any code change.
I'll poke around to figure out the build environment of binary releases.

[1]: https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:2,lang:___c,selection:(endColumn:3,endLineNumber:2,positionColumn:3,positionLineNumber:2,selectionStartColumn:3,selectionStartLineNumber:2,startColumn:3,startLineNumber:2),source:'int++main(void)+%7B%0A++return+__builtin_copysignf128(0,0)%3B%0A%7D'),l:'5',n:'1',o:'C+source+%232',t:'0')),header:(),k:43.96151293961513,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:carm64g730,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:___c,libs:!(),options:'-O3+-Wall',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:2),l:'5',n:'0',o:'+ARM64+GCC+7.3.0+(Editor+%232)',t:'0')),header:(),k:56.038487060384874,l:'4',m:58.33333333333332,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+(trunk)',editorid:2,fontScale:14,fontUsePx:'0',j:2,wrap:'1'),l:'5',n:'0',o:'Output+of+ARM64+GCC+7.3.0+(Compiler+%232)',t:'0')),header:(),l:'4',m:41.66666666666666,n:'0',o:'',s:0,t:'0')),k:56.038487060384874,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4
[2]: https://gcc.gnu.org/gcc-7/changes.html

XrXr:
Closing, as I was able to piece enough together to send https://github.com/llvm/llvm-project/pull/93890. Hopefully I get a response. Thanks for the guidance!

PR #624, opened 2024-05-29
Add f128 int to float conversions

tgross35:


tgross35:
This should be ready now

Cc @m-ou-se since you wrote the original int to float algorithms. The generic function's monos seem to get reasonably similar codegen, running the benchmarks seems like changes are within a margin of error.

m-ou-se:
Haven't looked at it in detail, but was it necessary to make it generic? For f128, all integers up to 64 bit fit into it losslessly without rounding, so those conversions should be nearly trivial, right?

Anyway, happy to review this in detail next week.

r? m-ou-se

tgross35:
> Haven't looked at it in detail, but was it necessary to make it generic?

Not strictly, but nine near-identical functions would be a lot. I think the constants are easier to follow than the magic numbers in any case.

> Anyway, happy to review this in detail next week.

Thanks!

m-ou-se:
Having reviewed in detail now, I'm still not convinced we should make this function generic. It now has quite a few different cases to handle, adding complexity. And now it always starts with an "if 0" check, even though a few functions (e.g. `u64_to_f32_bits`) were branch free (after codegen) before.

tgross35:
Thanks for taking a look. I think I struck a happier medium - moved the redundant parts to generic functions to make things less magic-number-y, but did not combine the algorithms. Also included your suggested u32 -> f128 conversion.

tgross35:
It looks like the system NaN functions are converting to max values rather than zero for f128. Not sure why this would be but I'll disable them and look into it.

@m-ou-se gentle nudge, could you take another look at this?

src/float/conv.rs
```diff
@@ -9,70 +9,180 @@ use super::Float;
...
+/// The algorithm is explained here: <https://blog.m-ou.se/floats/>. It roughly does the following:
+/// - Calculate the exponent based on the base-2 logarithm of `i` (leading zeros)
+/// - Calculate a base mantissa by shifting the integer into mantissa position
+/// - Figure out if rounding needs to occour by classifying truncated bits. Some patterns apply
+///   here, so they may be "squashed" into smaller numbers to spmplifiy the classification.
 mod int_to_float {
+    use super::*;
+
+    /// Calculate the exponent from the number of leading zeros.
+    fn exp<I: Int, F: Float<Int: CastFrom<u32>>>(n: u32) -> F::Int {
```
m-ou-se, line 21: It doesn't calculate the exponent. It calculates the exponent minus one.

src/float/conv.rs
```diff
@@ -9,70 +9,180 @@ use super::Float;
...
+        F::Int::cast_from(I::BITS + F::EXPONENT_BIAS - 2 - n)
+    }
+
+    /// Shift the integer into the float's mantissa bits. Keep the lowest exponent bit intact.
+    fn m_base<I: Int, F: Float<Int: CastFrom<I>>>(i_m: I) -> F::Int {
```
m-ou-se, line 26: It keeps the *highest* bit in tact.

src/float/conv.rs
```diff
@@ -9,70 +9,180 @@ use super::Float;
...
+    /// Adjust a mantissa with dropped bits
+    fn m_adj<F: Float>(m_base: F::Int, dropped_bits: F::Int) -> F::Int {
+        // Branchlessly extract a `1` if rounding up should happen
+        let adj = (dropped_bits - (dropped_bits >> (F::BITS - 1) & !m_base)) >> (F::BITS - 1);
+
+        // Add one when we need to round up. Break ties to even.
+        m_base + adj
+    }
+
+    /// Combine a final float repr from an exponent and mantissa.
```
m-ou-se, line 55: It's an important detail that it doesn't just combine the exponent and mantissa fields, but instead is designed to take an off-by-one exponent and a mantissa with an extra 1 bit on the high side.

m-ou-se (commented):
I don't find it more readable when split up over multiple functions like this. For example, the `exp` function doesn't actually calculate the exponent; it calculates the exponent minus one, because of the trick with the mantissa containing an extra 1 bit that will unconditinally overflow into the exponent later. Separating tricks over multiple functions just makes it harder to follow and review.

That said, if you truly find this more maintainable, I'll be happy to approve this given that the code is well tested and benchmarked.

tgross35:
Ah this kept falling to the back of my todo list, but I finally got around to it. I changed it so none of the mantissa shifting is hidden in functions, only the values they shift by, so compared to the original it's just easier to understand where the numbers come from. I left the obvious functions `exp` and `repr`, also `m_adj` because that is a tricky pattern used the same way everywhere.

Take a look when you get the chance @m-ou-se, feel free to push on top of this if you feel it's needed.

tgross35:
@m-ou-se could you take another look at this? Last puzzle piece to making `f128` testable on all platforms :)

m-ou-se (approved):
Thanks, this looks good to me now. Feel free to merge. :)

examples/intrinsics.rs
```diff
@@ -264,14 +264,18 @@ mod intrinsics {
...
+
     // floatsidf
     pub fn aeabi_i2d(x: i32) -> f64 {
         x as f64
     }
 
-    // floatsisf
-    pub fn aeabi_i2f(x: i32) -> f32 {
-        x as f32
+    pub fn floatsitf(x: i32) -> f128 {
```
bjorn3, line 277: These need `#[cfg(f128_enable)]`, right?

tgross35:
Yay! Many thanks for the reviews :)

examples/intrinsics.rs
tgross35: Do tests in this repo ever get run with cranelift? This "example" is what tests linking when the current crate is used as `compiler_builtins` rather than the sysroot.

PR #625, opened 2024-05-29
Update the Ubuntu docker image to the latest version

tgross35:


tgross35:
I think that some of the problems I am having reproducing bugs locally comes from more recent fixes to qemu, possibly also to gcc.

Is there a reason for using an older version? I'll just drop this if so, otherwise will try to chase down the ppc bug.

Amanieu:
No, there's no reason to use an older version.

tgross35:
I had to disable testing against system f128 functions on powerpc64le since addkf3 and mulkf3 were hitting a SIGILL. I assume that the old version of system libraries may have been using softfloat but the newer versions use f128 arithmetic instructions, which Qemu seems to not support very well.

PR #626, opened 2024-05-29
Add generic unorddf2vfp() and unordsf2vfp()

sbechet:
A (weak) contribution to trying to understand this great project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,8 +150,8 @@ rely on CI.
 - [ ] arm/udivmodsi4.S (generic version is done)
 - [ ] arm/udivsi3.S (generic version is done)
 - [ ] arm/umodsi3.S (generic version is done)
-- [ ] arm/unorddf2vfp.S
-- [ ] arm/unordsf2vfp.S
+- [x] arm/unorddf2vfp.S (generic version is done)
+- [x] arm/unordsf2vfp.S (generic version is done)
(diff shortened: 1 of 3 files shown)


Amanieu:
Since these only exist on ARM, please mark them with `#[cfg(target_arch = "arm")]`.

sbechet:
> Since these only exist on ARM, please mark them with `#[cfg(target_arch = "arm")]`.

You right. I will correct that. Before I submit an update are we ok about check boxes. I ask the question because I saw that certain boxes were not checked for certain functions even though the generic version of the code exists.

Amanieu:
I double-checked the *vfp functions and it turns out that they are only used on old versions of 32-bit iOS with ARMv5. Since we don't support this in any current target, we can probably just remove all the *vfp functions.

sbechet:
That's perfect, in this case, I can close this pull request without further delay.
Are you interested in a pull request to update the README and deprecate all these functions or will you do it? There must not be much left to do in this project I imagine?

Amanieu:
Yes, I would be interested in a PR to update the README and remove all the existing *vfp functions since they serve no purpose.

tgross35:
@sbechet since we probably don't need this change as-is I am just going to close this PR. Feel free to create a new one that removes the symbols as mentioned above.

PR #627, opened 2024-05-30
Build with `-Werror=implicit-function-declaration`

XrXr:
To fail-fast in situations like
https://github.com/rust-lang/rust/issues/125619, where an upstream
source compiles but creates a link error way downstream.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -327,6 +327,8 @@ mod c {
             // in https://github.com/rust-lang/compiler-rt/blob/c8fbcb3/cmake/config-ix.cmake#L19.
             cfg.flag_if_supported("-fomit-frame-pointer");
             cfg.define("VISIBILITY_HIDDEN", None);
+            // Avoid implicitly creating references to undefined functions
+            cfg.flag("-Werror=implicit-function-declaration");
         }
 
         // int_util.c tries to include stdlib.h if `_WIN32` is defined,


XrXr:
This is revealing the issue in https://github.com/rust-lang/rust/issues/125619 indeed. Worth considering once that's solved.

tgross35:
> This is revealing the issue in [rust-lang/rust#125619](https://github.com/rust-lang/rust/issues/125619) indeed (aarch64-unknown-linux-gnu is failing). Worth considering once that's solved.

Does that just need https://github.com/llvm/llvm-project/pull/93890 to work its way into our LLVM, or do we need to make changes to what `.c` files we include here?

XrXr:
It should be good once our LLVM updates.

tgross35:
Mind rebasing so we get an up to date CI run? I'd like to see what the errors are.

https://github.com/llvm/llvm-project/pull/93890 probably could have been requested for a cherry pick but it's likely too late now :/

tgross35:
Oh I guess that did make it to our LLVM somehow, interesting.

One thing I am not sure about is whether we want this check all the time or only for testing. Are there any cases where this could cause a build failure where the user doesn't expect it? I wonder if it would be better to add something like a `COMPILER_BUILTINS_STRICT_C` flag that we set in CI to enable this and (later) possibly some other warnings/werror.

@Amanieu I'll defer to you for what would be best here.

XrXr:
It looks like CI runs with LLVM 18.0-2024-02-13, which doesn't include https://github.com/llvm/llvm-project/pull/93890, but with 3032f496c56a524a7de756ec7af1a66186e5d7df, we're now using a new enough GCC to dodge the build error. (By the way, "updating the compiler" was also the [route](https://github.com/rust-lang-ci/rust/commit/65d1a73aabc03c7f94633bf00eee1b5548c28c14) rustc took to fix the build issue for the time being.)

tgross35:
We can upgrade LLVM so I put up a PR to do that https://github.com/rust-lang/compiler-builtins/pull/703.

Amanieu:
I think we should probably enable this, at the very least it will notify users that their GCC and/or LLVM needs to be updated to avoid the issue.

Amanieu (approved):

tgross35:
Released with the update https://github.com/rust-lang/compiler-builtins/pull/700. Could you make a PR to rust-lang/rust updating to 0.1.131?

PR #631, opened 2024-06-22
Enable `f128 -> f16` tests on Linux

tgross35:
Since updating the docker images in <https://github.com/rust-lang/compiler-builtins/pull/625>, it looks like `__extendhftf2` and `__trunctfhf2` are available on all 64-bit Linux platforms.

### DIFF
--- a/testcrate/build.rs
+++ b/testcrate/build.rs
@@ -38,12 +38,9 @@ fn main() {
     }
 
     if target.starts_with("i586") || target.starts_with("i686") {
-        // 32-bit x86 seems to not have `__fixunstfti`, but does have everything else
+        // 32-bit x86 does not have `__fixunstfti`/`__fixtfti` but does have everything else
         features.insert(Feature::NoSysF128IntConvert);
-    }
-
-    if target.contains("-unknown-linux-") {
-        // No `__extendhftf2` on x86, no `__trunctfhf2` on aarch64
+        // FIXME: 32-bit x86 has a bug in `f128 -> f16` system libraries
         features.insert(Feature::NoSysF16F128Convert);
     }
 


testcrate/build.rs
```diff
@@ -37,13 +37,10 @@ fn main() {
         features.insert(Feature::NoSysF16F128Convert);
     }
 
-    if target.starts_with("i586") || target.starts_with("i686") {
+    if target.starts_with("i586") {
         // 32-bit x86 seems to not have `__fixunstfti`, but does have everything else
```
Amanieu, line 41: I think this is fixed for i586 as well now?

tgross35: I just tried it, looks like they are still not available on i586. I updated the note to mention this

Amanieu: I still don't understand why this function is available on i686 but not i586: they should be using the exact same docker image and toolchain.

tgross35: Ok, think I figured it out - i586 and i686 are both missing `__fixunstfti` and `__fixtfti`. i686 was just always disabled at https://github.com/rust-lang/compiler-builtins/pull/631/files#diff-05ae0211bd995d6870067db299c32944cf2b743226442889bb07d48232afc2c5L26 because it has a f128 math bug, https://github.com/rust-lang/compiler-builtins/actions/runs/9640552090/job/26584543080?pr=631#step:13:244.

i586 probably also has that bug, but float tests without sse2 have been disabled since at least https://github.com/rust-lang/compiler-builtins/pull/384 https://github.com/rust-lang/compiler-builtins/blob/0ccc1bfb962883e04bcce1bdc6a6a3a6ac2d2bbc/testcrate/tests/addsub.rs#L113-L132

PR #632, opened 2024-06-23
Fix building on AVR

beetrees:
As `__adddf3` and `__addsf3` are marked with `#[avr_skip]` they don't exist on AVR. This PR ensures any usage of them is within `#[avr_skip]` intrinsics.


tgross35:
Could we maybe add a CI check for AVR since it has so much special handling?

PR #634, opened 2024-06-25
Get rid of a warning

tea:
I get this error when building the crate using fresh rust:

```
error: `extern crate` is not idiomatic in the new edition
   --> /home/aturkin/redox/compiler-builtins/build.rs:225:5
    |
225 |     extern crate cc;
    |     ^^^^^^^^^^^^^^^^
    |
    = note: `-D unused-extern-crates` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unused_extern_crates)]`
help: convert it to a `use`
    |
225 |     use cc;
    |     ~~~
```

Indeed 2018 edition documentation says `extern crate` should not be used anymore "in 99% of circumstances". So this PR follows the advice.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ readme = "README.md"
 repository = "https://github.com/rust-lang/compiler-builtins"
 homepage = "https://github.com/rust-lang/compiler-builtins"
 documentation = "https://docs.rs/compiler_builtins"
-edition = "2018"
+edition = "2021"
 description = """
 Compiler intrinsics used by the Rust compiler. Also available for other targets
 if necessary!
--- a/build.rs
+++ b/build.rs
@@ -219,8 +219,6 @@ fn configure_check_cfg() {
 
 #[cfg(feature = "c")]
 mod c {
-    extern crate cc;
-
     use std::collections::{BTreeMap, HashSet};
     use std::env;
     use std::fs::{self, File};
--- a/crates/panic-handler/Cargo.toml
+++ b/crates/panic-handler/Cargo.toml
@@ -2,5 +2,6 @@
 name = "panic-handler"
 version = "0.1.0"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
+edition = "2021"
 
 [dependencies]


Amanieu:
The `use` isn't even necessary: crate names are always automatically imported.

tea:
> The `use` isn't even necessary: crate names are always automatically imported.

Removed now. CI is failing due to some seemingly unrelated stuff.

tgross35:
Could you try setting `edition = 2021` in crates/panic-handler/Cargo.toml?

tgross35:
You can probably change to `edition = 2021` in the root Cargo.toml too, I don't think we have any reason to stay on 2018

tea:
> Could you try setting `edition = 2021` in crates/panic-handler/Cargo.toml? That is an interesting failure

No dice.

tgross35:
Oh nevermind, this is just a new lint in the latest nightly https://github.com/rust-lang/rust/commit/c4c7859e40efcfff640af442fb5d1fab3718d374. Looks like lines 135 and 263 try to use `define_rust_probestack` but the macro `cfg`ed out on the relevant platforms.

To match the current behavior you could add another `define_rust_probestack` macro that returns nothing and has `cfg` to only be available on the platforms that are missing it (really a better job for cfg_if but we don't have that).

Even if that matches the current behavior though, it might not be the intended behavior (Amanieu will know)

tea:
> Oh nevermind, this is just a new lint in the latest nightly [rust-lang/rust@c4c7859](https://github.com/rust-lang/rust/commit/c4c7859e40efcfff640af442fb5d1fab3718d374). Looks like lines 135 and 263 try to use `define_rust_probestack` but the macro `cfg`ed out on the relevant platforms.

That is weird. The macro is defined within the same file; entire file is only cfged in for x86/x64, and conditionals on the 4 macro defines seem to cover every possible target.

Anyway, this is a bit above my head (I'm just learning Rust so decided to do a light deep immersion by trying to add risc-v support to Redox; just doing something easy, y'know :)).

tgross35:
> That is weird. The macro is defined within the same file; entire file is only cfged in for x86/x64, and conditionals on the 4 macro defines seem to cover every possible target.

It is weird, the new lint might just be wrong. I asked at https://github.com/rust-lang/rust/issues/126984.

> Anyway, this is a bit above my head (I'm just learning Rust so decided to do a light deep immersion by trying to add risc-v support to Redox; just doing something easy, y'know :)).

You should probably be able to just do what it suggests and add `#![allow(out_of_scope_macro_calls)]` to the top of the file so the warning doesn't show up, we can figure out the cause later.

Quite the light deep immersion indeed, welcome to Rust :)

PR #637, opened 2024-06-30
Fix incorrect rounding with subnormal/zero results of float multiplication

### LINKED ISSUE #616: Rounding error in `__mulsf3`
This is a problem in Rust's compiler-builtins, but also the LLVM library (which is demonstrated here). The correct answer should be 0.0.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=54ee2169049eb22eb1b44d41c6f37e1e

```rust
extern "C" {
    fn __mulsf3(a: f32, b: f32) -> f32;
}

fn main() {
    let a = 7.83651e-39f32;
    let b = 2.7310085e-8f32;
    let res_asm = a * b ;
    let res_mulsf3 = unsafe { __mulsf3(a, b) };
    println!("asm {} {:#010x}", res_asm, res_asm.to_bits());
    println!("mulsf3 {} {:#010x}", res_mulsf3, res_mulsf3.to_bits());
}
```

```
asm 0 0x00000000
mulsf3 0.000000000000000000000000000000000000000000001 0x00000001
```

beetrees:
Fix float multiplication that results in subnormals (or zero), and enable tests for those cases. Also remove dead code left over from when this was ported from `compiler-rt` (the other branches of the `if` statement are unreachable as `shift` is always less than `bits` at this point). The [equivalent code](https://github.com/llvm/llvm-project/blob/038bc1c18c786e14cc306401b00144265f8860f5/compiler-rt/lib/builtins/fp_lib.h#L250) in `compiler-rt` does not have this bug.

Fixes #616

### DIFF
--- a/testcrate/src/bench.rs
+++ b/testcrate/src/bench.rs
@@ -17,9 +17,10 @@ pub fn skip_sys_checks(test_name: &str) -> bool {
         "extend_f16_f32",
         "trunc_f32_f16",
         "trunc_f64_f16",
-        // FIXME(f16_f128): rounding error
+        // FIXME(#616): re-enable once fix is in nightly
         // <https://github.com/rust-lang/compiler-builtins/issues/616>
-        "mul_f128",
+        "mul_f32",
+        "mul_f64",
     ];
 
     // FIXME(f16_f128): error on LE ppc64. There are more tests that are cfg-ed out completely
@@ -29,7 +30,13 @@ pub fn skip_sys_checks(test_name: &str) -> bool {
 
     // FIXME(f16_f128): system symbols have incorrect results
     // <https://github.com/rust-lang/compiler-builtins/issues/617#issuecomment-2125914639>
-    const X86_NO_SSE_SKIPPED: &[&str] = &["add_f128", "sub_f128", "powi_f32", "powi_f64"];
+    const X86_NO_SSE_SKIPPED: &[&str] =
+        &["add_f128", "sub_f128", "mul_f128", "powi_f32", "powi_f64"];
+
+    // FIXME(f16_f128): Wide multiply carry bug in `compiler-rt`, re-enable when nightly no longer
+    // uses `compiler-rt` version.
+    // <https://github.com/llvm/llvm-project/issues/91840>
+    const AARCH64_SKIPPED: &[&str] = &["mul_f128"];
 
     // FIXME(llvm): system symbols have incorrect results on Windows
     // <https://github.com/rust-lang/compiler-builtins/issues/617#issuecomment-2121359807>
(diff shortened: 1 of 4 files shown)


tgross35:
There has to be a bug in either GCC or LLVM's libraries at least on some platforms for f128, I was able to reproduce in C https://github.com/llvm/llvm-project/issues/91840.

The clippy failure is from https://github.com/rust-lang/compiler-builtins/pull/634#issuecomment-2191071129

tgross35:
Probably just `#![allow(out_of_scope_macro_calls)]` to fix the rest until https://github.com/rust-lang/rust/pull/126987 gets merged

testcrate/src/bench.rs
```diff
@@ -17,9 +17,10 @@ pub fn skip_sys_checks(test_name: &str) -> bool {
         "extend_f16_f32",
         "trunc_f32_f16",
         "trunc_f64_f16",
-        // FIXME(f16_f128): rounding error
+        // FIXME(#616): re-enable once fix is in nightly
         // <https://github.com/rust-lang/compiler-builtins/issues/616>
-        "mul_f128",
+        "mul_f32",
+        "mul_f64",
```
tgross35, line 23: Why will these be fixed with the nightly bump? I guess if `__mul[sd]f3` don't exist since they are expected to lower to assembly, then linking the `extern "C"` symbols is linking to the weak ones from this library?

testcrate/src/bench.rs
```diff
@@ -29,7 +30,12 @@ pub fn skip_sys_checks(test_name: &str) -> bool {
 
     // FIXME(f16_f128): system symbols have incorrect results
     // <https://github.com/rust-lang/compiler-builtins/issues/617#issuecomment-2125914639>
-    const X86_NO_SSE_SKIPPED: &[&str] = &["add_f128", "sub_f128", "powi_f32", "powi_f64"];
+    const X86_NO_SSE_SKIPPED: &[&str] =
+        &["add_f128", "sub_f128", "mul_f128", "powi_f32", "powi_f64"];
+
+    // FIXME(f16_f128): Incorrect rounding.
+    // <https://github.com/llvm/llvm-project/issues/91840>
+    const AARCH64_SKIPPED: &[&str] = &["mul_f128"];
```
tgross35, line 38: It seems like this must have been reproduced in CI then, which is good. Any idea where the problem is coming from? https://github.com/llvm/llvm-project/blob/8598bcb9934dca16ea16d87304e00defc85d986c/compiler-rt/lib/builtins/fp_lib.h and https://github.com/llvm/llvm-project/blob/8598bcb9934dca16ea16d87304e00defc85d986c/compiler-rt/lib/builtins/fp_mul_impl.inc don't seem to have any architecture-specific behavior except for `fmaxf`.

tgross35: Ah great, guess I just didn't test this on other platforms. Thanks for testing

PR #638, opened 2024-07-01
Remove unnecessary benchmark files

tgross35:
These were just for my comparisons, didn't mean to actually commit them.


tgross35:
Error is from https://github.com/rust-lang/rust/issues/126984

PR #639, opened 2024-07-06
Implement remaining __clz*i2 intrinsics

tea:
This one was a bit interesting to figure out.

Firstly, the existing implementation was wrong, as it was working with whatever native word is, whereas `__clzsi2` should be 32-bit specifically (SI is 32-bit exactly). compiler-rt implementation is always 32-bit. Interestingly, existing CI tests didn't catch this because testcrate only tests against compiler-rt when compiled with `--no-default-features -F c` (because testcrate defaults `mangled-names`). When tested in this way, testcreate surely enough failed `__clzsi2` test on 64 bit target.

Even more interestingly, the compiler-rt implementation is also wrong! It is 32-bit implementation but it uses int type for its argument, so on 16-bit platform it couldn't get 32-bit value. I guess compiler-rt is not used much for 16 bit?

Now, libgcc implementation is the correct one. It always uses 16 bit types for HI functions, 32 bit types for SI, 64 bit for DI and 128 bit for TI. But it has its own twist: it only implements two functions out of all possible - one for native word size and one for double that. So on 16 bit it would be HI and SI, on 32 bit SI/DI and on 64 bit DI/TI. Which is why buggy existing implementation didn't caused any dramas: it would never be called as the compilers know it shouldn't even exist (indeed, even trying hard to call it via `__builtin_clz(uint32_t)` results in something akin to `_clzdi2(x as u64) - 32`).

So I did the same thing as with bswap - simply relegate work to the LLVM and let it spit out proper builtin implementation. i686 and x86-64 versions look great using `bsr`. Riscv64... cannot say if it became better or worse than it was; its a different algorithm, though both are branchless and logarithmic in complexity. One would need to benchmark it like crazy to decide which is better :)

I like new version better because of two things:
1) it could get better "for free" as llvm improves (definitely more eyes on llvm builtins than on compiler-builtins repo).
2) it could do much better if allowed (e.g. when there is a native cpu instruction to do it; Riscv64gc+Zbb functions const of two instructions).

With this implementation tests are not quite as useful; they can only work as a documentation of sorts since they check the same implementation against itself. Well, they also can check against compiler-rt so there's that.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -164,8 +164,6 @@ fn configure_check_cfg() {
         "__bswapsi2",
         "__bswapdi2",
         "__bswapti2",
-        "__clzsi2",
-        "__clzdi2",
         "__divdi3",
         "__divsi3",
         "__divmoddi4",
@@ -346,8 +344,6 @@ mod c {
             ("__absvsi2", "absvsi2.c"),
             ("__addvdi3", "addvdi3.c"),
             ("__addvsi3", "addvsi3.c"),
-            ("__clzdi2", "clzdi2.c"),
-            ("__clzsi2", "clzsi2.c"),
             ("__cmpdi2", "cmpdi2.c"),
             ("__ctzdi2", "ctzdi2.c"),
             ("__ctzsi2", "ctzsi2.c"),
@@ -435,8 +431,6 @@ mod c {
                 ("__aeabi_frsub", "arm/aeabi_frsub.c"),
                 ("__bswapdi2", "arm/bswapdi2.S"),
                 ("__bswapsi2", "arm/bswapsi2.S"),
-                ("__clzdi2", "arm/clzdi2.S"),
-                ("__clzsi2", "arm/clzsi2.S"),
                 ("__divmodsi4", "arm/divmodsi4.S"),
                 ("__divsi3", "arm/divsi3.S"),
                 ("__modsi3", "arm/modsi3.S"),
@@ -572,9 +566,6 @@ mod c {
                 }
             }
             sources.remove(&to_remove);
-
-            // But use some generic implementations where possible
-            sources.extend(&[("__clzdi2", "clzdi2.c"), ("__clzsi2", "clzsi2.c")])
         }
 
         if llvm_target[0] == "thumbv7m" || llvm_target[0] == "thumbv7em" {
--- a/build.rs
+++ b/build.rs
@@ -165,6 +165,7 @@ fn configure_check_cfg() {
         "__bswapdi2",
         "__bswapti2",
         "__clzsi2",
+        "__clzdi2",
         "__divdi3",
         "__divsi3",
         "__divmoddi4",
@@ -382,7 +383,6 @@ mod c {
             sources.extend(&[
                 ("__absvti2", "absvti2.c"),
                 ("__addvti3", "addvti3.c"),
-                ("__clzti2", "clzti2.c"),
                 ("__cmpti2", "cmpti2.c"),
                 ("__ctzti2", "ctzti2.c"),
                 ("__ffsti2", "ffsti2.c"),
--- a/src/int/leading_zeros.rs
+++ b/src/int/leading_zeros.rs
@@ -3,10 +3,12 @@
 // adding a zero check at the beginning, but `__clzsi2` has a precondition that `x != 0`.
 // Compilers will insert the check for zero in cases where it is needed.
 
+use crate::int::{CastInto, Int};
+
 public_test_dep! {
 /// Returns the number of leading binary zeros in `x`.
 #[allow(dead_code)]
-pub(crate) fn usize_leading_zeros_default(x: usize) -> usize {
+pub(crate) fn leading_zeros_default<T: Int + CastInto<usize>>(x: T) -> usize {
     // The basic idea is to test if the higher bits of `x` are zero and bisect the number
     // of leading zeros. It is possible for all branches of the bisection to use the same
     // code path by conditionally shifting the higher parts down to let the next bisection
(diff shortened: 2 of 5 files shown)


Amanieu:
Unfortunately that won't work here because LLVM *does* in fact use this builtin, for example on the `armv4t-unknown-linux-gnueabi` target.

> it could do much better if allowed (e.g. when there is a native cpu instruction to do it; Riscv64gc+Zbb functions const of two instructions).

In that case the implementation in compiler-builtins won't be called anyways since LLVM/GCC will just use the Zbb instruction directly.

tgross35:
Possible generic version:

```rust
fn leading_zeros<I: Int>(mut x: I) -> usize {
    let mut z = I::BITS;
    let mut t: I;
    let mut shift = I::BITS / 2;

    while shift > 1 {
        t = x >> shift;
        if t != I::ZERO {
            z -= shift;
            x = t;
        }

        shift /= 2;
    }

    t = x >> 1;
    let res = if t != I::ZERO { z - 2 } else { z - x.cast() };
    res as usize
}
```

LLVM correctly unrolls the loop to get the same result as the existing usize-only version. Which means it doesn't make use of e.g. `bsr` on x86, but I think we'd have to handwrite those if we want them emitted reliably (and not much purpose since LLVM knows that lowering on relevant platforms).

testcrate/tests/misc.rs
```diff
@@ -65,31 +65,46 @@ fn fuzz_values() {
...
+            if x == 0 {
+                return; // undefined value for an intrinsic
+            }
+            let lz = x.leading_zeros() as usize;
+            let lz0 = __clzti2(x);
+            if lz0 != lz {
+                panic!("__clzti2({}): std: {}, builtins: {}", x, lz, lz0);
+            }
+        });
+    }
```
tgross35, line 107: What exactly is happening on 16 & 32-bit platforms for the gating - is LLVM lowering `.leading_zeros()` to a symbol that is missing? From CI it looks like MSVC is missing `___clzti2` on arm and 32-bit so you'll have to update the gates.

If this logic is a bit more complex than just pointer sizes, it might be better to add a `NoSysU128Clz` option to https://github.com/rust-lang/compiler-builtins/blob/06db2de1c098f9c3cf69a52e7a273f19f0dce061/testcrate/build.rs#L5-L10, update that file, and then gate on that (easier if the gates need to be reused).

tea: I gated these, and the intrinsics themselves, the same way the compilers do (well, libgcc does so they have to follow along?). Makes no sense to test for intrinsics which aren't ever used.

As for 128 bit on 64-bit msvc - I guess they don't have 128 bit int somehow, according to compiler-rt. I just took "optimized c" attribute off the implementation instead of trying to gate it. Single branch isn't that bad.

tgross35: Ah, I missed that the gates were on the intrinsics as well. Could you please remove them? The reasons that `libgcc` doesn't supply a specific intrinsic on specific platforms may vary (no integers of that type, GCC itself doesn't use them, platform bugs, untested, just not done yet etc). But if we don't have any specific reason not to make them available, we should just provide them. And if there is a specific reason not to provide them (bugs), it needs to be commented.

Also a lot of the purpose of this library is to fill in for the shortcomings of libgcc, so it doesn't make any sense to copy their limitations :).

> well, libgcc does so they have to follow along?

Unfortunately this isn't very accurate, LLVM (and GCC) both pretty frequently emit symbols that aren't available on lesser-known platforms. Rust hits this more often because we make less common types available on all platforms (e.g. `u128`, `f16`, `f128`) where there might not be a C equivalent available.

Not like this specific symbol is likely to be a problem anywhere. But we should avoid unnecessary `cfg` because it just means somebody might have to chase them down in the future and wonder why it's disabled.

src/int/leading_zeros.rs
```diff
@@ -16,46 +18,47 @@ pub(crate) fn usize_leading_zeros_default(x: usize) -> usize {
...
     // the number of potential leading zeros
-    let mut z = usize::MAX.count_ones() as usize;
+    let mut z = T::BITS as usize;
     // a temporary
-    let mut t: usize;
-    #[cfg(target_pointer_width = "64")]
-    {
+    let mut t: T;
+
+    assert!(T::BITS <= 64);
```
tgross35, line 25:
```suggestion
    const { assert!(T::BITS <= 64) };
```

Might as well turn this into a comptime error (same for the other `assert!` places

tea: Indeed, that was my intention. Just somehow failed to do that...

testcrate/tests/misc.rs
tea: Removed all the gate. Let the flood begin!

As for libgcc and their builtins, the reason there is simple: that's how they do it. Basically they have two and only two functions __clzSI2 and __clzDI2 (note uppercase letters). Despite their names, they operate on native word size and double that respectively. These function names (and argument types) get macroed into whatever proper name and argument type is for the platform.

Same is true for many other builtins there.

tea: ping?

src/int/leading_zeros.rs
```diff
@@ -16,46 +18,47 @@ pub(crate) fn usize_leading_zeros_default(x: usize) -> usize {
...
-    let mut z = usize::MAX.count_ones() as usize;
+    let mut z = T::BITS as usize;
     // a temporary
-    let mut t: usize;
-    #[cfg(target_pointer_width = "64")]
-    {
+    let mut t: T;
+
+    const { assert!(T::BITS <= 64) };
+    if T::BITS >= 64 {
```
tgross35, line 26: Optional nit: the second condition could become `== 64` since `> 64` is rejected right above

build.rs
```diff
@@ -382,7 +383,6 @@ mod c {
             sources.extend(&[
                 ("__absvti2", "absvti2.c"),
                 ("__addvti3", "addvti3.c"),
-                ("__clzti2", "clzti2.c"),
```
tgross35, line 385: I think you could also remove `__clzdi2` and `__clzsi2` on lines 349-350 now that these aren't broken. Possibly https://github.com/rust-lang/compiler-builtins/blob/2bf0425e3f4853935b1ea1d2865edf5d3bca3074/build.rs#L577 too if our implementation works there.

testcrate/tests/misc.rs
tgross35: Thanks for the change. I think you might be able to remove some C sources that we are building (mentioned below) but otherwise the implementation looks good to me now.

build.rs
tea: I wasn't sure what's the policy on these. Is it "only apply C replacements for broken/missing functions"?

src/int/leading_zeros.rs
tea: I'd rather leave this as is, as it is more future-proof (in case any 128 bit implementations would actually pop up, and whoever applies 128 bit part happens to be both incredibly careless and doesn't test anything...)

More seriously though, result of the expression is a compile-time constant so it doesn't matter at all which operator to use here, and >= is more consistent with 32 bit part down below.

build.rs
tgross35: That is accurate, part of the goal of this crate is to be able to target some platforms without needing a C toolchain. I think that usually once something gets ported over and tested, it can be removed from the C sources lists.

tea: Done.

Aarch64 CI is down because of new rustc I guess (0/1 asm labels are now off limits instead of just discouraged). This particular thing about numeric labels amazes me so much... This is due to llvm bug, but instead of dealing with that people decided to document it in rust book (so I guess it is a feature now) and make it a compile-time error.

tgross35: I don't think it's an easy llvm bug to fix. That lint should almost certainly be x86-only though where LLVM was already throwing an error, I brought it up.

You can add

```
#![allow(binary_asm_labels)] // aarch64 has no problems with binary labels
```

to the top of `src/aarch64_linux.rs` just to get that to pass.

tgross35 (approved):
LGTM now, thanks. @Amanieu still needs to do the final review. The CI error obviously isn't your problem.

tgross35:
Fix coming for the lint btw, probably will be on tomorrow's nightly. Just rebase once it's available https://github.com/rust-lang/rust/pull/127935

PR #642, opened 2024-07-18
[experiment] android testing, make sync symbols strong

tgross35:
Try working around the android bug by making sync symbols strong.

Based on guesses from https://github.com/rust-lang/compiler-builtins/issues/641#issuecomment-2221798895

PR #644, opened 2024-07-18
Sync readme intrinsics with LLVM

tgross35:
I updated our with what is now available in compiler-rt. Almost all of these seem unneeded except maybe the riscv and avr assembly implementations.

Also reorganize the float section by type.

These updates come from running the following from within the LLVM compiler-rt directory:

```sh
fd '.*\.[cS]' | sort | xargs -iARG sh -c 'rg -q "\\bARG\\b" ../../../README.md || echo ARG'
```


tgross35:
Everything is failing due to the asm lint, which I disabled as part of https://github.com/rust-lang/compiler-builtins/pull/642

PR #646, opened 2024-07-27
Make float division generic

tgross35:
Currently there are two separate functions for for `f32` and `f64`. Combine them and extract some repeated functionality to a trait, which will make adding `f128` easier.

### DIFF
--- a/src/float/div.rs
+++ b/src/float/div.rs
@@ -1,462 +1,94 @@
-// The functions are complex with many branches, and explicit
-// `return`s makes it clear where function exit points are
-#![allow(clippy::needless_return)]
-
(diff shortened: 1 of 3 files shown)


tgross35:
Split this off of https://github.com/rust-lang/compiler-builtins/pull/622 so it should be easier to figure out the `f128`-specific issues in isolation.

tgross35:
Benchmarked just to make sure this didn't regress anything, changes are within noise tolerance for both f32 and f64.

tgross35:
https://github.com/rust-lang/compiler-builtins/pull/622 is actually getting closer so I don't think there is any need to have this on its own anymore.

PR #652, opened 2024-08-02
 Configure which platforms have `f16` and `f128` enabled by default 

tgross35:
By moving the logic for which platforms get symbols to `compiler_builtins` rather than rust-lang/rust, we can control where symbols get enabled without relying on Cargo features. Using Cargo features turned out to be a problem in [1].

This will help resolve errors like [2].

[1]: https://github.com/rust-lang/rust/issues/128358
[2]: https://github.com/rust-lang/rust/issues/128401


tgross35:
I was going to ping you once I figured this out but you beat me to it. Thanks, that is incredibly helpful.

Do you think it is reasonable to disable those targets here? As I mentioned in the diff I think it would be better in `rust-lang/rust`, but I can't come up with a good way to do it that doesn't rely on cargo features.

tgross35:
Are you testing on a version with LLVM19? I know https://github.com/llvm/llvm-project/issues/93894 caused a problem when we were updating builtins but that has since been resolved. I am pleasantly surprised if that was the only selection failure for interfaces.

tgross35:
> I don't think it matters whether the workaround for LLVM bugs to make `compiler-builtins` compile on all targets is in `rust-lang/rust` or `rust-lang/compiler-builtins`: the end result is (apart from Cargo resolver issues) the same.

Agreed. I am just not looking forward to the cycle of PR here -> release -> Rust PR -> Rust CI, if there are failures. But maybe this won't be much of an issue if there aren't too many completely broken targets.

I think the 7-31 nightly is the first one since https://github.com/rust-lang/rust/pull/127513, so I guess that was all. Awesome.

tgross35:
> * 32-bit `sparc`: f16 compiles; f128 fails due to https://github.com/llvm/llvm-project/issues/41838

Does this one actually crash or does it just have runtime issues?

tgross35:
Alright, I think this should be good to go. There were a handful of small changes needed to get CI passing again, there are more details in the commit messages.

tgross35:
Thanks. @Amanieu could you do a release? https://github.com/rust-lang/compiler-builtins/pull/657

kleisauke:
I noticed that `wasm32` is missing from this switch case:
https://github.com/rust-lang/compiler-builtins/blob/3ad4d9c3f4e56637da999cb0effa7b7f87a9c385/build.rs#L275-L288

Was this omission intentional? I'm encountering the same build failures described here:
https://github.com/rust-lang/compiler-builtins/blob/3ad4d9c3f4e56637da999cb0effa7b7f87a9c385/testcrate/build.rs#L56-L58

when building for `wasm32-unknown-emscripten` and using `-Zbuild-std`. However, I was able to work around the issue by also including `-Zbuild-std-features=compiler-builtins-no-f16-f128`.

tgross35:
We do seem to build successfully in rust-lang/rust for this target so I’m not sure what the difference is. But it makes sense to omit wasm32 for now if it is causing failures, mind putting up a PR?

kleisauke:
I just opened (draft) PR #665 for this.

PR #654, opened 2024-08-02
[test] check CI

tgross35:
Just see if CI passes on platforms where we no longer build C symbols.


tgross35:
This explains the failures in https://github.com/rust-lang/compiler-builtins/pull/652, on some platforms we were using symbols provided by compiler_builtins for our comparison basis. Ideally we would build the C compiler-rt and test against that, but not much sense in doing that now.

PR #656, opened 2024-08-04
Eliminate the use of `public_test_dep!`

tgross35:
Replace `public_test_dep!` by placing optionally public items into new modules, then controlling what is exported with the `public-test-deps` feature.

This is nicer for automatic formatting and diagnostics.


tgross35:
CI will fail until https://github.com/rust-lang/compiler-builtins/pull/652 lands.

tgross35:
It looks like this created some surprising inference at https://github.com/rust-lang/rust/pull/128691#issuecomment-2272297610. I don't really understand why, asked on Zulip https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Associated.20type.20suggestions.

The other change is more pressing to get into the main repo, so I am just going to revert this and revisit later.

PR #660, opened 2024-08-05
Mark symbols as weak on Windows and Apple as well

ZequanWu:


tgross35:
Some discussion on weak linkage for macos and windows https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/compiler_builtins.20undefined.20symbols/near/434550992 (unrelated to the failure).

aeubanks:
I'm not an expert in weak linkage for macos and windows and have heard that it's weird, but still does sorta work, as evidenced by this patch fixing some duplicate symbol issues in our builds where we also link LLVM compiler-rt builtins (that were previously working with the `weak-builtins` feature)

there are more places in `src/macros.rs` that have this same win/apple weak linkage check, those should probably be consistent

Amanieu:
Weak symbols are broken on windows-gnu targets, which is why we originally didn't enable weak symbols on Windows. Is that a target that you are using?

aeubanks:
we're using the win msvc targets, not the win gnu targets. if people have complained about win gnu before, maybe just restrict non-weak linkage to win gnu?

as for apple, are there known issues with weak linkage? again, that configuration was working for us before

tgross35:
I am going to close this as superseded by https://github.com/rust-lang/compiler-builtins/pull/672, which makes the symbols weak on everything but Windows GNU. This matches what we had before https://github.com/rust-lang/compiler-builtins/pull/598 for the `string.h` functions.

PR #665, opened 2024-08-11
Configure `f16` and `f128` support for WebAssembly

kleisauke:
See: https://github.com/rust-lang/compiler-builtins/pull/652#issuecomment-2282184106

Tested with:
<details>
  <summary>Details</summary>

```console
$ curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal --target wasm32-unknown-emscripten --default-toolchain nightly-2024-08-11 --component rust-src
$ source "$HOME/.cargo/env"
$ cargo new foo
$ cd foo
$ RUSTFLAGS="-Clink-arg=-Wl,--fatal-warnings" cargo build --release -Zbuild-std --target wasm32-unknown-emscripten
[...]
wasm-ld: error: function signature mismatch: __truncsfhf2
>>> defined as (f32) -> i32 in /home/kleisauke/foo/target/wasm32-unknown-emscripten/release/deps/libcompiler_builtins-903959ee8c3bdcea.rlib(compiler_builtins-903959ee8c3bdcea.compiler_builtins.f8125c6c5a7ecb3b-cgu.08.rcgu.o)
>>> defined as (f32) -> f32 in /home/kleisauke/foo/target/wasm32-unknown-emscripten/release/deps/libcompiler_builtins-903959ee8c3bdcea.rlib(compiler_builtins-903959ee8c3bdcea.compiler_builtins.f8125c6c5a7ecb3b-cgu.04.rcgu.o)

wasm-ld: error: function signature mismatch: __extendhfsf2
>>> defined as (i32) -> f32 in /home/kleisauke/foo/target/wasm32-unknown-emscripten/release/deps/libcompiler_builtins-903959ee8c3bdcea.rlib(compiler_builtins-903959ee8c3bdcea.compiler_builtins.f8125c6c5a7ecb3b-cgu.03.rcgu.o)
>>> defined as (f32) -> f32 in /home/kleisauke/foo/target/wasm32-unknown-emscripten/release/deps/libcompiler_builtins-903959ee8c3bdcea.rlib(compiler_builtins-903959ee8c3bdcea.compiler_builtins.f8125c6c5a7ecb3b-cgu.08.rcgu.o)
[...]
error: could not compile `foo` (bin "foo") due to 1 previous error
$ cat >> $(rustc --print sysroot)/lib/rustlib/src/rust/library/Cargo.toml <<EOL
compiler_builtins = { git = "https://github.com/kleisauke/compiler-builtins.git", branch = "wasm-f16-f128-feature" }
EOL
$ RUSTFLAGS="-Clink-arg=-Wl,--fatal-warnings" cargo build --release -Zbuild-std --target wasm32-unknown-emscripten
[...]
    Finished `release` profile [optimized] target(s) in 7.85s
```
</details>

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,44 +1,8 @@
 use std::{collections::BTreeMap, env, path::PathBuf, sync::atomic::Ordering};
 
-#[allow(dead_code)]
-struct Target {
-    triple: String,
-    os: String,
-    arch: String,
-    vendor: String,
-    env: String,
-    pointer_width: u8,
-    little_endian: bool,
-    features: Vec<String>,
-}
+mod configure;
 
-impl Target {
-    fn from_env() -> Self {
-        let little_endian = match env::var("CARGO_CFG_TARGET_ENDIAN").unwrap().as_str() {
-            "little" => true,
-            "big" => false,
-            x => panic!("unknown endian {x}"),
-        };
-
-        Self {
-            triple: env::var("TARGET").unwrap(),
-            os: env::var("CARGO_CFG_TARGET_OS").unwrap(),
-            arch: env::var("CARGO_CFG_TARGET_ARCH").unwrap(),
-            vendor: env::var("CARGO_CFG_TARGET_VENDOR").unwrap(),
-            env: env::var("CARGO_CFG_TARGET_ENV").unwrap(),
-            pointer_width: env::var("CARGO_CFG_TARGET_POINTER_WIDTH")
-                .unwrap()
-                .parse()
-                .unwrap(),
-            little_endian,
-            features: env::var("CARGO_CFG_TARGET_FEATURE")
-                .unwrap_or_default()
-                .split(",")
-                .map(ToOwned::to_owned)
-                .collect(),
-        }
-    }
-}
+use configure::{configure_f16_f128, Target};
 
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
@@ -259,49 +223,6 @@ fn configure_check_cfg() {
     // FIXME: these come from libm and should be changed there
     println!("cargo::rustc-check-cfg=cfg(feature, values(\"checked\"))");
     println!("cargo::rustc-check-cfg=cfg(assert_no_panic)");
-}
-
-/// Configure whether or not `f16` and `f128` support should be enabled.
-fn configure_f16_f128(target: &Target) {
-    // Set whether or not `f16` and `f128` are supported at a basic level by LLVM. This only means
-    // that the backend will not crash when using these types. This does not mean that the
-    // backend does the right thing, or that the platform doesn't have ABI bugs.
-    //
-    // We do this here rather than in `rust-lang/rust` because configuring via cargo features is
-    // not straightforward.
-    //
-    // Original source of this list:
-    // <https://github.com/rust-lang/compiler-builtins/pull/652#issuecomment-2266151350>
-    let (f16_ok, f128_ok) = match target.arch.as_str() {
-        // `f16` and `f128` both crash <https://github.com/llvm/llvm-project/issues/94434>
-        "arm64ec" => (false, false),
-        // `f16` crashes <https://github.com/llvm/llvm-project/issues/50374>
-        "s390x" => (false, true),
-        // `f128` crashes <https://github.com/llvm/llvm-project/issues/96432>
-        "mips64" | "mips64r6" => (true, false),
-        // `f128` crashes <https://github.com/llvm/llvm-project/issues/101545>
-        "powerpc64" if &target.os == "aix" => (true, false),
-        // `f128` crashes <https://github.com/llvm/llvm-project/issues/41838>
-        "sparc" | "sparcv9" => (true, false),
-        // `f16` miscompiles <https://github.com/llvm/llvm-project/issues/96438>
-        "wasm32" | "wasm64" => (false, true),
-        // Most everything else works as of LLVM 19
-        _ => (true, true),
-    };
-
-    // If the feature is set, disable these types.
-    let disable_both = env::var_os("CARGO_FEATURE_NO_F16_F128").is_some();
-
-    println!("cargo::rustc-check-cfg=cfg(f16_enabled)");
-    println!("cargo::rustc-check-cfg=cfg(f128_enabled)");
-
-    if f16_ok && !disable_both {
-        println!("cargo::rustc-cfg=f16_enabled");
-    }
-
-    if f128_ok && !disable_both {
-        println!("cargo::rustc-cfg=f128_enabled");
-    }
 }
 
 #[cfg(feature = "c")]
--- a/build.rs
+++ b/build.rs
@@ -283,6 +283,8 @@ fn configure_f16_f128(target: &Target) {
         "powerpc64" if &target.os == "aix" => (true, false),
         // `f128` crashes <https://github.com/llvm/llvm-project/issues/41838>
         "sparc" | "sparcv9" => (true, false),
+        // `f16` miscompiles <https://github.com/llvm/llvm-project/issues/96438>
+        "wasm32" | "wasm64" => (false, true),
         // Most everything else works as of LLVM 19
         _ => (true, true),
     };
--- a/configure.rs
+++ b/configure.rs
@@ -1,0 +1,86 @@
+// Configuration that is shared between `compiler_builtins` and `testcrate`.
+
+use std::env;
+
+#[allow(dead_code)]
+pub struct Target {
+    pub triple: String,
+    pub os: String,
+    pub arch: String,
+    pub vendor: String,
+    pub env: String,
+    pub pointer_width: u8,
+    pub little_endian: bool,
+    pub features: Vec<String>,
+}
+
+impl Target {
+    pub fn from_env() -> Self {
+        let little_endian = match env::var("CARGO_CFG_TARGET_ENDIAN").unwrap().as_str() {
+            "little" => true,
+            "big" => false,
+            x => panic!("unknown endian {x}"),
+        };
+
+        Self {
+            triple: env::var("TARGET").unwrap(),
+            os: env::var("CARGO_CFG_TARGET_OS").unwrap(),
+            arch: env::var("CARGO_CFG_TARGET_ARCH").unwrap(),
+            vendor: env::var("CARGO_CFG_TARGET_VENDOR").unwrap(),
+            env: env::var("CARGO_CFG_TARGET_ENV").unwrap(),
+            pointer_width: env::var("CARGO_CFG_TARGET_POINTER_WIDTH")
+                .unwrap()
+                .parse()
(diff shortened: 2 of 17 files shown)


build.rs
```diff
@@ -283,6 +283,13 @@ fn configure_f16_f128(target: &Target) {
         "powerpc64" if &target.os == "aix" => (true, false),
         // `f128` crashes <https://github.com/llvm/llvm-project/issues/41838>
         "sparc" | "sparcv9" => (true, false),
+        // `f16` requires half-precision feature, `f128` is unsupported
+        "wasm32" | "wasm64" => (
+            // TODO: This requires updating `WASM_ALLOWED_FEATURES`:
+            // https://github.com/rust-lang/rust/blob/1.80.0/compiler/rustc_target/src/target_features.rs#L310
+            target.features.contains(&"half-precision".to_owned()),
+            false,
```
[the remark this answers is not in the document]
kleisauke: Thanks! I've applied the changes. It turns out the `f128` builtins work perfectly on Wasm - I had mistakenly assumed they wouldn't work.

kleisauke:
I've updated the reproducer in the initial post. It seems that the `function signature mismatch` issues are treated as warnings by default, but they become visible when passing `RUSTFLAGS="-Clink-arg=-Wl,--fatal-warnings"`. This behavior is automatically enforced when using Emscripten's [`STRICT`](https://emsettings.surma.technology/#STRICT) mode.
https://github.com/emscripten-core/emscripten/blob/c614fa57ea8048a34341891dbd6f99fb78b6869a/tools/building.py#L267-L268

CI is still failing for `wasm32-unknown-unknown`. I think we need to conditionally disable this test somehow:
https://github.com/rust-lang/compiler-builtins/blob/3ad4d9c3f4e56637da999cb0effa7b7f87a9c385/testcrate/src/bench.rs#L353-L354

tgross35:
> I've updated the reproducer in the initial post. It seems that the `function signature mismatch` issues are treated as warnings by default, but they become visible when passing `RUSTFLAGS="-Clink-arg=-Wl,--fatal-warnings"`. This behavior is automatically enforced when using Emscripten's [`STRICT`](https://emsettings.surma.technology/#STRICT) mode. https://github.com/emscripten-core/emscripten/blob/c614fa57ea8048a34341891dbd6f99fb78b6869a/tools/building.py#L267-L268

Just to see, I tested making linker errors fatal on all targets in https://github.com/rust-lang/compiler-builtins/pull/667. `wasm32-unknown-unknown` seems to fail for other reasons.

> CI is still failing for `wasm32-unknown-unknown`. I think we need to conditionally disable this test somehow:
> 
> https://github.com/rust-lang/compiler-builtins/blob/3ad4d9c3f4e56637da999cb0effa7b7f87a9c385/testcrate/src/bench.rs#L353-L354

Ah, these will be the only targets that are disabled in `/build.rs` but tested in CI. I think we need to emit the same `enable_f16` and `enable_f128` feature in `testcrate`, then gate relevant things. To avoid duplicating the logic you should be able to do something like:

```rust
// testcrate/build.rs

mod builtins_build {
    include!("../build.rs");
}

fn main() {
    // ...

    // will need to mark some things public in `/build.rs`
    builtins_build::configure_f16_f128(builtins_build::Target::from_env());
}
```

I would only do the `enable_f16` gating for now since `enable_f128` isn't relevant for `testcrate` at this time.

kleisauke:
Thanks! I just pushed commit 42ac64efd234748c2163a65d9f19555e4cf9ed78. However, CI still fails for the `wasm32-unknown-unknown` target, curiously.

testcrate/build.rs
```diff
@@ -46,16 +50,18 @@ fn main() {
...
     // skip related tests. Most of these are missing `f16 <-> f32` conversion routines.
-    if (target.starts_with("aarch64-") && target.contains("linux"))
-        || target.starts_with("arm")
-        || target.starts_with("powerpc-")
-        || target.starts_with("powerpc64-")
-        || target.starts_with("powerpc64le-")
-        || target.starts_with("i586-")
-        || target.contains("windows-")
+    if (target.arch == "aarch64" && target.os == "linux")
+        || target.arch == "arm"
```
tgross35, line 54: This isn't the same since there are a few different architectures that start with `arm`, probably best to change it to `target.starts_with("arm")`.

configure.rs
```diff
@@ -0,0 +1,84 @@
+use std::env;
+
+#[allow(dead_code)]
+pub(crate) struct Target {
```
tgross35, line 4: Everything here can just be `pub` rather than `pub(crate)`, this isn't user-facing

build.rs
```diff
@@ -1,44 +1,8 @@
...
-            little_endian,
-            features: env::var("CARGO_CFG_TARGET_FEATURE")
-                .unwrap_or_default()
-                .split(",")
-                .map(ToOwned::to_owned)
-                .collect(),
-        }
-    }
-}
+use crate::configure::{configure_f16_f128, Target};
```
tgross35, line 5:
```suggestion
use configure::{configure_f16_f128, Target};
```

configure.rs
```diff
@@ -0,0 +1,84 @@
+use std::env;
```
tgross35, line 1:
```suggestion
//! Configuration that is shared between `compiler_builtins` and `testcrate`.

use std::env;
```

testcrate/src/bench.rs
```diff
@@ -350,7 +350,7 @@ macro_rules! impl_testio {
     }
 }
 
-#[cfg(not(feature = "no-f16-f128"))]
+#[cfg(all(f16_enabled, f128_enabled))]
 impl_testio!(float f16, f128);
```
tgross35, line 354: Since we may need one with the other disabled, these need to be separate

```rust
#[cfg(f16_enabled)]
impl_testio!(float f16);
#[cfg(f1238_enabled)]
impl_testio!(float f128);
```

tgross35 (commented):
> Thanks! I just pushed commit [42ac64e](https://github.com/rust-lang/compiler-builtins/commit/42ac64efd234748c2163a65d9f19555e4cf9ed78). However, CI still fails for the `wasm32-unknown-unknown` target, curiously.

Unfortunately the files in `benches/` also need to be updated, so anything that tests `f16` should get gated behind `f16_enabled` - both the `float_bench!` and the `criterion_group!` invocations will need adjusting. The `testcrate/src/bench.rs` change is needed too.

Some of the benches already have different `criterion_group!` per `cfg`, e.g. https://github.com/rust-lang/compiler-builtins/blob/3ad4d9c3f4e56637da999cb0effa7b7f87a9c385/testcrate/benches/float_extend.rs#L85-L102. When updating, it is probably better to not use `criterion_group` and just configure manually, something like

```rust
pub fn float_extend() {
    let mut criterion = Criterion::default().configure_from_args();

    #[cfg(f16_enabled)]
    {
        ...
    }
    
    extend_f32_f64(&mut criterion);
    extend_f32_f64(&mut criterion);
}
```

testcrate/build.rs
kleisauke: Good spot, changed to `target.arch.starts_with("arm")`.

configure.rs
kleisauke: Done.

configure.rs
kleisauke: Applied, thanks!

build.rs
kleisauke: Applied, thanks!

testcrate/src/bench.rs
kleisauke: Done.

kleisauke:
All done! This is now ready for review. Thanks to @beetrees and @tgross35 for the help :)

configure.rs
tgross35: Could you change to `//!`? Right now it is describing the `std::env` import :)

kleisauke: Ah, I tried that, but it failed with:
```
error[E0753]: expected outer doc comment
 --> testcrate/../configure.rs:1:1
  |
1 | //! Configuration that is shared between `compiler_builtins` and `testcrate`.
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2 | use std::env;
  | ------------- the inner doc comment doesn't annotate this `use` import
  |
help: to annotate the `use` import, change the doc comment from inner to outer style
  |
1 | /// Configuration that is shared between `compiler_builtins` and `testcrate`.
  |   ~
For more information about this error, try `rustc --explain E0753`.
error: could not compile `testcrate` (build script) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

kleisauke: Adding a new line doesn't help either, fwiw:
```
error[E0753]: expected outer doc comment
 --> testcrate/../configure.rs:1:1
  |
1 | //! Configuration that is shared between `compiler_builtins` and `testcrate`.
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2 |
3 | use std::env;
  | ------------- the inner doc comment doesn't annotate this `use` import
  |
help: to annotate the `use` import, change the doc comment from inner to outer style
  |
1 | /// Configuration that is shared between `compiler_builtins` and `testcrate`.
  |   ~
```

tgross35: Ah I guess that doesn't work well within `include!()`. We can call it good with a non-doc `//`

kleisauke: OK, done. :+1:

tgross35:
Thanks! You can submit a PR to update rust-lang/rust to 0.1.119 once https://github.com/rust-lang/compiler-builtins/pull/668 merges.

tgross35:
Ah, publish failed because `build.rs` uses multiple files. @kleisauke can you submit a PR to either add the new file to the `include` list https://github.com/rust-lang/compiler-builtins/blob/a4cd445d99d578c4f15836f9db1c822314f6316c/Cargo.toml#L17, or have `testcrate/build.rs` just include the entire `/build.rs` as mentioned here https://github.com/rust-lang/compiler-builtins/pull/665#issuecomment-2283008444?

kleisauke:
Apologies for the breakage. I'll submit a PR to add the new file to the `include` list.

Initially, I attempted to incorporate the entire `/build.rs` into `testcrate/build.rs`, but that approach failed, as shown here:
https://github.com/kleisauke/compiler-builtins/actions/runs/10356628453/job/28666867698#step:13:519

This is why I decided to move that logic into a separate file.

tgross35:
No worries, thanks for the clarification. I'll take a look when it is up.

kleisauke:
I just opened PR #669 for this.

tgross35:
0.1.119 published with that fix, feel free to submit a rust-lang/rust PR.

PR #667, opened 2024-08-11
Make all linker warnings errors in CI

tgross35:
Give us a better chance of detecting any kind of link issues or user-visible warnings in advance.

Inspired by https://github.com/rust-lang/compiler-builtins/pull/665#issuecomment-2282830722

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -14,7 +14,7 @@ env:
   CARGO_TERM_VERBOSE: true
   LIBM_BUILD_VERBOSE: true
   RUSTDOCFLAGS: -Dwarnings
-  RUSTFLAGS: -Dwarnings
+  RUSTFLAGS: -Dwarnings -Dlinker_messages
   RUST_BACKTRACE: full
   BENCHMARK_RUSTC: nightly-2026-08-05 # Pin the toolchain for reproducable results
 


tgross35:
Failure notes:

- Apple and thumb (via rust-lld) do not seem to support the flag
- `powerpc-` has errors from `bss-plt forced due to ...`. 
- wasm gets a lot of `is neither Wasm object file nor LLVM bitcode` warnings. These seem to be all the object files from the C code we build, e.g. `absvdi2`. I am not sure why we build the C versions in the first place, since the readme says they are never called by LLVM. There is some context to why `fatal-warnings` isn't passed on wasm, not directly related https://github.com/rust-lang/rust/pull/120278.
- windows-gnu has corrupt `.drectve` warnings. Looks like this is a known issue https://github.com/rust-lang/rust/issues/112368

ChrisDenton:
For MSVC it's [`/WX`](https://learn.microsoft.com/en-us/cpp/build/reference/wx-treat-linker-warnings-as-errors?view=msvc-170)

tgross35:
Does it ignore unrecognized link args? `fatal-warnings` is still getting passed and it doesn't seem to complain https://github.com/rust-lang/compiler-builtins/actions/runs/10343839608/job/28628523945?pr=667#step:11:23

ChrisDenton:
I mean... `/Wx` isn't being passed yet?

```
> rustc hello.rs -C link-args='/WX --fatal-warnings'

error: linking with `link.exe` failed: exit code: 4044
[...]
  = note: LINK : warning LNK4044: unrecognized option '/-fatal-warnings'; ignored
          LINK : error LNK1218: warning treated as error; no output file generated
```

tgross35:
Oh, unrecognized linker args are warnings by default rather than errors? I was just expecting it to complain like the other targets that don't support `-Wl,--fatal-warnings` do.

tgross35:
The wasm issue is https://github.com/rust-lang/rust/issues/132802, I still haven't updated the Windows flags.

tgross35:
Since 1.97, `-Dlinker_warnings` should do most of what we're interested in https://blog.rust-lang.org/2026/07/09/Rust-1.97.0/.

Previous version, for reference:

```sh
# Make all linker warnings errors
if [[ "$target" == *"-apple-"* || "$target" == "thumb"* ]]; then
    extra_rustflags=""
else
    extra_rustflags=" -Clink-arg=-Wl,--fatal-warnings"
fi

export RUSTFLAGS="${RUSTFLAGS:-}$extra_rustflags"
```

PR #672, opened 2024-08-22
Fix weak linkage on windows and apple platforms

### LINKED ISSUE #653: Various symbols on win/apple are no longer weak after #598
After upgrading our rust toolchain version, on multiple win/apple platforms (arm64 win specifically for this error) we're seeing link errors like

```
../../third_party/llvm-build/Release+Asserts/bin/lld-link --rsp-quoting=posix "/OUT:./test_rust_metadata_cc_exe.exe" /nologo -libpath:../../third_party/llvm-build/Release+Asserts/lib/clang/20/lib/windows /winsysroot:../../third_party/depot_tools/win_toolchain/vs_files/7393122652 /MACHINE:ARM64  "/PDB:./test_rust_metadata_cc_exe.exe.pdb" "@./test_rust_metadata_cc_exe.exe.rsp"
lld-link: error: duplicate symbol: __udivti3
>>> defined at libcompiler_builtins_compiler_builtins.rlib(libcompiler_builtins_compiler_builtins.compiler_builtins.878e573648bac277-cgu.0.rcgu.o)
>>> defined at clang_rt.builtins-aarch64.lib(udivti3.c.obj)

lld-link: error: duplicate symbol: __umodti3
>>> defined at libcompiler_builtins_compiler_builtins.rlib(libcompiler_builtins_compiler_builtins.compiler_builtins.878e573648bac277-cgu.3.rcgu.o)
>>> defined at clang_rt.builtins-aarch64.lib(umodti3.c.obj)

lld-link: error: duplicate symbol: __muloti4
>>> defined at libcompiler_builtins_compiler_builtins.rlib(libcompiler_builtins_compiler_builtins.compiler_builtins.878e573648bac277-cgu.0.rcgu.o)
>>> defined at clang_rt.builtins-aarch64.lib(muloti4.c.obj)
ninja: build stopped: subcommand failed.
```

I believe this is due to #598. We were using the weak-intrinsics feature before. #598 removes that feature and makes everything weak by default, except for on win/apple platforms. That seems to be because some intrinsics unconditionally were not marked weak on win/apple. However, not all intrinsics were like that, e.g. the symbols in the error message above.

The following diff seems to make the error go away

```
diff --git a/src/macros.rs b/src/macros.rs
index fb14660..721d242 100644
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -464,7 +464,7 @@ macro_rules! intrinsics {
         mod $name {
             $(#[$($attr)*])*
             #[no_mangle]
-            #[cfg_attr(all(not(windows), not(target_vendor = "apple")), linkage = "weak")]
+            #[linkage = "weak"]
             $(unsafe $($empty)?)? extern $abi fn $name( $($argname: $ty),* ) $(-> $ret)? {
                 super::$name($($argname),*)
             }
```

Amjad50:
Fixes #653 

There were some issues regarding windows and apple platform, we were exporting symbols that are already provided by the compiler but weren't marked as `weak` which resulted in conflicted symbols in the linking process.

Initially, we didn't add `weak` because we thought it is not supported on windows and apple platforms, but it looks like its only not supported on windows-gnu platforms

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -256,7 +256,7 @@ macro_rules! intrinsics {
         #[cfg(all(any(windows, target_os = "uefi"), target_arch = "x86_64", not(feature = "mangled-names")))]
         mod $name {
             #[no_mangle]
-            #[cfg_attr(all(not(windows), not(target_vendor = "apple")), linkage = "weak")]
+            #[cfg_attr(not(all(windows, target_env = "gnu")), linkage = "weak")]
             extern $abi fn $name( $($argname: $ty),* )
                 -> $crate::macros::win64_128bit_abi_hack::U64x2
             {
@@ -298,7 +298,7 @@ macro_rules! intrinsics {
         #[cfg(all(target_arch = "arm", not(feature = "mangled-names")))]
         mod $name {
             #[no_mangle]
-            #[cfg_attr(all(not(windows), not(target_vendor = "apple")), linkage = "weak")]
+            #[cfg_attr(not(all(windows, target_env = "gnu")), linkage = "weak")]
             $(#[$($attr)*])*
             extern $abi fn $name( $($argname: $ty),* ) $(-> $ret)? {
                 super::$name($($argname),*)
@@ -308,7 +308,7 @@ macro_rules! intrinsics {
         #[cfg(all(target_arch = "arm", not(feature = "mangled-names")))]
         mod $alias {
             #[no_mangle]
-            #[cfg_attr(all(not(windows), not(target_vendor="apple")), linkage = "weak")]
+            #[cfg_attr(not(all(windows, target_env = "gnu")), linkage = "weak")]
             $(#[$($attr)*])*
             extern "aapcs" fn $alias( $($argname: $ty),* ) $(-> $ret)? {
                 super::$name($($argname),*)
@@ -375,7 +375,7 @@ macro_rules! intrinsics {
         mod $name {
             $(#[$($attr)*])*
             #[no_mangle]
-            #[cfg_attr(all(not(windows), not(target_vendor = "apple")), linkage = "weak")]
+            #[cfg_attr(not(all(windows, target_env = "gnu")), linkage = "weak")]
             unsafe extern $abi fn $name( $($argname: $ty),* ) $(-> $ret)? {
                 super::$name($($argname),*)
             }
@@ -400,7 +400,7 @@ macro_rules! intrinsics {
             #[naked]
             $(#[$($attr)*])*
             #[cfg_attr(not(feature = "mangled-names"), no_mangle)]
-            #[cfg_attr(all(not(windows), not(target_vendor = "apple")), linkage = "weak")]
+            #[cfg_attr(not(all(windows, target_env = "gnu")), linkage = "weak")]
             pub unsafe extern $abi fn $name( $($argname: $ty),* ) $(-> $ret)? {
                 $($body)*
             }
@@ -467,7 +467,7 @@ macro_rules! intrinsics {
         mod $name {
             $(#[$($attr)*])*
             #[no_mangle]
-            #[cfg_attr(all(not(windows), not(target_vendor = "apple")), linkage = "weak")]
+            #[cfg_attr(not(all(windows, target_env = "gnu")), linkage = "weak")]
             $(unsafe $($empty)?)? extern $abi fn $name( $($argname: $ty),* ) $(-> $ret)? {
                 super::$name($($argname),*)
             }


tgross35:
Thanks for the fix, this matches what was in `src/mem/mod.rs` before https://github.com/rust-lang/compiler-builtins/pull/598. What exactly happens on windows GNU?

tgross35:
Could you create a PR to rust-lang/rust updating compiler-builtins 0.1.120 after the release PR merges? https://github.com/rust-lang/compiler-builtins/pull/671

Amjad50:
Looks like it crashes for some reason on windows-gnu, for some reason with `weak` it gives `undefined reference` errors: there is a PR that tried to fix the issue by making everything `weak` (#660)

Amjad50:
> Could you create a PR to rust-lang/rust updating compiler-builtins 0.1.120 after the release PR merges? #671

Sure

mati865:
> Looks like it crashes for some reason on windows-gnu, for some reason with `weak` it gives `undefined reference` errors: there is a PR that tried to fix the issue by making everything `weak` (#660)

Do you have full error messages?
AFAIK weak symbols are unsupported in `windows-msvc`, there is some support in `windows-gnu` (GCC+Binutils), but it works only with static libraries and they mostly work with `windows-gnullvm`, although there are some DLL inherited limitations with shared libraries.

src/macros.rs
```diff
@@ -256,7 +256,7 @@ macro_rules! intrinsics {
         #[cfg(all(any(windows, target_os = "uefi"), target_arch = "x86_64", not(feature = "mangled-names")))]
         mod $name {
             #[no_mangle]
-            #[cfg_attr(all(not(windows), not(target_vendor = "apple")), linkage = "weak")]
+            #[cfg_attr(not(all(windows, target_env = "gnu")), linkage = "weak")]
```
mati865, line 259: It should likely also check `not(target_abi = "llvm"`, how do you test it? I'd like to test with gnullvm.

tgross35:
> > Looks like it crashes for some reason on windows-gnu, for some reason with `weak` it gives `undefined reference` errors: there is a PR that tried to fix the issue by making everything `weak` (#660)
> 
> Do you have full error messages?

The error messages are at the PR linked there #660, direct link https://github.com/rust-lang/compiler-builtins/actions/runs/10257106333/job/28377511730?pr=660. Basically a lot of `Warning: .drectve `-exclude-symbols:"..." ' unrecognized␍`, followed by `undefined reference`es. Possibly related to https://github.com/rust-lang/rust/issues/112368.

> AFAIK weak symbols are unsupported in `windows-msvc`, there is some support in `windows-gnu` (GCC+Binutils), but it works only with static libraries and they mostly work with `windows-gnullvm`, although there are some DLL inherited limitations with shared libraries.

MSVC does support some form of weak linking, see https://github.com/rust-lang/compiler-builtins/issues/653 and a comment about DLLs there. We were already using this for the `string.h` functions so it seems to have worked, see the changes to `src/mem/mod.rs` in https://github.com/rust-lang/compiler-builtins/pull/598.

Are you usually able to link windows-gnu binaries with weak symbols?

src/macros.rs
tgross35: Just testing this crate locally? You should be able to `cd` to `testcrate` and `cargo test --target ...`.

mati865:
> The error messages are at the PR linked there #660, direct link [rust-lang/compiler-builtins/actions/runs/10257106333/job/28377511730?pr=660](https://github.com/rust-lang/compiler-builtins/actions/runs/10257106333/job/28377511730?pr=660).

Bummer, it requires inspection of the objects and won't be as simple as hoped.

> Basically a lot of `Warning: .drectve `-exclude-symbols:"..." ' unrecognized␍`, followed by `undefined reference`es. Possibly related to [rust-lang/rust#112368](https://github.com/rust-lang/rust/issues/112368).

Yeah, that's caused by an outdated linker. It's harmless but causes binaries to contain more symbols than they need to (MSVC behaves the same way modulo the warnings as this is LLVM's invention).

> MSVC does support some form of weak linking, see https://github.com/rust-lang/compiler-builtins/issues/653 and a comment about DLLs there.

Oh, I haven't heard about them before. Regarding DLLs, there is a "we have weak symbols at home" attribute called `selectany`, while it's not nearly as powerful as weak symbols, it's sometimes used to emulate it. What it does is making the linker pick the first occurrence of a symbol when there are multiple definitions.

tgross35:
Do you have any specific concerns about this, or just that it would be ideal if they worked on windows-gnu too? I ran a msvc try job on the update PR at https://github.com/rust-lang/rust/pull/129400 and it passed, so I don't think there is anything too worrying. Especially since it seems builtins always gets statically linked, per my comment here https://github.com/rust-lang/compiler-builtins/issues/653#issuecomment-2304685975 - are weak symbols for DLLs still relevant with this in mind?

For what it's worth, if there is a newer linker version that doesn't hit the listed errors then feel free to send a PR updating our CI jobs here as needed. I ran into it here too https://github.com/rust-lang/compiler-builtins/pull/667.

src/macros.rs
mati865: Oh, I thought this issue would manifest only during Rust build, gonna look into it this weekend.

mati865:
> Do you have any specific concerns about the change here, or just that it would be ideal if they worked on windows-gnu too?

Regarding `windows-gnu` just a slight concern, something might be still broken and was just hidden.
When it comes to `windows-gnuillvm` (as the maintainer), I believe it should work on par with both MSVC and GNU. Otherwise it has to be fixed in LLVM.

> Especially since it seems builtins always gets statically linked, per my comment here https://github.com/rust-lang/compiler-builtins/issues/653#issuecomment-2304685975 - are weak symbols for DLLs still relevant with this in mind?

There should be no problem for MSVC since weak symbols were just proven to work. I'm concerned only about windows-gnu breaking during some supposedly irrelevant changes.

> For what it's worth, if there is a newer linker version that doesn't hit the listed errors then feel free to send a PR updating our CI jobs here as needed. I ran into it here too https://github.com/rust-lang/compiler-builtins/pull/667.

I attempted to build a new version with recent components using crosstool-ng to save time (I was terribly wrong here...) and wasted dozens of hours on it: https://github.com/mati865/mingw-build https://github.com/rust-lang/rust/pull/119229
But the mingw-builds project is getting a new release, so hopefully it can be solved soonish.

src/macros.rs
tgross35: The rust-lang/rust update of compiler-builtins merged, so it should also be available to test with the next rustc nightly (couple hours after 0:00 UTC)

tgross35:
`windows-gnu` shouldn't have ever had weak intrinsics now or before, right?

Would you mind opening a new issue so we can keep a better eye on this?

mati865:
> `windows-gnu` shouldn't have ever had weak intrinsics now or before, right?

TBH I have no idea how they are used and whats their purpose, so I cannot answer that.

> Would you mind opening a new issue so we can keep a better eye on this?

Regarding `cfg` attributes from this PR or something else?

---

FWIW I've tested both windows-gnu and windows-gnullvm locally with PR reverted. Running `cargo test --target i686-pc-windows-gnu` (similarly for gnullvm) inside `testcrate/` works fine, even `./ci/run.sh i686-pc-windows-gnu` got to the point when it errored about missing compiler-rt:
```
  thread 'main' panicked at build.rs:600:17:
  RUST_COMPILER_RT_ROOT is not set. You may need to download compiler-rt.
```
Either CI GNU C toolchain (the linker and stuff) has a bug, or this error happens before reaching CI's error.

---

Also it struck me just now, the 64-bit target has passed, so this might be an issue with name mangling. 32-bit targets (all of them, be it GNU, LLVM, MSVC) use additional underscore before symbol names. Maybe that's the problem here?

tgross35:
> > Would you mind opening a new issue so we can keep a better eye on this?
> 
> Regarding `cfg` attributes from this PR or something else?

Whatever you would like to continue investigating or are still looking into :) I just don't want to continue discussion on a merged PR, so you can repost the above couple comments to a new issue.

PR #673, opened 2024-08-22
Use array simd in `U64x2`

scottmcm:
Another step in https://github.com/rust-lang/compiler-team/issues/621


tgross35:
Coolieo, thanks

tgross35:
@scottmcm release for 0.1.121 will be available after https://github.com/rust-lang/compiler-builtins/pull/674 merges, then we can update std. There is an in-progress PR for 0.1.120, should probably let that go through then send another update https://github.com/rust-lang/rust/pull/129400

PR #675, opened 2024-08-23
Fix ABI for `f16` builtins on Intel Apple targets

beetrees:
For builtins on x86 (32-bit and 64-bit) Apple platforms, `f16` is passed and returned like a `u16` unless the builtin involves `f128`. This PR adds two `#[win64_128bit_abi_hack]`-like attributes (`#[apple_f16_arg_abi]` and `#[apple_f16_ret_abi]`) to the `intrinsics` macro to handle this. This should fix the CI failure that's blocking rust-lang/rust#128349.

### DIFF
--- a/src/float/extend.rs
+++ b/src/float/extend.rs
@@ -86,6 +86,7 @@ intrinsics! {
 intrinsics! {
     #[avr_skip]
     #[aapcs_on_arm]
+    #[apple_f16_arg_abi]
     #[arm_aeabi_alias = __aeabi_h2f]
     #[cfg(f16_enabled)]
     pub extern "C" fn __extendhfsf2(a: f16) -> f32 {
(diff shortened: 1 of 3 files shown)


tgross35:
Any clue what the history is here? Maybe `f16` as GPRs make sense in some legacy aspect, but changing the ABI when `f128` is involved seems like it could be an unintentional choice by LLVM. Are existing runtime libs even achieving this?

[Apple's docs](https://developer.apple.com/documentation/xcode/writing-64-bit-intel-code-for-apple-platforms) say it defers to the psABI, which does specify both `_Float16` and `__float128` since June.

tgross35:
Thanks for the background. I have access to an arm64 mac, I'll try to double check this locally a bit later.

There are apple systems on cfarm if you have access to that, just not x86 either https://portal.cfarm.net/machines/list/ (If you don't have access, it is pretty quick. I think I got mine in about one day.)

tgross35:
I didn't get the chance to test this but it looks reasonable to me with everything in mind. I can't really think of a straightforward way to test this in crate, but I think we will be covered by rust-lang/rust.

tgross35:
https://github.com/rust-lang/compiler-builtins/pull/677 should release 0.1.122. https://github.com/rust-lang/rust/pull/129481 is already enqueued but rust-lang/rust can be bumped again after that, and probably enable the x86 apple tests.

PR #676, opened 2024-08-24
Use `macos-latest` in CI for `aarch64-apple-darwin`

beetrees:
Per [the GitHub actions docs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories), `macos-latest` now refers to the M1-CPU-using `macos-14`. This PR changes the `aarch64-apple-darwin` CI to match the other CI targets, which all use the `-latest` OS versions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
           os: ubuntu-latest
           rust: nightly
         - target: x86_64-apple-darwin
-          os: macos-latest
+          os: macos-13
           rust: nightly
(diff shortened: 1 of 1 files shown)


tgross35:
Oh that is going to be an interesting change. Could you also change the x86 image to be macos-13?

I am pretty surprised they didn't distinguish these further than whatever the `latest` happens to be.

tgross35:
Thanks for catching this, this sort of change probably has to happen at a ton of our repos.

PR #678, opened 2024-08-26
Remove unsupported *vfp fn

sbechet:
Hello @Amanieu and @tgross35,

I update README and remove fn as i promised. I remove tests too.

See closed #626 pull request for more informations.


Amanieu:
CI is failing because of missing newlines at the end of some files.

tgross35:
For reference these were added in https://github.com/rust-lang/compiler-builtins/pull/225 (cc @paoloteti). I can't even get llvm to emit these with the target Amanieu mentioned in https://github.com/rust-lang/compiler-builtins/pull/626#issuecomment-2184042547, unless there is a specific config needed https://godbolt.org/z/nra4Pb43Y.

Thanks for the cleanup.

PR #680, opened 2024-08-27
Don't include `math` for `unix` and `wasi` targets

Amjad50:
This fixes such as (https://github.com/rust-lang/rust/issues/128386) where, our implementation is being used on systems where there is already `math` library and its more performant and accurate.

So with this change, linux will go back to the previous behavior and not include these functions, windows and apple were generally not affected.

Looking at the targets we have builtin now in rust, everything else is probably good to have the math symbols.

> A note on the above, the `hermit` os uses `libm` directly for itself, but I think its Ok to keep providing math in `compiler_builtin` for it, its technically the same implementation either from `compiler_builtin`  or `hermit-builtins`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,9 +47,16 @@ mod macros;
 pub mod float;
 pub mod int;
 
-// Disabled on x86 without sse2 due to ABI issues
-// <https://github.com/rust-lang/rust/issues/114479>
-#[cfg(not(all(target_arch = "x86", not(target_feature = "sse2"))))]
+// Disable for any of the following:
+// - x86 without sse2 due to ABI issues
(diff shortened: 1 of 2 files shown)


tgross35:
Thanks for posting this. Should this keep the same config as before https://github.com/rust-lang/compiler-builtins/pull/594/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759? `uefi`, `xous`, etc

tgross35:
Oh nevermind, I read that backwards

tgross35:
Thanks for this, seems reasonable to me. Once https://github.com/rust-lang/compiler-builtins/pull/681 merges could you create a PR to rust-lang/rust updating builtins to 0.1.123?

PR #682, opened 2024-08-31
[hexagon] Remove aliases w/o leading __

androm3da:
These hexagon builtins incorrectly created aliases in the global namespace which can (and in at least one case, did) conflict with symbols defined by other programs.

This should address the issue reported as https://github.com/rust-lang/rust/issues/129823:

	   Compiling compiler_builtins v0.1.123
	   Compiling core v0.0.0 (/home/ben/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core)
	   Compiling rustc-std-workspace-core v1.99.0 (/home/ben/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/rustc-std-workspace-core)
	   Compiling byteorder v1.5.0
	   Compiling zerocopy v0.7.34
	error: symbol 'fma' is already defined

	error: could not compile `compiler_builtins` (lib) due to 1 previous error

Also: some of the symbols defined were not just aliases, so those are now qualified with `__hexagon_`.  The compiler does not yet emit calls to these particular ones, but if/when it does, it can use the new names.

### DIFF
--- a/src/hexagon/dffma.s
+++ b/src/hexagon/dffma.s
@@ -3,8 +3,6 @@
         .type __hexagon_fmadf4,@function
  .global __hexagon_fmadf5
         .type __hexagon_fmadf5,@function
- .global fma
- .type fma,@function
  .global __qdsp_fmadf5 ; .set __qdsp_fmadf5, __hexagon_fmadf5
  .p2align 5
 __hexagon_fmadf4:
--- a/src/hexagon/dfminmax.s
+++ b/src/hexagon/dfminmax.s
@@ -1,17 +1,12 @@
  .text
  .global __hexagon_mindf3
  .global __hexagon_maxdf3
- .global fmin
- .type fmin,@function
- .global fmax
- .type fmax,@function
  .type __hexagon_mindf3,@function
  .type __hexagon_maxdf3,@function
  .global __qdsp_mindf3 ; .set __qdsp_mindf3, __hexagon_mindf3
  .global __qdsp_maxdf3 ; .set __qdsp_maxdf3, __hexagon_maxdf3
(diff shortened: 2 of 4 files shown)


androm3da:
@tgross35 or @Amanieu are you willing to merge this change?

tgross35:
Does changing the symbols from `.global` to `.weak` resolve this issue without losing whatever benefit there is to having them in the first place? That is effectively what we do for the symbols on other platforms.

> @tgross35 or @Amanieu are you willing to merge this change?

Fyi we get pinged on new PRs already, just need to allow some time for reviews (this PR has only been open 8 hours).

androm3da:
It would, but it was an error to define them that way at all - I don't believe we ever had calls to the unqualified symbols.  Probably just an artifact from how it was originally written and tested.

> benefit there is to having them in the first place

I don't think there is any.

> Fyi we get pinged on new PRs already

Ok, my mistake.  I'm not in any hurry to get this merged, just wanted to avoid it falling through the cracks.  Good to know it wouldn't have.

tgross35 (approved):
Makes sense. Thanks for confirming!

tgross35:
This change will be in 0.1.124, you can put up a PR to rust-lang/rust to update compiler-builtins after https://github.com/rust-lang/compiler-builtins/pull/683 merges.

androm3da:
> you can put up a PR to rust-lang/rust to update compiler-builtins after https://github.com/rust-lang/compiler-builtins/pull/683 merges.

Ok - will, do.  Thanks!

PR #686, opened 2024-09-10
use `naked_asm!` in `#[naked]` functions

folkertdev:
Part of an effort to stabilize `#[naked]` functions https://github.com/rust-lang/rust/issues/90957

This PR uses `naked_asm!` instead of `asm!` in all `#[naked]` functions, and removes `options(noreturn)` which is implicit with `naked_asm!`

the PR introducing `naked_asm!` just got merged https://github.com/rust-lang/rust/pull/130146, so it'll likely take some time before this can get merged. Does `compiler-builtins` use the latest nightly, or is it a little behind?

cc @Amanieu

### DIFF
--- a/src/aarch64.rs
+++ b/src/aarch64.rs
@@ -6,7 +6,7 @@ intrinsics! {
     #[naked]
     #[cfg(all(target_os = "uefi", not(feature = "no-asm")))]
     pub unsafe extern "C" fn __chkstk() {
-        core::arch::asm!(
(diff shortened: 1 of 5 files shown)


tgross35:
https://github.com/rust-lang/rust/pull/130146 only recently merged so this just needs to wait for the nightly (few hours after 00:00 UTC)

folkertdev:
all good now

PR #692, opened 2024-09-26
Fixed `__divtf3` having wrong cfg for f128

Tazdevil971:
This caused target with `f128` automatically disabled (as opposed to disabled via `no-f16-f128`) to not compile and complain about `f128` being unstable


tgross35:
Ah that’s right, thanks!

PR #702, opened 2024-09-30
Switch repository layout to use a virtual manifest

tgross35:
The current setup has the `Cargo.toml` for `compiler-builtins` at the
repository root, which means all support crates and other files are
located within the package root. This works for now but is not the
cleanest setup since files that should or shouldn't be included in the
package need to be configured in `Cargo.toml`. If we eventually merge
`libm` development into this repository, it would be nice to make this
separation more straightforward.

Begin cleaning things up by moving the crate source to a new
`compiler-builtins` directory and adding a virtual manifest. For now the
`libm` submodule is also moved, but in the future it can likely move
back to the top level (ideally `compiler-builtins/src` would contain a
symlink to `libm/src/math`, but unfortunately it seems like Cargo does
not like something about the submodule + symlink combination).


RalfJung:
An unfortunate side-effect of this PR is that nothing builds any more after a `git clone`; one now has to initialize submodules. :/

tgross35:
I think that might have been from https://github.com/rust-lang/compiler-builtins/pull/763 rather than this PR, since `libm` is now always required. I'm hoping to merge the two repos at some point since I'd love to reuse a lot of the test infrastructure, and submodule annoyances - should be in the near future now that I finally figured out how make this PR work.

PR #707, opened 2024-10-06
fix(int): avoid infinite recursion on left shift

raskyld:
Please, see this discussion for the full
context: https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/.5Bwasm32.5D.20Infinite.20recursion.20.60compiler-builtins.60.20.60__multi3.60


tgross35:
I wanted to add a comment but it wouldn't let me do that here, so I just picked up this patch and another one and merged it in https://github.com/rust-lang/compiler-builtins/pull/708. Thank you for the fix!

PR #708, opened 2024-10-06
 fix(int): avoid infinite recursion on left shift #707 

tgross35:
Recreated PR of https://github.com/rust-lang/compiler-builtins/pull/707 with an adjusted message and a second patch adding a note.


tgross35:
Cc @thejpster, this should hopefully fix your sparc failure

PR #710, opened 2024-10-06
Implement copysign for libcalls

workingjubilee:
As mentioned elsewhere[^1], it's suspiciously easy to do this, so why doesn't LLVM handle it?

The answer is it might, in fact, do that[^2], so accepting this would be a paranoid stopgap at best. Nonetheless, if it pleases you, you may!

[^1]: https://github.com/rust-lang/rust/pull/131304
[^2]: https://github.com/llvm/llvm-project/pull/111269

### DIFF
--- a/src/float/bitops.rs
+++ b/src/float/bitops.rs
@@ -12,6 +12,7 @@ fn fabs_f64(f: f64) -> f64 {
 }
 
 // FIXME: delete when we move fabs to core and it reaches stable
+#[cfg(f128_enabled)]
 fn fabs_128(f: f128) -> f128 {
     f128::from_bits(f.to_bits() & const { !(i128::MIN as u128) })
 }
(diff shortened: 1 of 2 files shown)


tgross35:
`abs` is useful enough that you should just put it on the `Float` trait in `src/float/mod.rs` honestly, we use it everywhere (e..g. https://github.com/rust-lang/compiler-builtins/blob/8a610fbbd16a3245822695e3af0d5d0fcf8fc8e2/src/float/div.rs#L115-L116).

This should get something in `testcrate/tests`, and a call in `examples/intrinsics.rs` to make sure either linking or a fallback works (I think you can just call the intrinsic here since the API isn't in core).

tgross35:
Also we do have this symbol in `libm` (excluding the f128 version), it just isn't exposed (though we would like it to be). It seems fine to have it here until the `libm` situation improves somewhat, but I think this would be the first symbol that isn't part of compiler-rt.

I'll still defer to @Amanieu to confirm that though

workingjubilee:
The evolution of this is my determination that we don't need it in https://github.com/llvm/llvm-project/pull/111269

Amanieu:
Is it possible we might need copysign to support gcc-generated code? This may be need for musl targets where we ship a gcc-compiled musl.

tgross35:
We could probably make adjustments so that we can expose only specific functions from our `libm` on all platforms, then do so for the symbols here rather than duplicating them in builtins. Instead of the current all-or-nothing `libm`. (I think we might want this in the future anyway)

workingjubilee:
> Is it possible we might need copysign to support gcc-generated code? This may be need for musl targets where we ship a gcc-compiled musl.

this feels unlikely as there was no situation in which LLVM generated copysign for a type that we also use. if gcc generates that much worse code, then I suggest we file a compiler bug.

PR #715, opened 2024-10-21
Re-enable math module on i686-unknown-uefi

YtvwlD:
In 9ba77d1583e6de5ab9cf7c9b82827ba8fcb9062f, this was disabled for x86 without sse2. It should be fine to re-enable it for UEFI, as explained at <https://github.com/rust-lang/rust/issues/128533#issuecomment-2408699671>.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,10 +44,16 @@ pub mod int;
 // Disable for any of the following:
 // - x86 without sse2 due to ABI issues
 //   - <https://github.com/rust-lang/rust/issues/114479>
+//   - but exclude UEFI since it is a soft-float target
+//     - <https://github.com/rust-lang/rust/issues/128533>
 // - All unix targets (linux, macos, freebsd, android, etc)
 // - wasm with known target_os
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -44,10 +44,16 @@ pub mod int;
 // Disable for any of the following:
 // - x86 without sse2 due to ABI issues
 //   - <https://github.com/rust-lang/rust/issues/114479>
+//   - but re-enable for UEFI
+//     - <https://github.com/rust-lang/rust/issues/128533>
```
tgross35, line 48: You can just say:

```rust
//   - Exclude UEFI since it is a soft-float target
```

So we know why without linking back to the issue

tgross35:
Thanks for the fix, one nit and then lgtm

tgross35:
Thanks, you can open a PR bumping rust-lang/rust to 0.1.135 as soon as https://github.com/rust-lang/compiler-builtins/pull/716 merges.

PR #717, opened 2024-10-25
Update licensing to MIT AND Apache-2.0 WITH LLVM-exception

tgross35:
Currently, Cargo.toml specifies Apache-2.0 OR MIT, but LICENSE.txt describes MIT OR NCSA. compiler-builtins is derived from LLVM's compiler-rt. LICENSE.txt correctly reflects the state of compiler-rt prior to relicensing on 2019-01-19, during which time software was available for use under either MIT or the University of Illinois NCSA license. After relicensing, however, compiler-rt is available for use only under Apache-2.0 with the LLVM exception; this is not reflected anywhere in the repository.

Update the SPDX license identifier to `MIT AND Apache-2.0 WITH LLVM-exception AND (MIT OR Apache-2.0)`. Each AND portion covers something specific:

* Apache-2.0 WITH LLVM-exception: this covers work that is derived from the LLVM repository since after the LLVM relicensing.
* MIT: This covers work that is derived from LLVM before the LLVM relicensing (under MIT OR NCSA), as well as the vendored `libm` components.
* MIT AND Apache-2.0: This ensures that any contributions to this repository, in addition to meeting the above required licenses, is also released for use under the Rust-standard Apache-2.0 with no LLVM exception.

See also the parallel license update in rust-lang/libm [1].

Fixes: https://github.com/rust-lang/compiler-builtins/issues/307
Closes: https://github.com/rust-lang/compiler-builtins/pull/511
Link: https://rust-lang.zulipchat.com/#narrow/channel/335408-foundation/topic/Request.20for.20legal.20team.20input.20on.20crate.20licensing
Link: https://github.com/rust-lang/libm/pull/317 [1]

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 authors = ["Jorge Aparicio <japaricious@gmail.com>"]
 name = "compiler_builtins"
 version = "0.1.135"
-license = "MIT/Apache-2.0"
+license = "MIT AND Apache-2.0 WITH LLVM-exception AND (MIT OR Apache-2.0)"
 readme = "README.md"
 repository = "https://github.com/rust-lang/compiler-builtins"
 homepage = "https://github.com/rust-lang/compiler-builtins"
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,61 +1,15 @@
-==============================================================================
-compiler-builtins License
-==============================================================================
+compiler-builtins as a whole is available for use under both the MIT license
+and the Apache-2.0 license with the LLVM exception (MIT AND Apache-2.0 WITH
+LLVM-exception).
 
-The compiler-builtins crate is dual licensed under both the University of
-Illinois "BSD-Like" license and the MIT license.  As a user of this code you may
-choose to use it under either license.  As a contributor, you agree to allow
-your code to be used under both.
+As a contributor, you agree that your code can be used under either the MIT
+license, or the Apache-2.0 license, or the Apache-2.0 license with the LLVM
+exception.
 
-Full text of the relevant licenses is included below.
+Text of the relevant licenses is provided below:
 
-==============================================================================
-
-University of Illinois/NCSA
-Open Source License
-
-Copyright (c) 2009-2016 by the contributors listed in CREDITS.TXT
-
-All rights reserved.
-
-Developed by:
-
-    LLVM Team
-
-    University of Illinois at Urbana-Champaign
-
-    http://llvm.org
-
-Permission is hereby granted, free of charge, to any person obtaining a copy of
-this software and associated documentation files (the "Software"), to deal with
-the Software without restriction, including without limitation the rights to
-use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
-of the Software, and to permit persons to whom the Software is furnished to do
-so, subject to the following conditions:
-
-    * Redistributions of source code must retain the above copyright notice,
-      this list of conditions and the following disclaimers.
-
-    * Redistributions in binary form must reproduce the above copyright notice,
-      this list of conditions and the following disclaimers in the
-      documentation and/or other materials provided with the distribution.
-
-    * Neither the names of the LLVM Team, University of Illinois at
-      Urbana-Champaign, nor the names of its contributors may be used to
-      endorse or promote products derived from this Software without specific
-      prior written permission.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
-FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
-CONTRIBUTORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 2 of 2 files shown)


tgross35:
One thing I am not certain about here; anyone who contributed to this repository between the 2019-01-19 LLVM relicense and now did so with no mention of `Apache-2.0 WITH LLVM-exception` being present in this repository. Do we need to get their acknowledgement to license this crate for use under `Apache-2.0 WITH LLVM-exception AND MIT`?

I think it could be sufficient that any work derived from an `Apache-2.0 WITH LLVM-exception` source is implicitly made available under that same license, even though there is no mention in this repository, but I am not sure of these technicalities.

tgross35:
> IANAL, but I don't think it works like that. Even with copyleft licenses, any modifications aren't automatically licensed under the same license (with copyleft, not licensing the modifications compatibly just means that you're violating the original license, not that your modifications are automatically correctly licensed).

Yeah, I think you are right. I am hoping for some sort of sound way to correct the licensing without needing to get an okay from all contributors since LLVM's relicense (~150) but I'm not sure if that is feasible.

There is some more discussion on the Zulip link asked in the top post if you want to drop in, but this is one question I haven't yet heard back about.

> That said, there's no requirement for changes to `Apache-2.0 WITH LLVM-exception` code to be licensed identically, so AFAICT any modifications to the LLVM `Apache-2.0 WITH LLVM-exception` code would be licensed the same as any other original contribution to `compiler-builtins`.

I don't think I follow here - what is the `Apache-2.0 WITH LLVM-exception` code? AIUI there is none in this repository since it was never part of the license here, even though it should all be under that license.

joshtriplett:
@tgross35 The license updates look correct to me. We also need to make sure we preserve any relevant copyright notices here.

tgross35:
I don't know of any copyrights that are relevant here. Searching for "copyright" in the LLVM directory only comes up with a few results we don't use https://github.com/search?q=repo%3Allvm%2Fllvm-project+path%3A%2F%5Ecompiler-rt%5C%2F%2F+copyright&type=code and the current license literally says "Copyright [yyyy] [name of copyright owner]"

LICENSE.txt
```diff
@@ -1,91 +1,270 @@
...
+license:
+
+* compiler-builtins is derived from LLVM's compiler-rt (https://llvm.org/).
+  Work derived form compiler-rt prior to 2019-01-19 is usable under the MIT
+  license, with the following copyright:
+
+      Copyright (c) 2009-2016 by the contributors listed in CREDITS.TXT
+
+  The relevant CREDITS.TXT is located at
+  https://github.com/llvm/llvm-project/blob/main/compiler-rt/CREDITS.TXT.
```
tgross35, line 261: @joshtriplett but I did include this to cover the copyright in the old license

joshtriplett: While I don't think we have a legal requirement to, we *might* want to consider reproducing `CREDITS.TXT` (inlined into `LICENSE.txt`). It's short.

The one caveat there being that if we start including it, we should update our included version if the upstream version changes and we copy any new code.

Cargo.toml
```diff
@@ -2,7 +2,7 @@
 authors = ["Jorge Aparicio <japaricious@gmail.com>"]
 name = "compiler_builtins"
 version = "0.1.135"
-license = "MIT/Apache-2.0"
+license = "Apache-2.0 WITH LLVM-exception AND MIT AND (MIT OR Apache-2.0)"
```
joshtriplett, line 5:
```suggestion
license = "MIT AND Apache-2.0 WITH LLVM-exception AND (MIT OR Apache-2.0)"
```

(Minor consistency/readability nit, not a blocker.)

LICENSE.txt
```diff
@@ -1,91 +1,270 @@
-==============================================================================
-compiler-builtins License
-==============================================================================
+compiler-builtins as a whole is available for use under both the MIT license
+and the Apache-2.0 license with the LLVM exception (Apache-2.0 WITH
+LLVM-exception AND MIT).
```
joshtriplett, line 3:
```suggestion
compiler-builtins as a whole is available for use under both the MIT license
and the Apache-2.0 license with the LLVM exception (`MIT AND Apache-2.0 WITH
LLVM-exception`).
```

Minor consistency and readability nit. Not a blocker.

joshtriplett (approved):
Posted a couple of minor nits, but they're not blockers.

LICENSE.txt
tgross35: I did consider this but it didn't seem like it was ever really used on LLVM's side. The latest update is from two years before `compiler-builtins` existed, and it had only ever been updated twice before that.

Cargo.toml
tgross35: Agreed, made this change

LICENSE.txt
tgross35: Updated

RalfJung:
> MIT AND Apache-2.0

This should be "MIT OR Aapache-2.0", right? You were listing the 3 AND clauses, but the third thing you list does not match the third AND clause.

LICENSE.txt
```diff
@@ -1,91 +1,270 @@
...
+and the Apache-2.0 license with the LLVM exception (MIT AND Apache-2.0 WITH
+LLVM-exception).
 
-The compiler-builtins crate is dual licensed under both the University of
-Illinois "BSD-Like" license and the MIT license.  As a user of this code you may
-choose to use it under either license.  As a contributor, you agree to allow
-your code to be used under both.
+As a contributor, you agree that your code can be used under either the MIT
+license, or the Apache-2.0 license, or the Apache-2.0 license with the LLVM
+exception.
```
RalfJung, line 7: To be clear, this means I agree that you can choose any of these licenses, right? It doesn't mean I as the contributor can choose one of these licenses to make the contributor under? IMO this is not clear enough.

joshtriplett:
@RalfJung Fixed.

PR #722, opened 2024-11-01
Disable `f16` for LoongArch64

heiher:
Disable `f161` for LoongArch64 due to incorrect code generation on LLVM 19, which causes failures in `testcrate/tests/conv.rs`. This workaround will remain in place until llvm/llvm-project#109093 is merged or we upgrade to LLVM 20.

### DIFF
--- a/configure.rs
+++ b/configure.rs
@@ -64,7 +64,7 @@ pub fn configure_f16_f128(target: &Target) {
         "arm64ec" => (false, false),
         // `f16` crashes <https://github.com/llvm/llvm-project/issues/50374>
         "s390x" => (false, true),
-        // `f16` test failures
+        // FIXME(llvm): `f16` test failures fixed by <https://github.com/llvm/llvm-project/pull/107791>
         "loongarch64" => (false, true),
         // `f128` crashes <https://github.com/llvm/llvm-project/issues/96432>
         "mips64" | "mips64r6" => (true, false),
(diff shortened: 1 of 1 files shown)


tgross35:
Do tests pass if you only disable FFI calls here https://github.com/rust-lang/compiler-builtins/blob/d4a39557885f7b57c3b98c661bb000d587165bc9/testcrate/build.rs#L55-L66 instead of disabling everything for the type?

heiher:
> Do tests pass if you only disable FFI calls here
> 
> https://github.com/rust-lang/compiler-builtins/blob/d4a39557885f7b57c3b98c661bb000d587165bc9/testcrate/build.rs#L55-L66
> instead of disabling everything for the type?

No, they are still failing.

configure.rs
```diff
@@ -64,6 +64,8 @@ pub fn configure_f16_f128(target: &Target) {
         "arm64ec" => (false, false),
         // `f16` crashes <https://github.com/llvm/llvm-project/issues/50374>
         "s390x" => (false, true),
+        // `f16` test failures
```
tgross35, line 67:
```suggestion
        // FIXME(llvm): `f16` test failures fixed by <https://github.com/llvm/llvm-project/pull/107791>
```

tgross35 (approved):
Thanks for checking, I updated the comment but this lgtm

tgross35:
If you need it in rust-lang/rust, this is released in 0.1.138 https://github.com/rust-lang/compiler-builtins/releases/tag/compiler_builtins-v0.1.138. Will probably be a few days before we can bump to that version since a few builtins updates are already queued.

PR #723, opened 2024-11-01
ci: add support for loongarch64-unknown-linux-gnu

heiher:
On top of #722 and #724


tgross35 (approved):
LGTM but needs a rebase

PR #724, opened 2024-11-01
Use `f16_enabled`/`f128_enabled` in `examples/intrinsics.rs`

heiher:


examples/intrinsics.rs
```diff
@@ -191,7 +197,10 @@ mod intrinsics {
...
     /* f128 operations */
 
-    #[cfg(not(any(target_arch = "powerpc", target_arch = "powerpc64")))]
+    #[cfg(all(
+        f16_enabled,
+        not(any(target_arch = "powerpc", target_arch = "powerpc64"))
+    ))]
     pub fn trunctfhf(x: f128) -> f16 {
         x as f16
     }
```
tgross35, line 206: This needs `f128_enabled` too

tgross35: Ah, there isn't existing gating for `f128`. It can be added to all the f128 functions if you are up for it, otherwise this is fine once CI passes.

tgross35 (approved):
Thanks!

PR #735, opened 2024-12-15
Use a C-safe return type for `__rust_[ui]128_*` overflowing intrinsics

tgross35:
Most of our Rust-specific overflowing intrinsics currently return `(i128, bool)`, which is not guaranteed to have a stable ABI. Switch to returning the overflow via a mutable parameter and only directly returning the integer result.

`__rust_i128_mulo` now matches the function signature of `__muloti4`, but they do not share the same ABI on Windows so we cannot easily deduplicate them.

### DIFF
--- a/src/int/addsub.rs
+++ b/src/int/addsub.rs
@@ -66,31 +66,39 @@ intrinsics! {
         AddSub::add(a,b)
     }
 
-    pub extern "C" fn __rust_i128_addo(a: i128, b: i128) -> (i128, bool) {
-        a.addo(b)
+    pub extern "C" fn __rust_i128_addo(a: i128, b: i128, oflow: &mut i32) -> i128 {
+        let (add, o) = a.addo(b);
+        *oflow = o.into();
+        add
     }
 
     pub extern "C" fn __rust_u128_add(a: u128, b: u128) -> u128 {
         AddSub::add(a,b)
     }
 
-    pub extern "C" fn __rust_u128_addo(a: u128, b: u128) -> (u128, bool) {
-        a.addo(b)
+    pub extern "C" fn __rust_u128_addo(a: u128, b: u128, oflow: &mut i32) -> u128 {
+        let (add, o) = a.addo(b);
+        *oflow = o.into();
+        add
     }
 
     pub extern "C" fn __rust_i128_sub(a: i128, b: i128) -> i128 {
         AddSub::sub(a,b)
     }
 
-    pub extern "C" fn __rust_i128_subo(a: i128, b: i128) -> (i128, bool) {
-        a.subo(b)
+    pub extern "C" fn __rust_i128_subo(a: i128, b: i128, oflow: &mut i32) -> i128 {
+        let (sub, o) = a.subo(b);
+        *oflow = o.into();
+        sub
     }
 
     pub extern "C" fn __rust_u128_sub(a: u128, b: u128) -> u128 {
         AddSub::sub(a,b)
     }
 
-    pub extern "C" fn __rust_u128_subo(a: u128, b: u128) -> (u128, bool) {
-        a.subo(b)
+    pub extern "C" fn __rust_u128_subo(a: u128, b: u128, oflow: &mut i32) -> u128 {
+        let (sub, o) = a.subo(b);
+        *oflow = o.into();
+        sub
     }
 }
--- a/testcrate/tests/addsub.rs
+++ b/testcrate/tests/addsub.rs
@@ -44,20 +44,22 @@ mod int_addsub {
                     use compiler_builtins::int::addsub::{$fn_add, $fn_sub};
 
                     fuzz_2(N, |x: $i, y: $i| {
-                        let add0 = x.overflowing_add(y);
-                        let sub0 = x.overflowing_sub(y);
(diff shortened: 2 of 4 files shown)


src/int/addsub.rs
```diff
@@ -66,31 +66,39 @@ intrinsics! {
         AddSub::add(a,b)
     }
 
-    pub extern "C" fn __rust_i128_addo(a: i128, b: i128) -> (i128, bool) {
-        a.addo(b)
+    pub extern "C" fn __rust_i128_addo(a: i128, b: i128, oflow: &mut i32) -> i128 {
```
bjorn3, line 69: Would you mind keeping the existing functions for a bit and adding new functions with the new ABI? That way the cg_clif update and the compiler-builtins update don't have to be done at the exact same time.

tgross35: I can do so if it helps, but why would this be needed? We pin the version of `compiler_builtins` with `=` now so it shouldn't get out of sync, and I won't merge this until the r-l/rust PR gets approved.

bjorn3: > We pin the version of compiler_builtins with = now so it shouldn't get out of sync, and I won't merge this until the r-l/rust PR gets approved.

In that case I guess it is fine to keep as-is.

src/int/addsub.rs
```diff
@@ -66,31 +66,39 @@ intrinsics! {
...
     pub extern "C" fn __rust_u128_sub(a: u128, b: u128) -> u128 {
         AddSub::sub(a,b)
     }
 
-    pub extern "C" fn __rust_u128_subo(a: u128, b: u128) -> (u128, bool) {
-        a.subo(b)
+    pub extern "C" fn __rust_u128_subo(a: u128, b: u128, oflow: &mut i32) -> u128 {
+        let (sub, o) = a.subo(b);
+        *oflow = o.into();
+        sub
```
bjorn3, line 102: cg_clif doesn't actually use the add/sub intrinsics, so I opened https://github.com/rust-lang/compiler-builtins/pull/745 to remove them.

testcrate/tests/addsub.rs
```diff
@@ -44,20 +44,22 @@ mod int_addsub {
                     use compiler_builtins::int::addsub::{$fn_add, $fn_sub};
 
                     fuzz_2(N, |x: $i, y: $i| {
-                        let add0 = x.overflowing_add(y);
-                        let sub0 = x.overflowing_sub(y);
-                        let add1: ($i, bool) = $fn_add(x, y);
-                        let sub1: ($i, bool) = $fn_sub(x, y);
-                        if add0.0 != add1.0 || add0.1 != add1.1 {
+                        let (add0, add_o0)= x.overflowing_add(y);
+                        let (sub0, sub_o0)= x.overflowing_sub(y);
```
bjorn3, line 48:
```suggestion
                        let (add0, add_o0) = x.overflowing_add(y);
                        let (sub0, sub_o0) = x.overflowing_sub(y);
```

testcrate/tests/addsub.rs
```diff
@@ -44,20 +44,22 @@ mod int_addsub {
...
+                        let mut add_o1 = 0;
+                        let mut sub_o1 = 0;
+                        let add1: $i = $fn_add(x, y, &mut add_o1);
+                        let sub1: $i = $fn_sub(x, y, &mut sub_o1);
+                        if add0 != add1 || i32::from(add_o0) != add_o1 {
                             panic!(
                                 "{}({}, {}): std: {:?}, builtins: {:?}",
-                                stringify!($fn_add), x, y, add0, add1
+                                stringify!($fn_add), x, y, (add0, add_o0) , (add1, add_o1)
```
bjorn3, line 56:
```suggestion
                                stringify!($fn_add), x, y, (add0, add_o0), (add1, add_o1)
```

testcrate/tests/addsub.rs
```diff
@@ -44,20 +44,22 @@ mod int_addsub {
...
                             );
                         }
-                        if sub0.0 != sub1.0 || sub0.1 != sub1.1 {
+                        if sub0 != sub1 || i32::from(sub_o0) != sub_o1 {
                             panic!(
                                 "{}({}, {}): std: {:?}, builtins: {:?}",
-                                stringify!($fn_sub), x, y, sub0, sub1
+                                stringify!($fn_sub), x, y, (sub0, sub_o0) , (sub1, sub_o1)
```
bjorn3, line 62:
```suggestion
                                stringify!($fn_sub), x, y, (sub0, sub_o0), (sub1, sub_o1)
```

bjorn3 (commented):
Rustfmt doesn't work inside macros.

testcrate/tests/addsub.rs
tgross35: Thanks. Considering a large portion of this crate's tests lives in macros, I may eventually pull in my rustfmt hack from https://github.com/rust-lang/libc/blob/1f68d578c430a9eaed30ad04e29d7e70ef7ab263/ci/style.sh#L24-L63.

bjorn3 (approved):

tgross35:
Thanks for the review. I won't merge this until bumping builtins is unblocked, and the rust-lang/rust PR is ready to go.

PR #736, opened 2024-12-22
Fix a bug in `abs_diff`

tgross35:
These implementations of `abs_diff` were added in c2ff1b3119 ("Completely overhaul fuzz testing"), but the signed implementation is wrong when |x| + |y| exceeds the integer's limits (e.g. `(-128).abs_diff(1)` should be 128 but currently these return 127.

Resolve this by just using `std`'s implementation since that is stable now. This isn't used anywhere critical, we probably just weren't hitting the edge case.

### DIFF
--- a/src/int/mod.rs
+++ b/src/int/mod.rs
@@ -240,11 +240,7 @@ macro_rules! int_impl {
             }
 
             fn abs_diff(self, other: Self) -> Self {
-                if self < other {
-                    other.wrapping_sub(self)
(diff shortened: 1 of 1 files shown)


tgross35:
Ah yeah that's a typo, how did you stumble across this and find that :). Updated the description.

RalfJung:
I saw the PR listed in TWiR and the title made me curious, then read the description and got confused. ;)

Issue #742, opened 2025-01-07
Build fails when `opt-level = "z"`


greenhat:
When `opt-level = "z"` is set in the `profile` in Cargo.toml project that is built with `build-std` option, the `compiler-builtins` crate compilation fails with:
```
thread 'main' panicked at /Users/dzadorozhnyi/.cargo/registry/src/index.crates.io-6f17d22bba15001f/compiler_builtins-0.1.141/configure.rs:35:63:
  called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }
```

This happens during the attempt to parse "z" as `u8` introduced in the commit:

https://github.com/rust-lang/compiler-builtins/commit/4359549eb8ee01d7b8298a0a3257a1fae7d02501#diff-0af32baaac2f5d27b5b314d8025adbffa02ac553981887261f12cd85b2fa08aeR35

Pinging @tgross35 

EDIT: "z" along with "s" are listed as valid values at https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level

tgross35:
Thanks for catching this, I’ll add a fix soon.

tgross35:
This should be resolved by https://github.com/rust-lang/compiler-builtins/pull/743.

PR #743, opened 2025-01-07
Account for optimization levels other than numbers

tgross35:
The build script currently panics with `opt-level=z` or `opt-level=s`. Account for this here.

This is the `compiler-builtins` version of [1]. This PR also includes the `libm` submodule update with [1], and a few other small changes.

Fixes: https://github.com/rust-lang/compiler-builtins/issues/742

[1]: https://github.com/rust-lang/libm/pull/417

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -121,7 +121,7 @@ fn configure_libm(target: &Target) {
     }
 
     println!("cargo:rustc-check-cfg=cfg(optimizations_enabled)");
-    if target.opt_level >= 2 {
+    if !matches!(target.opt_level.as_str(), "0" | "1") {
         println!("cargo:rustc-cfg=optimizations_enabled");
     }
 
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -121,7 +121,7 @@ fn configure_libm(target: &Target) {
     }
 
     println!("cargo:rustc-check-cfg=cfg(optimizations_enabled)");
-    if target.opt_level >= 2 {
+    if !matches!(target.opt_level.as_str(), "0" | "1") {
```
faptc, line 124: is "z" optimizations_enabled?

tgross35: For the purposes of this it is fine, the config is only used in `libm`'s test-related code. I only included it here for consistency if we wind up doing something similar in `compiler-builtins/testcrate`.

PR #745, opened 2025-01-09
Remove some unused rust specific intrinsics

bjorn3:
I think they were originally introduced when 128bit int support was added to rustc but before LLVM consistently supported 128bit int operations across platforms. Nowadays LLVM and GCC handle 128bit int operations for us rather than manually adding calls to rust specific intrinsics. The only exception is the Cranelift backend which still uses the __rust_u/i128_mulo intrinsics and as such this PR keeps those around.


tgross35:
From a quick look it seems like gc_gcc may still use these https://github.com/rust-lang/rust/blob/824759493246ee383beb9cd5ceffa0e15deb9fa4/compiler/rustc_codegen_gcc/src/int.rs#L180-L183 https://github.com/rust-lang/rust/blob/824759493246ee383beb9cd5ceffa0e15deb9fa4/compiler/rustc_codegen_gcc/src/intrinsic/mod.rs#L988, but maybe `is_native*` now return true so those codepaths don't get hit anymore?

Cc @rust-lang/wg-gcc-backend

antoyo:
GCC only supports the type `__int128` on some platforms as far as I know. But it might be possible to use the equivalent of C23's `_BitInt` now and stop using those intrinsics.

tgross35:
> GCC only supports the type `__int128` on some platforms as far as I know. But it might be possible to use the equivalent of C23's `_BitInt` now and stop using those intrinsics.

As in, using `_BitInt` only when `__int128` is not available? That sounds reasonable. Unfortunately you probably can't use `_BitInt(128)` all the time, since it is not ABI-compatible with `__int128` on x86 (which IMO is a terrible design decision).

I guess in either case we should probably keep these around at least for a little while.

antoyo:
> As in, using `_BitInt` only when `__int128` is not available?

Yes.

> Unfortunately you probably can't use `_BitInt(128)` all the time, since it is not ABI-compatible with `__int128` on x86 (which IMO is a terrible design decision).

Thanks for the info. I took note of that [in the issue](https://github.com/rust-lang/rustc_codegen_gcc/issues/514).

PR #747, opened 2025-01-15
Eliminate the use of `public_test_dep!`

tgross35:
Replace `public_test_dep!` by placing optionally public items into new modules, then controlling what is exported with the `public-test-deps` feature.

This is nicer for automatic formatting and diagnostics.

This is a reland of 2e2a9255 ("Eliminate the use of `public_test_dep!`"), which was reverted in 47e50fd2 ('Revert "Eliminate the use of..."') due to a bug exposed at [1]. This was fixed in [2], so the cleanup should be able to be applied again.

[1]: https://github.com/rust-lang/rust/pull/128691
[2]: https://github.com/rust-lang/rust/pull/135278


nikic:
Could this be reverted until the issues with this change are resolved?

tgross35:
Let me see if I can get it fixed today since the same issue is blocking future progress with `libm` too. If not then yeah, I will have to revert again in order to get the uefi fix into r-l/rust.

tgross35:
https://github.com/rust-lang/rust/pull/135501 should be ready to go, which unblocks this. Assuming that can merge in the near future, I will get https://github.com/rust-lang/compiler-builtins/pull/752 right after.

PR #752, opened 2025-01-28
Don't build out of line atomics support code for uefi

nikic:
https://github.com/llvm/llvm-project/pull/116706 added Windows support to cpu_model. Compiling for UEFI also goes through that code path, because we treat it as a windows target. However, including windows.h is not actually going to work (and the used API would not be available in an UEFI environment).

Disable building of cpu_model on UEFI to fix this.


nikic:
See https://github.com/rust-lang/rust/pull/135763#issuecomment-2616208547 for the build failure this causes with LLVM 20.

tgross35 (approved):

PR #753, opened 2025-01-31
Expose `erf{,c}{,f}` from `libm`

GrigorenkoPV:
For https://github.com/rust-lang/rust/pull/136324


tgross35 (approved):
This should be fine but I'm not going to merge it yet. The update to `compiler-builtins` in r-l/rust is more or less blocked, I'm going to wait for that to flush before adding any more functionality to `compiler-builtins` on the off chance it causes errors.

tgross35:
Wound up reverting the blocking PR so this is unstuck

PR #755, opened 2025-02-04
Revert "Eliminate the use of `public_test_dep!`"

tgross35:
[1] has not gone forward, so this needs to be reverted again in order to unblock a compiler-builtins upgrade that is necessary for the LLVM 20 upgrade.

This reverts commit a2494f14e99ae90c964f12bf0c059d63ccc07c2a.

[1]: https://github.com/rust-lang/rust/pull/135501


tgross35:
Cc @nikic

Issue #758, opened 2025-02-15
x86-64: Some `extern "C"` functions use `U64x2` on a softfloat target, which makes no sense


RalfJung:
Passing `U64x2` across an `extern "C"` boundary on x86-64  requires SSE; it is hence not possible to pass those types when SSE is disabled, which is the case for softfloat targets. (See https://github.com/rust-lang/rust/issues/116558 for more context on how ABI and SIMD target features interact.) This comes up, for instance, when building for `x86_64-unknown-uefi`.

Here are the places where a SIMD type is passed via `extern "C"`:
```
warning: this function definition uses a SIMD vector type that (with the chosen ABI) requires the `sse` target feature, which is not enabled
   --> src/macros.rs:260:13
    |
260 | /             extern $abi fn $name( $($argname: $ty),* )
261 | |                 -> $crate::macros::win64_128bit_abi_hack::U64x2
    | |_______________________________________________________________^ function defined here
    |
   ::: src/float/conv.rs:395:1
    |
395 | / intrinsics! {
396 | |     #[arm_aeabi_alias = __aeabi_f2uiz]
397 | |     pub extern "C" fn __fixunssfsi(f: f32) -> u32 {
398 | |         float_to_unsigned_int(f)
...   |
443 | | }
    | |_- in this macro invocation
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #116558 <https://github.com/rust-lang/rust/issues/116558>
    = help: consider enabling it globally (`-C target-feature=+sse`) or locally (`#[target_feature(enable="sse")]`)
    = note: `#[warn(abi_unsupported_vector_types)]` on by default
    = note: this warning originates in the macro `intrinsics` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: this function definition uses a SIMD vector type that (with the chosen ABI) requires the `sse` target feature, which is not enabled
   --> src/macros.rs:260:13
    |
260 | /             extern $abi fn $name( $($argname: $ty),* )
261 | |                 -> $crate::macros::win64_128bit_abi_hack::U64x2
    | |_______________________________________________________________^ function defined here
    |
   ::: src/float/conv.rs:395:1
    |
395 | / intrinsics! {
396 | |     #[arm_aeabi_alias = __aeabi_f2uiz]
397 | |     pub extern "C" fn __fixunssfsi(f: f32) -> u32 {
[...]

PR #759, opened 2025-02-18
remove win64_128bit_abi_hack

RalfJung:
For msvc and GNU targets, that hack is no longer needed -- rustc by itself now returns i128/u128 via xmm0 on these targets. That got implemented in https://github.com/rust-lang/rust/pull/134290.

For the x86_64-unknown-uefi target, the situation still seems a bit unclear, but using a `repr(simd)` is definitely wrong for that target as SIMD is not available. So I can't imagine that this PR would make anything any worse...

Fixes https://github.com/rust-lang/compiler-builtins/issues/758

### DIFF
--- a/src/float/conv.rs
+++ b/src/float/conv.rs
@@ -403,7 +403,6 @@ intrinsics! {
         float_to_unsigned_int(f)
     }
 
-    #[win64_128bit_abi_hack]
     pub extern "C" fn __fixunssfti(f: f32) -> u128 {
         float_to_unsigned_int(f)
     }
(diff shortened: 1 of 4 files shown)


RalfJung:
That aarch64 failure can't be related to this PR, I think.

The windows failure is suspicious but it's on win32 and this patch only affects 64bit systems so I don't think it can be that, either?

tgross35 (approved):
The change lgtm, but I just want to be certain of the uefi situation before merging. Also CI issues, I'll try to get those fixed https://github.com/rust-lang/compiler-builtins/pull/760.

tgross35:
r-l/rust update https://github.com/rust-lang/rust/pull/137297

PR #770, opened 2025-02-26
Enable `f16` for LoongArch

heiher:
r? @tgross35

Tracking issue for f16: https://github.com/rust-lang/rust/issues/116909


tgross35:
Since loongarch doesn't have the system version to test against, could you add a line here so it uses the apfloat fallback https://github.com/rust-lang/compiler-builtins/blob/453da663e447fe80001733417a89c45bb624d1f4/testcrate/build.rs#L59?

heiher:
> Since loongarch doesn't have the system version to test against, could you add a line here so it uses the apfloat fallback
> 
> https://github.com/rust-lang/compiler-builtins/blob/453da663e447fe80001733417a89c45bb624d1f4/testcrate/build.rs#L59
> ?

Thanks for your suggestion.

tgross35:
This should be good but I'm holding off since our compiler-builtins updates are a bit backed up.

testcrate/tests/conv.rs
```diff
@@ -306,7 +306,11 @@ mod extend {
     }
 
     #[cfg(all(f16_enabled, f128_enabled))]
-    #[cfg(not(any(target_arch = "powerpc", target_arch = "powerpc64")))]
+    #[cfg(not(any(
+        target_arch = "powerpc",
+        target_arch = "powerpc64",
+        target_arch = "loongarch64"
+    )))]
```
tgross35, line 313: From the earlier failures, it looked like loongarch doesn't have/need __gnu_h2f_ieee and __gnu_f2h_ieee. I added a way to disable only these symbols, could you try adding loongarch here? https://github.com/rust-lang/compiler-builtins/blob/b7b93103fb9293c0c502dc1ae34e2ad5c871bc39/testcrate/build.rs#L95-L97

tgross35:
I guess it is looking for those unmangled convert symbols to implement the other routines rather than only testing our `__gnu_(h2f|f2h)_ieee`. So you were right before, disabling the tests in testcrate/tests/conv.rs is probably the best thing to do until this makes it to a nightly rustc, then we can enable them again. Sorry about that.

heiher:
> I guess it is looking for those unmangled convert symbols to implement the other routines rather than only testing our `__gnu_(h2f|f2h)_ieee`. So you were right before, disabling the tests in testcrate/tests/conv.rs is probably the best thing to do until this makes it to a nightly rustc, then we can enable them again. Sorry about that.

Okay. I'll revert to the previous version, and I also sincerely appreciate your efforts in this.

tgross35 (approved):

tgross35:
Happy to help get the type working on more targets 👍 thanks for all the changes.

I have to hold off merging until https://github.com/rust-lang/rust/pull/137634 goes through but LGTM.

PR #776, opened 2025-03-02
Remove outdated information from the readme

tgross35:


tgross35:
@Amanieu we no longer have any reason to encourage using compiler-builtins as a direct dependency, correct?

Amanieu:
That's correct. In the long term we should probably stop publishing it on crates.io at all.

PR #778, opened 2025-03-04
Replace `bl!` with `sym` in arm trampolines

tgross35:
`bl!` exists to account for the leading `_` in Apple symbols. Switch to using `sym` which does this implicitly.


tgross35:
Actually this probably doesn't work, we don't want to use our paths when the C shims are available.

Issue #788, opened 2025-03-12
atomic_load_unordered and atomic_store_unordered should not be used


RalfJung:
The atomic_load_unordered and atomic_store_unordered intrinsics generate LLVM IR that cannot be mapped to the Rust AM (specifically, they cannot be represented in the C++ concurrency memory model), and as such there's no well-defined way to use them in a Rust program. Ideally we'd remove those intrinsics from the compiler. That is currently blocked on the issue that this crate uses those intrinsics.

Most of those uses seem to be to provide various intrinsics like `__llvm_memcpy_element_unordered_atomic_1`. Since Rust cannot use unordered atomics, those intrinsics should not be needed, right? Why do they even exist?

The other uses are in `src/arm_linux.rs` and `src/mem/impls.rs`. The latter code is already problematic for other reasons (https://github.com/rust-lang/compiler-builtins/issues/559). For both, I wonder if we could just use "relaxed" accesses instead, so that at least those are accesses that make sense in our memory model?

RalfJung:
Oh interesting, why are those `unordered`? That does not match https://github.com/rust-lang/rfcs/pull/3301 nor the C++ proposal, to my knowledge. The bytewise atomic memcpy uses the same orderings as other atomic accesses.

tgross35:
I believe this has been resolved with https://github.com/rust-lang/compiler-builtins/pull/790.

PR #789, opened 2025-03-12
remove element_unordered_atomic intrinsics

RalfJung:
It is not clear how these "unordered" operations are supposed to fit the memory model. The intrinsics used here are labeled with "DO NOT USE" in the docs. The plan in https://github.com/rust-lang/rfcs/pull/3301 and for C++, as I understand it, is to have "bytewise atomic memcpy" with the usual orderings: SeqCst, Release, Acquire, Relaxed. These "unordered" intrinsics seem to be a dead end.

These intrinsics are also unused as far as I can tell; https://github.com/rust-lang/rust/pull/59155 which planned to use them never landed.

Part of https://github.com/rust-lang/compiler-builtins/issues/788.


tgross35 (approved):
SGTM. It seems like these are emitted from https://llvm.org/docs/LangRef.html#element-wise-atomic-memory-intrinsics, but if we aren't using that and C++ also has no way to represent these operations, doesn't seem likely that anybody can actually be expecting them to be available.

Issue #559, opened 2023-12-01
Several functions perform out-of-bounds memory accesses (which is UB)


RalfJung:
I found this code which openly admits to doing OOB accesses:

https://github.com/rust-lang/compiler-builtins/blob/2a67ad74b77dd5c7dbd77a27156176136b0b606d/src/mem/impls.rs#L64-L70

This code also looks suspicious, e.g. if this is used to implement 2-byte atomic accesses then there is OOB here:

https://github.com/rust-lang/compiler-builtins/blob/56172fcd8bd045e38bbdf76697d1fcca1e965d6d/src/arm_linux.rs#L57-L89

I haven't done a thorough audit so there might be more. I found these by grepping for `atomic_load_unordered`.

RalfJung:
OOB accesses using Rust accesses (including atomic and volatile) are always UB.

The only way to do OOB accesses is using inline assembly. Then the inline asm block has to ensure that the access can never cause a program abort (so it should be on the same page as an address that we know we could access from Rust).

RalfJung:
If I understand https://github.com/rust-lang/rust/pull/113923 correctly, compiler-builtins can now participate in LTO, so there's no way to make an argument like "this crate is part of the privileged runtime so it can violate some of the usual rules of UB that we apply to all other code". This *is* a soundness issue.

bjorn3:
AFAIK the calls to compiler-builtins intrinsics are inserted somewhere in the backend well after optimizations. In fact that PR needed to explicitly mark these functions as used to prevent LLVM from optimizing all these intrinsics away as dead code.

RalfJung:
That sounds plausible for calls to the LLVM builtin (making that a libcall happens pretty late), but if the user directly calls the C function `memcpy`, why would that not be inlined?

bjorn3:
It is possible that LLVM will replace explicit memcpy calls with the llvm.memcpy intrinsic. I don't know if it is in practice guaranteed to do this and if it is LLVM itself that does this, or clang when compiling C code though.

RalfJung:
There now is a ["masked load" intrinsic](https://doc.rust-lang.org/nightly/std/intrinsics/simd/fn.simd_masked_load.html) that can be used to do a large load that is partially OOB. Would that help to fix the OOB in this code without regressing performance?

Lokathor:
I think it would take enough time to set up the mask that you could likely just read the initial byte chunk in a loop and have it be just as fast.

[...]

PR #790, opened 2025-03-12
Remove use of `atomic_load_unordered` and undefined behaviour from `arm_linux.rs`

beetrees:
`atomic_load_unordered` is in the process of being removed (#788), so replace it with a relaxed atomic load. Additionally, avoid doing an out-of-bounds Rust read (#559) on 1 and 2 byte atomics by using inline ASM.

The only changes to the resulting optimised assembly on the latest Rust nightly (`rustc 1.87.0-nightly (665025243 2025-03-11)`) are a few minor changes to instruction scheduling in the 1 and 2 byte atomic RMW functions.

cc @RalfJung - I've tried to summarise the approach of using inline ASM in the `SAFETY` comments, but would definitely appreciate someone checking they make sense.

### DIFF
--- a/compiler-builtins/src/arm_linux.rs
+++ b/compiler-builtins/src/arm_linux.rs
@@ -1,5 +1,6 @@
-use core::intrinsics;
+use core::arch;
 use core::mem;
+use core::sync::atomic::{AtomicU32, Ordering};
 
 // Kernel-provided user-mode helper functions:
 // https://www.kernel.org/doc/Documentation/arm/kernel_user_helpers.txt
@@ -7,6 +8,7 @@ unsafe fn __kuser_cmpxchg(oldval: u32, newval: u32, ptr: *mut u32) -> bool {
     let f: extern "C" fn(u32, u32, *mut u32) -> u32 = mem::transmute(0xffff0fc0usize as *const ());
     f(oldval, newval, ptr) == 0
 }
+
 unsafe fn __kuser_memory_barrier() {
     let f: extern "C" fn() = mem::transmute(0xffff0fa0usize as *const ());
     f();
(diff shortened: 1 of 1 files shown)


src/arm_linux.rs
```diff
@@ -54,13 +56,43 @@ fn insert_aligned(aligned: u32, val: u32, shift: u32, mask: u32) -> u32 {
     (aligned & !(mask << shift)) | ((val & mask) << shift)
 }
 
+/// Loads the value at `ptr`.
+///
+/// # Safety
+///
+/// `ptr` must be aligned and point to memory within a page that allows read access.
+/// If `T` has a size of 4, then `ptr` must be valid for an atomic read.
```
RalfJung, line 64:
```suggestion
/// If `T` has a size of 4, then `ptr` must be valid for a relaxed atomic read.
```
However, I am a bit confused about what extra requirements you have in mind here.

src/arm_linux.rs
```diff
@@ -54,13 +56,43 @@ fn insert_aligned(aligned: u32, val: u32, shift: u32, mask: u32) -> u32 {
...
+unsafe fn load_aligned<T>(ptr: *mut u32) -> u32 {
+    if mem::size_of::<T>() == 4 {
+        // SAFETY: As `T` has a size of 4, the caller garantees this is sound.
+        unsafe { AtomicU32::from_ptr(ptr).load(Ordering::Relaxed) }
+    } else {
+        // SAFETY:
+        // As all 4 bytes pointed to by `ptr` might not be dereferenceable due to the provenance of
+        // `ptr` when doing atomic operations on a `u8`/`i8`/`u16`/`i16`, inline ASM is used to
```
RalfJung, line 72: It's not just about provenance. it might be genuinely outside the bounds of the allocation.

src/arm_linux.rs
```diff
@@ -54,13 +56,43 @@ fn insert_aligned(aligned: u32, val: u32, shift: u32, mask: u32) -> u32 {
...
+}
+
 // Generic atomic read-modify-write operation
 unsafe fn atomic_rmw<T, F: Fn(u32) -> u32, G: Fn(u32, u32) -> u32>(ptr: *mut T, f: F, g: G) -> u32 {
     let aligned_ptr = align_ptr(ptr);
     let (shift, mask) = get_shift_mask(ptr);
 
     loop {
-        let curval_aligned = intrinsics::atomic_load_unordered(aligned_ptr);
+        let curval_aligned = load_aligned::<T>(aligned_ptr);
```
RalfJung, line 95: Since this is doing an atomic load, that should probably be in the name of the function.

RalfJung (commented):
I can't comment on the inline asm attributes, but the rest LGTM.

`__kuser_cmpxchg` is a bit questionable since `oldval` and `newval` cannot hold provenance. It would be more correct to use `MaybeUninit<u32>`.

src/arm_linux.rs
RalfJung: That sounds like an implementation detail of the function.

I'd expect the safety contract to be something like
- `ptr` must be 4-aligned (currently it just says "aligned" but it is ambiguous how aligned)
- `T` must have a size of at most 4
- `ptr` must be valid for a relaxed atomic read of size `size_of::<T>()`

Whether and when it uses atomics vs inline asm does not have to be exposed in the docs.

RalfJung:
> The 4 byte functions like __sync_fetch_and_add_4 are used to implement the strict_provenance_atomic_ptr functions like AtomicPtr::fetch_byte_add, however they (the __sync_*_4 functions) are currently implemented using operations on u32s. AFAICT they also need to be changed to use <*mut _>::map_addr or equivalent in order to preserve provenance?

Ah, I had not thought of that part yet, I was more concerned of the data that is outside the value we conceptually operate on -- there might be pointers there we have to preserve.

tgross35 (approved):
The changes here LGTM with some of the minor docs updates Ralf mentioned at https://github.com/rust-lang/compiler-builtins/pull/790#discussion_r1995374131, and possibly a mention somewhere that the read is atomic without ordering. I double checked that `ldr` is all LLVM emits anyway in this case https://rust.godbolt.org/z/cerYKhKd8.

RalfJung:
"Without ordering" sounds like a non-atomic read. Is that enough?

tgross35:
> "Without ordering" sounds like a non-atomic read. Is that enough?

At least on this target, probably, but isn't this an AM vs. hardware difference? E.g. as I understand it LLVM could decide to turn a non-atomic `u32` read as four `u8` reads, which would be problematic.

Amanieu (approved):

tgross35:
Unfortunately it looks like this is hitting the monomorphization error for Android https://github.com/rust-lang/rust/pull/138771#issuecomment-2743581681. I guess this file may not be covered in CI at all, given it is only used for very old arm https://github.com/rust-lang/compiler-builtins/blob/e49ff02ce147653497d18087814a2e0360cccf5a/compiler-builtins/build.rs#L75-L84

RalfJung:
I thought calling generic functions is fine since they get inlined anyway? What does "cannot call functions through upstream monomorphizations" mean?

tgross35:
I don't really understand what the exact restrictions are, @saethlin would know better. But in this case `AtomicX::from_ptr` surprisingly isn't marked `#[inline]` and isn't actually generic (despite the error message), I think inlining should fix it (patch WIP).

Manually inlining the `.cast()` here also seems to fix things.

RalfJung:
Ah ofc the atomic types are all monomorphic.

So I understand now why it errors, but still can't make sense of the error message.

tgross35:
It comes from an internal detail https://github.com/rust-lang/rust/blob/a4a11aca5ecf24dfff3c00715641026809951305/compiler/rustc_codegen_ssa/src/base.rs#L851. But yeah, I'm not sure what the significance of "upstream monomorphization" is instead of "non-inlinable call" or whatever.

saethlin:
> But yeah, I'm not sure what the significance of "upstream monomorphization" is instead of "non-inlinable call" or whatever.

Because whether a call is non-inlinable is not strictly the same as whether it is to an upstream monomorphization. Calls to upstream monomorphizations can still be inlined or otherwise optimized away by the codegen backend, and the point of this diagnostic is that we _don't_ take that into consideration.

So we could change the error to say "compiler-builtins can only make calls to functions defined in compiler-builtins or functions which are defined in other crates but lowered from MIR as part of compiling compiler-builtins" but that's just including the definition of "upstream monomorphization" in the diagnostic :shrug:

RalfJung:
> So we could change the error to say "compiler-builtins can only make calls to functions defined in compiler-builtins or functions which are defined in other crates but lowered from MIR as part of compiling compiler-builtins" but that's just including the definition of "upstream monomorphization" in the diagnostic 🤷

That would still be useful, given that hardly anyone knows what an "upstream monomorphization" is.^^ I did non-trivial work on the monomorphization logic, and I never encountered that term, and did not understand it when encountering it here.

Issue #711, opened 2024-10-14
Native support for AVR instrinsics


Patryk27:
At the moment compiler-builtins isn't useful for AVR, because that platform uses [a custom calling convention for intrinsics](https://gcc.gnu.org/wiki/avr-gcc#Exceptions_to_the_Calling_Convention) - supporting AVR here would require:
1. Implementing this custom calling convention in LLVM.
2. Exposing it in rustc.
3. Finally, using it within compiler-builtins.

There are no ongoing plans to do so yet, but over https://github.com/rust-lang/rust/pull/131651 it was pointed out it would be nice to have some kind of message of this issue left, so here we go 🙂

tgross35:
Thanks for filing this. I don't think we are blocked on LLVM having support for the calling convention, instead we should be able to use naked functions to translate the cc like we do for aeabi intrinsics https://github.com/rust-lang/compiler-builtins/blob/cb060052ab7e4bad408c85d44be7e60096e93e38/src/arm.rs#L21-L35.

Patryk27:
Oh, sure, that seems like a good idea 🙂

tgross35:
Separately you may be interested in getting AVR asm to stable. This isn't a blocker for this crate but it may be potentially useful in user code https://github.com/rust-lang/rust/issues/93335

tgross35:
Also, do you know if there is a version of these symbols in assembly that don't come from libgcc? Some of these are probably easy enough that we could provide our own assembly implementations, but we need to be careful with licensing.

Patryk27:
> Also, do you know if there is a version of these symbols in assembly that don't come from libgcc?

I'm aware only of the libgcc implementation, unfortunately.

Patryk27:
Status: was a little bit busy previous week, I'll take a look at those intrinsics and your example in a couple of days 🙂

Patryk27:
You're right - it seems that the only Funny Intrinsics™  used at the moment are `__udivmodqi4`, `__divmodqi4`, `__udivmodhi4`, and `__divmodhi4`. 

I have tried to summon `__umulhisi3` with:
```rust
#[inline(never)]
pub fn fun(a: &u16, b: &u16, x: &mut u16, y: &mut u16) {
    (*x, *y) = a.widening_mul(*b);
}
``` 
... but even this causes LLVM to go with 16b->32b extension followed by 32b x 32b multiplication (aka `__mulsi3`). 

[...]

PR #791, opened 2025-03-16
avr: Skip No More!

Patryk27:
This pull request gets rid of all of the `#[avr_skip]`s introduced in the past.

> My ~heart~ uC skips, skips, skips, skips, skips, skips ~a beat~ an intrinsic

... Olly Murs, frustrated about having to pull someone's intrinsics when working on AVR firmwares (colorized).

Changes here are related to https://github.com/rust-lang/compiler-builtins/issues/711, but note that we continue _not_ to provide `__u?divmod[hq]i4` - it shouldn't be terribly difficult to implement those (I even have a work-in-progress branch in rustc that exposes the proper calling convention), it's just a different kind of change that will be done in another pull request.

In general, the changes boil down to:

- removing overzealous `#[avr_skip]` on intrinsics that don't actually have special convention (such as `__addsf3`),
- implementing proper support for the other intrinsics (such as `__unordsf2` or `__divmodsi4`).

I've tested new code via [avr-tester](https://github.com/Patryk27/avr-tester/pull/6) - in particular, there's a test that [uses randomized math operations](https://github.com/Patryk27/avr-tester/blob/89c1820b50da9e41d6a2e9cb2896192321a3b544/avr-tester/tests/acc/eval.rs#L1) and another one that [checks floating point math](https://github.com/Patryk27/avr-tester/blob/89c1820b50da9e41d6a2e9cb2896192321a3b544/avr-tester/tests/acc/fractal.rs#L1), and both pass using the implementations provided by compiler-builtins.

Related discussions:

- https://github.com/Rahix/avr-hal/discussions/641
- https://github.com/rust-lang/compiler-builtins/issues/711

### DIFF
--- a/compiler-builtins/src/float/add.rs
+++ b/compiler-builtins/src/float/add.rs
@@ -189,14 +189,12 @@ where
 }
 
 intrinsics! {
-    #[avr_skip]
     #[aapcs_on_arm]
     #[arm_aeabi_alias = __aeabi_fadd]
     pub extern "C" fn __addsf3(a: f32, b: f32) -> f32 {
         add(a, b)
     }
 
-    #[avr_skip]
     #[aapcs_on_arm]
     #[arm_aeabi_alias = __aeabi_dadd]
     pub extern "C" fn __adddf3(a: f64, b: f64) -> f64 {
--- a/compiler-builtins/src/float/cmp.rs
+++ b/compiler-builtins/src/float/cmp.rs
@@ -3,6 +3,14 @@
 use crate::float::Float;
 use crate::int::MinInt;
 
+// https://github.com/llvm/llvm-project/blob/1e6ba3cd2fe96be00b6ed6ba28b3d9f9271d784d/compiler-rt/lib/builtins/fp_compare_impl.inc#L22
+#[cfg(target_arch = "avr")]
+pub type CmpResult = i8;
+
+// https://github.com/llvm/llvm-project/blob/1e6ba3cd2fe96be00b6ed6ba28b3d9f9271d784d/compiler-rt/lib/builtins/fp_compare_impl.inc#L25
+#[cfg(not(target_arch = "avr"))]
+pub type CmpResult = i32;
+
 #[derive(Clone, Copy)]
 enum Result {
     Less,
@@ -12,7 +20,7 @@ enum Result {
 }
 
 impl Result {
-    fn to_le_abi(self) -> i32 {
+    fn to_le_abi(self) -> CmpResult {
         match self {
             Result::Less => -1,
             Result::Equal => 0,
@@ -21,7 +29,7 @@ impl Result {
         }
     }
 
-    fn to_ge_abi(self) -> i32 {
+    fn to_ge_abi(self) -> CmpResult {
         match self {
             Result::Less => -1,
             Result::Equal => 0,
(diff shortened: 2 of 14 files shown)


src/float/cmp.rs
```diff
@@ -3,6 +3,12 @@
 use crate::float::Float;
 use crate::int::MinInt;
 
+#[cfg(target_arch = "avr")]
+type CmpResult = i8;
+
+#[cfg(not(target_arch = "avr"))]
+type CmpResult = i32;
```
tgross35, line 10: In https://github.com/Rahix/avr-hal/discussions/641#discussioncomment-12508954 you mentioned that this should be i16, but here it is i8. If `i16` is correct, could these just use `core::ffi::c_int`?

src/int/sdiv.rs
```diff
@@ -114,13 +114,50 @@ macro_rules! smod {
...
 );
+
+#[cfg(target_arch = "avr")]
+intrinsics! {
+    /// Returns `(a / b, a % b)`.
+    ///
+    /// Note that we can't actually use `-> (i32, i32)`, because it forces a
+    /// stack allocation instead of going through R18:R26, but packing things up
+    /// using bit operations is good enough.
+    pub extern "C" fn __divmodsi4(a: i32, b: i32) -> u64 {
```
tgross35, line 133: (nit) can all of the new functions go in the existing `intrinsics!` blocks, with `#[cfg(target_arch = "avr")]` or `#[cfg(not(target_arch = "avr"))]` on the individual functions? I think that should work the same.

src/macros.rs
```diff
@@ -256,6 +256,9 @@ macro_rules! intrinsics {
 
         #[cfg(all(target_vendor = "apple", any(target_arch = "x86", target_arch = "x86_64"), not(feature = "mangled-names")))]
         mod $name {
+            #[allow(unused_imports)]
+            use super::*;
```
tgross35, line 260: Why are all the new `use super::*` imports required?

src/int/sdiv.rs
tgross35: Also, regarding the doc comment here, we shouldn't actually use `(i32, i32)` since that is ABI-unstable. Maybe returning a `#[repr(packed)] struct i32i32 { lo: i32, hi: i32 }` works? If not then `u64` is fine but the doc comment should probably be updated.

tgross35:
Also did you mean for this PR to have draft in the title, but not actually be marked as a draft? :)

Patryk27:
> Also did you mean for this PR to have draft in the title, but not actually be marked as a draft? :)

Hah, #justgitlabthings

src/float/cmp.rs
Patryk27: Unfortunately it's not `c_int` - that's how LLVM defines it:

https://github.com/llvm/llvm-project/blob/1e6ba3cd2fe96be00b6ed6ba28b3d9f9271d784d/compiler-rt/lib/builtins/fp_compare_impl.inc#L22

src/int/sdiv.rs
Patryk27: Ad 1: `#[cfg(target_arch = ...)]` inside the `intrinsics!` macro doesn't work, because this macro (sometimes?) generates a separate `mod` block for each function, expanding this:

```rust
intrinsics! {
    #[cfg(target_arch = "foo")]
    fn sth(...) { ... }

    #[cfg(target_arch = "bar")]
    fn sth(...) { ... }
}
```

... into something akin to:

```rust
mod sth {
    #[cfg(target_arch = "foo")]
    fn sth(...) { ... }
}

mod sth {
    #[cfg(target_arch = "bar")]
    fn sth(...) { ... }
}
```

... which, naturally, ends up making the computer angry:

image

I think maybe the macro could be extended to support pulling the `#[cfg]` attribute specifically in front of the module, but my initial attempts have failed.

src/macros.rs
Patryk27: So that the `CmpResult` is in scope (see the discussion above on `intrinsics!` generating a new module for each function).

Patryk27:
Ok, it's ready!

I've tested the code on ATmega328P and ATtiny167 (just simavr, since I've recently relocated and my electrical stuff is still in boxes, waiting for final delivery 😭).

src/float/cmp.rs
tgross35: Thanks, looks like GCC has the same https://github.com/gcc-mirror/gcc/blob/16065b6239aab2eecfd7a50f58d38324ee6478ba/libgcc/config/avr/avr-lib.h#L25C17-L25C23.

src/macros.rs
tgross35: Thanks, that makes more sense. I think I would prefer `use crate::float::cmp::CmpResult` here, or `-> crate::float::cmp::CmpResult` in `float::cmp` since a glob import in the macro sounds like a recipe for future confusion. It would also be fine to reexport `CmpResult` back up to the top level too so that gets shortened to `crate::CmpResult`.

tgross35:
Sorry about the conflicts, this will need a rebase since I moved some things around (git should be able to figure it out). Just that and https://github.com/rust-lang/compiler-builtins/pull/791#discussion_r2002053800, then I think this will be ready to go.

Patryk27:
Okie, ready; I also added a comment to the `CmpResult` alias that links to LLVM, so that it's clear where the types come from.

tgross35 (approved):
Thanks!

PR #792, opened 2025-03-17
Re-enable memset tests on powerpc64(le)

Gelbpunkt:
Originally disabled in https://github.com/rust-lang/compiler-builtins/pull/481, these tests pass fine for me on powerpc64le-unknown-linux-gnu with LLVM 20.1 and `rustc 1.87.0-nightly (4d30011f6 2025-03-15)`.

Fixes: https://github.com/rust-lang/rust/issues/99853

cc @RalfJung @Amanieu @apiraino

Full test run (successful): https://gist.github.com/Gelbpunkt/b178b3838b32154e877799e2b6395546

### DIFF
--- a/testcrate/tests/mem.rs
+++ b/testcrate/tests/mem.rs
@@ -230,8 +230,6 @@ fn memmove_backward_aligned() {
     }
 }
 
-// PowerPC tests are failing: https://github.com/rust-lang/rust/issues/99853
-#[cfg(not(target_arch = "powerpc64"))]
 #[test]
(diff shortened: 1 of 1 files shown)


RalfJung:
> See https://github.com/rust-lang/rust/issues/99853

This should probably say "Fixes:" ?

Gelbpunkt:
> > See [rust-lang/rust#99853](https://github.com/rust-lang/rust/issues/99853)
> 
> This should probably say "Fixes:" ?

You're right, added that.

tgross35 (approved):
Great, thanks!

PR #794, opened 2025-03-18
Implement `__isPlatformVersionAtLeast` and `__isOSVersionAtLeast`

madsmtm:
## Motivation

When Objective-C code uses `@available(...)`, Clang inserts a call to [`__isPlatformVersionAtLeast`](https://github.com/llvm/llvm-project/blob/llvmorg-20.1.0/compiler-rt/lib/builtins/os_version_check.c#L276) (`__isOSVersionAtLeast` in older Clang versions). These symbols not being available in `compiler-builtins` sometimes ends up causing linker errors.

The workaround is to link `libclang_rt.osx.a`, see e.g. https://github.com/alexcrichton/curl-rust/issues/279. But that's very difficult for users to figure out (and the backreferences to that issue indicates that people are still running into this in their own projects every so often).

For a recent example, this is preventing `rustc` from using LLVM assertions on macOS, see https://github.com/rust-lang/rust/pull/62592#issuecomment-510670657 and https://github.com/rust-lang/rust/pull/134275#issuecomment-2543067830.

Apart from linker errors above, it is also a blocker for [setting the correct minimum OS version in `cc-rs`](https://github.com/rust-lang/rust/issues/136113), which is a huge correctness footgun: By default, if using e.g. `__builtin_available(macos 10.15, *)`, the symbol usually _happens_ to be left out, since `clang` defaults to compiling for the host macOS version, and thus things _seem_ to work. But there's a reason why people write availability checks, which is that otherwise the code won't work on the older OS! But if we want to fix this in `cc-rs`, we might end up introducing linker errors in places where they weren't before.

<details>
<summary>Example crate where this behaviour is illustrated</summary>

```toml
# Cargo.toml
[package]
name = "foo"
version = "0.0.0"
edition = "2024"

[build-dependencies]
cc = "1.0"
```

```c
// foo.c
int foo(void) {
    // Use some API that's a lot newer than the host machine
    if (__builtin_available(macos 20.0, *)) {
        return 1;
    } else {
        return 0;
    }
}
```

```rust
// build.rs
fn main() {
    println!("cargo:rerun-if-changed=foo.c");
    cc::Build::new().file("foo.c").compile("foo");
}
```

```rust
// main.rs
unsafe extern "C" {
    safe fn foo() -> core::ffi::c_int;
}

fn main() {
    println!("{}", foo());
}
```

</details>

(Finally, I'll reveal that my super secret evil agenda is to expose some variant of `@available`/`__builtin_available` in Rust's `std`, and that'll probably be easier if the bulk of the implementation is already here in `compiler-builtins`. But I believe adding this here has value regardless of those future plans).

## Solution

Implement `__isPlatformVersionAtLeast` and `__isOSVersionAtLeast`, see the code comments for more details. Of particular note:

I've choosen to stray a bit from LLVM's upstream implementation, and not use `_availability_version_check` since [it has problems when compiling with an older SDK](https://github.com/llvm/llvm-project/issues/64227). Instead, we use `sysctlbyname` when available to still avoid the costly PList lookup in most cases, but still with a fall back to the PList lookup when that is not available (this fallback is similar to LLVM's implementation).

I've also preferred to panic when hitting an unexpected state, instead of silently returning a wrong value, but am unsure what your policies are around this?

## Testing

Tested using the equivalent of `cargo test --manifest-path testcrate/Cargo.toml --test os_version_check` on the following configurations:
- macOS 14.7.3 on a Macbook Pro M2
    - `aarch64-apple-darwin`
    - `x86_64-apple-darwin` (under Rosetta)
    - `aarch64-apple-ios-macabi`
    - `x86_64-apple-ios-macabi` (under Rosetta)
    - `aarch64-apple-ios` (using Xcode's "Designed for iPad" setting)
    - `aarch64-apple-ios-sim` (in iOS Simulator, as iPhone with iOS 17.5)
    - `aarch64-apple-ios-sim` (in iOS Simulator, as iPad with iOS 18.2)
    - `aarch64-apple-tvos-sim` (in tvOS Simulator)
    - `aarch64-apple-watchos-sim` (in watchOS Simulator)
    - `aarch64-apple-ios-sim` (in visionOS simulator, using Xcode's "Designed for iPad" setting)
    - `aarch64-apple-visionos-sim` (in visionOS Simulator)
- macOS 15.3.1 VM
    - `aarch64-apple-darwin`
    - `aarch64-apple-ios-macabi`
- macOS 10.12.6 on an Intel Macbook from 2013
    - `x86_64-apple-darwin`
    - `i686-apple-darwin`
    - `x86_64-apple-ios` (in iOS Simulator)
- iOS 9.3.6 on a 1st generation iPad Mini
    - `armv7-apple-ios` with an older compiler

Along with manually inspecting the output of `version_from_sysctl()` and `version_from_plist()`, and verifying that they actually match what's expected.

I believe the only real omissions here would be:
- `aarch64-apple-ios` on a newer iPhone that has `sysctl` available (iOS 11.4 or above).
- `aarch64-apple-ios` on a Vision Pro using Xcode's "Designed for iPad" setting.

But I don't have the hardware available to test those.

### DIFF
--- a/src/os_version_check/darwin_impl.rs
+++ b/src/os_version_check/darwin_impl.rs
@@ -1,0 +1,590 @@
+use core::ffi::CStr;
+use core::{
+    ffi::{c_char, c_int, c_long, c_uint, c_void},
+    num::NonZero,
+    ptr::null_mut,
+    slice,
+    sync::atomic::{AtomicU32, Ordering},
+};
+
+/// Get the current OS version.
+#[inline]
+pub(super) fn current_version() -> u32 {
+    // Cache the lookup for performance.
+    //
+    // 0.0.0 is never gonna be a valid version, so we use that as our sentinel value.
+    static CURRENT_VERSION: AtomicU32 = AtomicU32::new(0);
+
+    // We use relaxed atomics, it doesn't matter if multiple threads end up racing to read or write
+    // the version, `lookup_version` should be idempotent and always return the same value.
+    //
+    // `compiler-rt` uses `dispatch_once`, but that's overkill for the reasons above.
+    let version = CURRENT_VERSION.load(Ordering::Relaxed);
+    if version == 0 {
+        let version = lookup_version().get();
+        CURRENT_VERSION.store(version, Ordering::Relaxed);
+        version
+    } else {
+        version
+    }
+}
+
+#[cold]
+// Use `extern "C"` to abort on panic, allowing `current_version` to be free of panic handling.
+pub(super) extern "C" fn lookup_version() -> NonZero<OSVersion> {
+    // Since macOS 10.15, libSystem has provided the undocumented `_availability_version_check` via
+    // `libxpc` (zippered, so requires platform parameter to differentiate between on macOS and Mac
+    // Catalyst) for doing the version lookup, though it's usage may be a bit dangerous, see:
+    // - https://reviews.llvm.org/D150397
+    // - https://github.com/llvm/llvm-project/issues/64227
+    //
+    // So instead, we use the safer approach of reading from `sysctl` (which is faster), and if that
+    // fails, we fall back to the property list (this is what `_availability_version_check` does
+    // internally).
+    let version = version_from_sysctl().unwrap_or_else(version_from_plist);
+
+    // Use `NonZero` to try to make it clearer to the optimizer that this will never return 0.
+    NonZero::new(version).expect("version cannot be 0.0.0")
+}
+
+/// Look up the current OS version(s) from `/System/Library/CoreServices/SystemVersion.plist`.
+///
+/// More specifically, from the `ProductVersion` and `iOSSupportVersion` keys, and from
+/// `$IPHONE_SIMULATOR_ROOT/System/Library/CoreServices/SystemVersion.plist` on the simulator.
+///
+/// This file was introduced in macOS 10.3, which is well below the minimum supported version by
+/// `rustc`, which is currently macOS 10.12.
+///
+/// # Panics
+///
+/// Panics if reading or parsing the version fails (or if the system was out of memory).
+///
+/// We deliberately choose to panic, as having this silently return an invalid OS version would be
+/// impossible for a user to debug.
+#[allow(non_upper_case_globals, non_snake_case)]
+pub(super) fn version_from_plist() -> OSVersion {
+    #[allow(clippy::upper_case_acronyms)]
+    enum FILE {}
+
+    const SEEK_END: c_int = 2;
+
+    const RTLD_LAZY: c_int = 0x1;
+    const RTLD_LOCAL: c_int = 0x4;
+
+    // SAFETY: Same signatures as in `libc`.
+    //
+    // NOTE: We do not need to link these; that will be done by `std` by linking `libSystem`
+    // (which is required on macOS/Darwin).
+    unsafe extern "C" {
+        unsafe fn getenv(s: *const c_char) -> *mut c_char;
+        safe fn malloc(size: usize) -> *mut c_void;
+        unsafe fn free(p: *mut c_void);
+        unsafe fn strcpy(dst: *mut c_char, src: *const c_char) -> *mut c_char;
+        unsafe fn strcat(s: *mut c_char, ct: *const c_char) -> *mut c_char;
+
+        unsafe fn fopen(filename: *const c_char, mode: *const c_char) -> *mut FILE;
+        unsafe fn fseek(stream: *mut FILE, offset: c_long, whence: c_int) -> c_int;
+        unsafe fn ftell(stream: *mut FILE) -> c_long;
+        unsafe fn rewind(stream: *mut FILE);
+        unsafe fn fread(ptr: *mut c_void, size: usize, nobj: usize, stream: *mut FILE) -> usize;
+        unsafe fn fclose(file: *mut FILE) -> c_int;
+
+        unsafe fn dlopen(filename: *const c_char, flag: c_int) -> *mut c_void;
+        unsafe fn dlsym(handle: *mut c_void, symbol: *const c_char) -> *mut c_void;
+        // NOTE: Cannot use this because we cannot Debug print `CStr` in `compiler-builtins`.
+        // safe fn dlerror() -> *mut c_char;
+        unsafe fn dlclose(handle: *mut c_void) -> c_int;
+    }
+
+    // We do not need to do a similar thing as what Zig does to handle the fake 10.16 versions
+    // returned when the SDK version of the binary is less than 11.0:
+    // <https://github.com/ziglang/zig/blob/0.13.0/lib/std/zig/system/darwin/macos.zig>
+    //
+    // <https://github.com/apple-oss-distributions/xnu/blob/xnu-11215.81.4/libsyscall/wrappers/system-version-compat.c>
+    //
+    // The reasoning is that we _want_ to follow Apple's behaviour here, and return 10.16 when
+    // compiled with an older SDK; the user should upgrade their tooling.
+    //
+    // NOTE: `rustc` currently doesn't set the right SDK version when linking with ld64, so this
+    // will have the wrong behaviour with `-Clinker=ld` on x86_64. But that's a `rustc` bug:
+    // <https://github.com/rust-lang/rust/issues/129432>
+
+    struct Deferred<F: FnMut()>(F);
+    impl<F: FnMut()> Drop for Deferred<F> {
+        fn drop(&mut self) {
+            (self.0)();
+        }
+    }
+
+    let path = c"/System/Library/CoreServices/SystemVersion.plist";
+    let _path_free;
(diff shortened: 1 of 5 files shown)


testcrate/tests/os_version_check.rs
```diff
@@ -0,0 +1,153 @@
...
+    assert_eq!(__isOSVersionAtLeast(major, minor, subminor + 1), 0);
+    assert_eq!(__isOSVersionAtLeast(major, minor + 1, subminor), 0);
+    assert_eq!(__isOSVersionAtLeast(major + 1, minor, subminor), 0);
+}
+
+// Test internals
+
+#[path = "../../src/os_version_check/darwin_impl.rs"]
+#[allow(dead_code)]
+mod darwin_impl;
```
madsmtm, line 60: A bit unsure about this, should I make it a unit-test inside `src/os_version_check/darwin_impl.rs` instead of an integration test?

madsmtm:
I don't have perms to do:
@rustbot ping apple

So gonna ping people manually:
CC @hkratz, @inflation, @shepmaster, @thomcc, @nvzqz, @madsmtm, @BlackHoleFox

inflation:
Slightly off-topic, I’ve also encountered issues when library need symbols from `clang_rt` on macOS, e.g. x86 feature detection from the underlying C library. Would there be a better way to let the downstream knows we need `clang_rt` linked?

src/os_version_check/darwin_impl.rs
```diff
@@ -0,0 +1,590 @@
...
+            Some(val) => match val.checked_add(digit as _) {
+                Some(val) => ret = val,
+                None => panic!("version is too large"),
+            },
+            None => panic!("version is too large"),
+        };
+    }
+
+    (ret, bytes)
+}
```
madsmtm, line 590: Unfamiliar with `compiler-builtins`, so I can't tell if there's a better way of doing this kind of parsing?

madsmtm: Things are helped by this being specific to Apple platforms, so we already depend on `libSystem.dylib`, and could use for example `sscanf`. What do you prefer?

madsmtm: Actually, I think I'd prefer that myself, it's also what LLVM does (the manual parsing here was a remnant of [my implementation in `objc2`](https://github.com/madsmtm/objc2/blob/53a0cb58649a9ca1951d39544bef1cbaa7210d90/crates/objc2/src/__macro_helpers/os_version/apple.rs)). Will wait 'till I hear back, then I'll probably implement that.

src/os_version_check/darwin_impl.rs
```diff
@@ -0,0 +1,590 @@
...
+        CURRENT_VERSION.store(version, Ordering::Relaxed);
+        version
+    } else {
+        version
+    }
+}
+
+#[cold]
+// Use `extern "C"` to abort on panic, allowing `current_version` to be free of panic handling.
+pub(super) extern "C" fn lookup_version() -> NonZero<OSVersion> {
```
bjorn3, line 34: Why is this extern C?

src/os_version_check/darwin_impl.rs
```diff
@@ -0,0 +1,590 @@
...
+///
+/// This file was introduced in macOS 10.3, which is well below the minimum supported version by
+/// `rustc`, which is currently macOS 10.12.
+///
+/// # Panics
+///
+/// Panics if reading or parsing the version fails (or if the system was out of memory).
+///
+/// We deliberately choose to panic, as having this silently return an invalid OS version would be
+/// impossible for a user to debug.
```
bjorn3, line 63: compiler-builtins may not panic as that requires libcore but compiler-builtins can't depend on any other crates as it is last on the liker cli. Instead you will have to abort.

src/os_version_check/darwin_impl.rs
```diff
@@ -0,0 +1,590 @@
...
+
+    const SEEK_END: c_int = 2;
+
+    const RTLD_LAZY: c_int = 0x1;
+    const RTLD_LOCAL: c_int = 0x4;
+
+    // SAFETY: Same signatures as in `libc`.
+    //
+    // NOTE: We do not need to link these; that will be done by `std` by linking `libSystem`
+    // (which is required on macOS/Darwin).
```
bjorn3, line 77: What if the user is building a no_std macOS application?

tgross35, line 77: Could you double check that building a `#![no_std]` crate with this version of `compiler_builtins` links without errors? We won't have the system libraries in that case so it seems like this would fail. There aren't any other cases where we call `libc` functions in this crate.

tgross35: Ah, Bjorn just said the same thing.

testcrate/tests/os_version_check.rs
tgross35: Unit tests unfortunately don't really work in `compiler-builtins`. It is fine to just make everything public so you don't need to do the module path trick.

src/os_version_check/darwin_impl.rs
tgross35: I'll take a closer look at the parsing in a bit, but it is probably better not to depend on `sscanf` for the reason mentioned in https://github.com/rust-lang/compiler-builtins/pull/794#discussion_r2000295614.

src/os_version_check/darwin_impl.rs
madsmtm: > as it is last on the liker cli

Pretty sure Apple's linker doesn't require ordering in the same way as traditional ELF linkers?

I'll have to test it to be sure though, I'll see if I can build a custom toolchain with this branch of `compiler-builtins`.

src/os_version_check/darwin_impl.rs
madsmtm: Wait... I think I may be mistaken here.

Seems like using `MACOSX_DEPLOYMENT_TARGET=12.0` doesn't require the symbol! So you're probably right that we'd need to add a `#[link(name = "System")]`.

madsmtm:
> Could it make sense to provide this as a weak symbol in `std` for now, and then consider exposing that directly later?

I'm not familiar enough with the linking constraints that `compiler-builtins` are under to be sure? Would making it a weak symbol enough to not conflict with the symbol in `libclang_rt.dylib`?

To be clear, I'd be fine with requiring `std` to not get the linker errors I discussed above, I only proposed it here to mirror `compiler-rt`. Is the only reason to have something in `compiler-builtins` to allow `#![no_core]` to use it too?

> if doing things correctly requires env / filesystem / allocation

It seems crazy to me too, but yes, it does require all of these.

> For similar reasons I am assuming LLVM would also prefer these to be `libc` or `libgcc` functions rather than compiler-rt, but doesn't have control over those things on Apple platforms to add them.

Ah, yeah, thinking about it, I suspect they added it to `compiler-rt` such that it could work even on older OSes (since `libc`/`libSystem` are shipped with the OS). But we don't have that same constraint with `std`.

Actually, does compiler built-ins in Clang that call `libc` functions exist? Nvmd, allocation. Yeah, in that case it's stupid to have this in `compiler-rt`.

> This might be worth bringing up on Zulip, and/or just submitting an ACP to add a version of `__builtin_available`.

I've opened [a Zulip thread](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/Provide.20.60__isPlatformVersionAtLeast.60.20in.20.60std.60.3F), will submit an ACP for a version of `__builtin_available` later.

src/os_version_check/darwin_impl.rs
bjorn3: Even if the linker accepts it, you run the risk of cyclic dependencies with the panic handler (as `__isPlatformVersionAtLeast` panics, which calls the panic handler, which in turn calls a function that calls `__isPlatformVersionAtLeast`).

tgross35:
> I'm not familiar enough with the linking constraints that `compiler-builtins` are under to be sure? Would making it a weak symbol enough to not conflict with the symbol in `libclang_rt.dylib`?

This has to be tested but I think it should. Amanieu mentioned before that static+weak symbols may get picked over dynamic+strong, so ours might get favored. But that doesn't seem like a problem.

> To be clear, I'd be fine with requiring `std` to not get the linker errors I discussed above, I only proposed it here to mirror `compiler-rt`. Is the only reason to have something in `compiler-builtins` to allow `#![no_core]` to use it too?

`compiler-builtins` doesn't actually get added by default for `#![no_core]` https://github.com/rust-lang/rust/blob/75530e9f72a1990ed2305e16fd51d02f47048f12/compiler/rustc_metadata/src/creader.rs#L1116-L1122. I think this crate exists mostly for organization, the compiler expects a handful of `extern "C"` functions to be available but we don't really have anything like that in `core`.

madsmtm:
I have opened https://github.com/rust-lang/rust/pull/138944 instead, will pursue that and close this for now, thanks for the input so far!

PR #797, opened 2025-03-18
Remove unneeded `--manifest-path` from CI scripts

tgross35:
Since 942ab9fc37 ("Move `examples/intrinsics.rs` to its own crate"), `testcrate` is a default workspace member so there is no need to specify it explicitly when testing or building. Use this to clean up CI scripts.


tgross35:
This doesn't work because the `compiler-builtins` feature is enabled by default.

PR #713, opened 2024-10-17
Use wrapping pointer arithmetic in mem/impls.rs

saethlin:
@Amanieu you suggested wrapping arithmetic *and* volatile loads/stores. Some quick experimentation indicates that volatile loads/stores do indeed have an optimization impact, but wrapping pointer arithmetic maybe doesn't: https://godbolt.org/z/Y8MP46dEE

Also, since these intrinsics (i.e. `memcpy`) are well-known to the compiler, I suspect that if a compiler is going to optimize out a read/write to the last byte in address space, it'll do that by analyzing the call site of the `memcpy` and not consider whether or not the implementation is full of volatiles. So I suspect using volatiles in here would be dubious benefit for making the Rust toolchain do what our users want to use it for, and probably inflict a slowdown on everyone using compiler-builtins with optimizations enabled.

PR #801, opened 2025-03-19
Deny `unsafe_op_in_unsafe_fn` by default

tgross35:
Edition 2024 requires that we avoid this. There is a lot of code that will need to be adjusted, so start the process here.

### DIFF
--- a/builtins-test-intrinsics/src/main.rs
+++ b/builtins-test-intrinsics/src/main.rs
@@ -649,14 +649,14 @@ fn something_with_a_dtor(f: &dyn Fn()) {
     f();
 }
 
-#[no_mangle]
+#[unsafe(no_mangle)]
 #[cfg(not(thumb))]
 fn main(_argc: core::ffi::c_int, _argv: *const *const u8) -> core::ffi::c_int {
     run();
     0
 }
 
-#[no_mangle]
+#[unsafe(no_mangle)]
 #[cfg(thumb)]
 pub fn _start() -> ! {
     run();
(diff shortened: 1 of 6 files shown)


src/arm.rs
```diff
@@ -23,143 +23,261 @@ intrinsics! {
...
-            src = src.offset(1);
+            // SAFETY: `dst` and `src` are both valid for at least 4 bytes, from
+            // `memcpy` preconditions and the loop guard.
+            unsafe { *dst = *src };
+
+            // TODO
+            unsafe {
+                dst = dst.offset(1);
+                src = src.offset(1);
+            }
```
tgross35, line 124: @RalfJung is using `offset` to create a pointer one past the end of the allocation allowed? This caught my eye when adding safety comments; the docs say "the entire memory range between self and the result must be in bounds of that allocated object" and Miri doesn't complain, but I had been thinking that it was at least indeterminate (and now I can't find the issue).

RalfJung: If an allocation has size 4, and a ptr starts at offset 0 and you add 4 to it, then the memory range between the old and new pointer has size 4. Clearly that is entirely inbounds. Not sure what you mean by "indeterminate" here -- we don't use that word in Rust, and in C it is used to refer to *values*, not *operations*, so I can't make sense of it here.

I think "one past the end" is bad terminology, since the pointer is not *past* the end, it is *at* the end -- it points after the last element.

tgross35: Yeah that's straightforward enough. I must be misremembering; I thought there was an open issue somewhere related to differences between Rust and C here regarding creating a pointer past (or at) the end of an allocation and how that related to `memcpy`, but maybe this was only about wrapping at the end of the address space.

> Not sure what you mean by "indeterminate" here -- we don't use that word in Rust

Indeterminate only because it had not yet been decided, not technical terminology.

tgross35: I'm probably thinking of https://github.com/rust-lang/unsafe-code-guidelines/issues/465 and related discussion, which has been decided and I'm just a year out of date :) So as I understand it, a Rust program can never access the last byte of the address space and that effectively means our `memcpy` doesn't have to worry about it - with the result that these offset calculations will never wrap.

RalfJung:
Yeah the last byte of the addr space does not have a pointer "after" it, making it special. At least clang has the same limitation as Rust here, I do not know about standard C.

tgross35:
Finally found what I was thinking of related to memcpy https://github.com/rust-lang/compiler-builtins/pull/713 and its context around [#t-compiler > Hello World on sparc-unknown-none-elf crashes @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Hello.20World.20on.20sparc-unknown-none-elf.20crashes/near/477271947).

PR #802, opened 2025-03-19
Don't call `rust_begin_unwind`

tgross35:


tgross35:
Now I see why this was done rather than calling `panic!()` in the first place, looks like panic handling needs a lot more symbols (`__CxxFrameHandler3`, `string.h`, etc)

PR #810, opened 2025-04-05
Reenable the use of the force-soft-floats feature

### LINKED ISSUE #649: Infinite recursion in sqrt
Hi.
This is a continuation of [this Zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/Infinite.20recursion.20in.20compiler-builtins) as it seems this might be a bug in compiler-builtins (please close if this is not the case).

The context was:

> I'm trying to update compiler-builtins in rustc_codegen_gcc that was pinned to 0.1.109 until the support for f16 and f128 was implemented.
> I used to not generate symbols like `sqrt` from `compiler_builtins` (instead, we were using the one from libgcc) in order to avoid infinite recursion like:
> ```
> compiler_builtins::math::libm::sqrt::sqrt
> compiler_builtins::math::sqrt
> sqrt
> core::core_arch::x86::sse2::_mm_sqrt_pd
> ```
> because rustc_codegen_gcc implements SIMD `sqrt` by calling `sqrt` on each element of the vector for now.
> But with the new update, it seems I cannot do that anymore. This workaround seemed like a hack anyway.

@Amanieu mentioned the following which might be the solution:
> Really compiler_builtins shouldn't be calling the native instruction, it should only contain the fallback.
> Since the compiler is responsible for emitting the proper instruction and not calling the builtin in the first place.

Thanks.

antoyo:
This fixes an infinite recursion in sqrt for codegen backends that implement intrinsics like simd_fsqrt by calling sqrt on every element of the vector.

Fix #649

This requires https://github.com/rust-lang/libm/pull/532 to fix the issue, so I'll need to update the submodule when this PR is merged.

### DIFF
--- a/compiler-builtins/build.rs
+++ b/compiler-builtins/build.rs
@@ -95,15 +95,19 @@ fn configure_libm(target: &Target) {
     println!("cargo:rustc-check-cfg=cfg(intrinsics_enabled)");
     println!("cargo:rustc-check-cfg=cfg(arch_enabled)");
     println!("cargo:rustc-check-cfg=cfg(optimizations_enabled)");
-    println!("cargo:rustc-check-cfg=cfg(feature, values(\"unstable-public-internals\"))");
+    println!("cargo:rustc-check-cfg=cfg(feature, values(\"unstable-public-internals\", \"force-soft-floats\"))");
 
     // Always use intrinsics
     println!("cargo:rustc-cfg=intrinsics_enabled");
 
+    // Codegen backends (e.g. rustc_codegen_gcc) that implement intrinsics like simd_fsqrt by
+    // calling sqrt on every element of the vector ends up with an infinite recursion without this
+    // feature because sqrt would call simd_fsqrt, which in turn calls sqrt on those codegen
+    // backends.
+    println!("cargo:rustc-cfg=feature=\"force-soft-floats\"");
+
     // The arch module may contain assembly.
-    if cfg!(feature = "no-asm") {
-        println!("cargo:rustc-cfg=feature=\"force-soft-floats\"");
-    } else {
+    if !cfg!(feature = "no-asm") {
         println!("cargo:rustc-cfg=arch_enabled");
     }
 


tgross35:
Sorry about the regression here. Could this be temporally fixed by setting the `compiler-builtins-no-asm` feature when building with cg_gcc https://github.com/rust-lang/rust/blob/d4f880f8ce832cd7560bb2f1ebc34f967055ffd7/library/std/Cargo.toml#L99? I think bootstrap has a way to configure this.

The change in `libm` was made to move away from subtractive features like `force-soft-float` because that doesn't work well with Cargo and config was getting complicated (now gone in `libm` but still exists in `compiler-builtins`). If something like this is needed, I'd prefer splitting out separate features to enable assembly and to enable the rest of `core::arch`, rather than adding back `force-soft-float`. I can do this work, just let me know if `no-asm` is an option.

antoyo:
> just let me know if `no-asm` is an option.

Just to make sure, this will disable some optimizations, right?
If so, I'd prefer to not use `no-asm` since `rustc_codegen_gcc` supports inline assembly.

tgross35:
Yeah you're right, that would affect `compiler-builtins`. It's not a problem in `libm` since there are no uses of `asm!` though.

So what might be best is to comment out `println!("cargo:rustc-cfg=arch_enabled");` with a `FIXME` to add this back in the future once `libm` drops all `core::arch` calls. This is the plan after the functions are in `core` (I'm hoping https://github.com/rust-lang/rust/pull/138087 can go to stable after only a couple of cycles).
 
The existing mention of `force-soft-floats` can be deleted too, it isn't used in either crate anymore so I guess I forgot to clean it up.

antoyo:
> So what might be best is to comment out `println!("cargo:rustc-cfg=arch_enabled");` with a `FIXME` to add this back in the future once `libm` drops all `core::arch` calls.

Ok, do you want me to make a PR for this or will you do it?
I might need more details if you want me to make the PR.

Thanks!

antoyo:
Closing since this was fixed elsewhere.

PR #812, opened 2025-04-07
Add `__chkstk` for `arm64ec-pc-windows-msvc`

codeOverFlow:
This is my first PR so I tried one that did not looked overly complicated.

This is base on  llvm/llvm-project@946a81f

As this is my first PR fill free to fire all mistake you can spot :)

I am also not sure for the `#[cfg]` gate I put.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ rely on CI.
 - [x] arm/aeabi_memset.S
 - [x] arm/aeabi_uidivmod.S
 - [x] arm/aeabi_uldivmod.S
-- [ ] arm/chkstk.S
+- [x] arm/chkstk.S
 - [ ] arm/divmodsi4.S (generic version is done)
 - [ ] arm/divsi3.S (generic version is done)
 - [ ] arm/modsi3.S (generic version is done)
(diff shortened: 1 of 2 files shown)


tgross35:
I’ll take a look at the implementation soon, do you have a repo that gets linker errors without this function?

codeOverFlow:
I did not know what to put :/ maybe I should just put a "cfg (windows)"  as
LLVM says it is windows only ?

I just tried to see which arm target is for Windows :/ and just found this
one

tgross35:
Could you clarify whether this is attempting to solve a problem you have experienced (e.g. linker error) or just adding an implementation that is missing?

We don't really want to add these implementations if Rust never emits them, but it's possible that is happening in circumstances we aren't yet testing. `thumbv7a-pc-windows-msvc` and `thumbv7a-uwp-windows-msvc` are the only Windows targets for 32-bit Arm and `arm64ec-pc-windows-msvc` is the only arm64ec target; I'm not sure that either of these need `__chkstk` or `__chkstk_arm64ec`, but maybe they do for `#![no_std]` binaries? Cc @bdbaia and @dpaoliello (listed target maintainers) in case you have any ideas.

If it's possible to create a linker error looking for these symbols then we can add them, but otherwise we could move it to the "Unimplemented functions" section.

PR #816, opened 2025-04-12
avr: implement unsigned division intrinsics

tones111:
This PR adds a missing AVR intrinsic for 8-bit unsigned division (see #711) and is based on the popular [long division](https://en.wikipedia.org/wiki/Divisio    n_algorithm#Long_division) algorithm.

GCC uses a [special calling convention](https://gcc.gnu.org/wiki/avr-gcc#Exceptions_to_the_Calling_Convention) for multiplication and division primitives, requiring a `naked_asm` implementation.

This implementation attempts to be an assembly port of the following Rust implementation of the long division algorithm.

```rust
pub fn long_div_u8(mut num_quo: u8, divisor: u8) -> (u8, u8) {
    let mut remainder = 0u8;
    for _ in 0..8u8 {
        let overflow;
        (num_quo, overflow) = num_quo.overflowing_add(num_quo);
        (remainder, _) = remainder.carrying_add(remainder, overflow);

        if let Some(rem) = remainder.checked_sub(divisor) {
            remainder = rem;
            num_quo += 1;
        }
    }
    (num_quo, remainder)
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_long_div_u8() {
        for n in 0..=u8::MAX {
            for d in 1..=u8::MAX {
                assert_eq!(long_div_u8(n, d), (n / d, n % d));
            }
        }
    }
```

### DIFF
--- a/compiler-builtins/src/int/udiv.rs
+++ b/compiler-builtins/src/int/udiv.rs
@@ -43,6 +43,92 @@ intrinsics! {
 
         ((rem as u64) << 32) | (div as u64)
     }
+
+    #[unsafe(naked)]
+    pub unsafe extern "C" fn __udivmodqi4() {
+        // compute unsigned 8-bit `n / d` and `n % d`.
+        //
+        // Note: GCC implements a [non-standard calling convention](https://gcc.gnu.org/wiki/avr-gcc#Exceptions_to_the_Calling_Convention) for this function.
+        // Inputs:
+        //     R24: dividend
+        //     R22: divisor
+        // Outputs:
+        //     R24: quotient  (dividend / divisor)
+        //     R25: remainder (dividend % divisor)
+        // Clobbers:
+        //     R23: loop counter
+        core::arch::naked_asm!(
+            // This assembly routine implements the [long division](https://en.wikipedia.org/wiki/Division_algorithm#Long_division) algorithm.
+            // Bits shift out of the dividend and into the quotient, so R24 is used for both.
+            "clr R25",      // remainder = 0
+
+            "ldi R23, 8",   // for each bit
+            "1:",
+            "lsl R24",      //     shift the dividend MSb
+            "rol R25",      //     into the remainder LSb
+
+            "cp  R25, R22", //     if remainder >= divisor
+            "brlo 2f",
+            "sub R25, R22", //         remainder -= divisor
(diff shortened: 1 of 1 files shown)


tones111:
Pinging @Patryk27 for testing assistance.  As indicated in the description, I've validated the algorithm with `cargo test` on x86-64, however, I haven't executed the code on AVR hardware or equivalent simulator.  Hopefully you can run it through your [AVR math fuzzer](https://github.com/Patryk27/avr-tester/blob/main/avr-tester/tests/acceptance/eval.rs).

Patryk27:
Ok, I've checked the code under avr-tester's testsuite and it seems to work, nice job!

One thing left to check is whether all AVRs support the instructions here (don't know by heart which ones belong to the base instruction set).

tones111:
Thanks for testing @Patryk27 .

I referenced Microchip's [AVR ISA](https://ww1.microchip.com/downloads/en/DeviceDoc/AVR-InstructionSet-Manual-DS40002198.pdf) for the opcodes.  I think we should be good since none of the instructions used are listed as "N/A" in the chapter 5 tables.

The gcc implementation uses macros to conditionally switch between `mov` and `movw`.  They also switch between `call` and `rcall` in the signed functions.  I would hope this could be implemented via an optimization pass, but that doesn't appear to be the case with current `rustc`.

If you're willing to mentor a bit I'd be interested in trying to help improve the AVR codegen (but don't have any LLVM-specific dev experience).

Patryk27:
> If you're willing to mentor a bit I'd be interested in trying to help improve the AVR codegen (but don't have any LLVM-specific dev experience).

That'd be nice, but I think it's sort of gray area in this specific case - I'm pretty sure that `asm!()` blocks are supposed to be placed verbatim into the output binary, without doing any optimizations on them (otherwise the optimizer could just elide all `asm!("nop");` for instance). So even if there was a peephole optimization pass within LLVM, I'm pretty sure it wouldn't / couldn't be activated in this case; if we want to handle it, we should probably do it explicitly in here (using `#[cfg]`).

tgross35 (approved):
I double checked the clobbers and they seem to line up for both the default calling convention and for reduced tiny. I didn't check the routines in depth but they look reasonable to me.

Thanks for getting this started!

PR #817, opened 2025-04-13
use #[naked] as an unsafe attribute

folkertdev:
Per https://github.com/rust-lang/rust/pull/134213#issuecomment-2755984503, we want to make the `#[naked]` attribute an unsafe attribute.

Based on the CI failures here https://github.com/rust-lang/rust/pull/139753, I believe that `compiler-builtins` need to be updated first.

### DIFF
--- a/compiler-builtins/src/aarch64.rs
+++ b/compiler-builtins/src/aarch64.rs
@@ -3,7 +3,7 @@
 use core::intrinsics;
 
 intrinsics! {
-    #[naked]
+    #[unsafe(naked)]
     #[cfg(all(target_os = "uefi", not(feature = "no-asm")))]
     pub unsafe extern "C" fn __chkstk() {
         core::arch::naked_asm!(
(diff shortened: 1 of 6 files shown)


tgross35:
Would it be possible for rustc to accept both forms until this gets updated, to bootstrap this? Unfortunately I don't know if we could merge and release this with the failures here.

Another case where a subtree would be nice...

folkertdev:
Right, yeah not ideal but we'll make it work. This PR will then likely need some `#[allow(unused-unsafe)]` for a bit too while we make this change, but we'll see when that PR makes it into a nightly build.

folkertdev:
Allright, this works now that https://github.com/rust-lang/rust/pull/139797 is in nightly. After merging this, I guess  the next step is to create a release of `compiler-builtins`, and then make rustc use that new release?

tgross35 (approved):
Thanks for the upstream changes, your description is correct. I'll sync it through today.

PR #821, opened 2025-04-17
try using `#[cfg(bootstrap)]` for `#[unsafe(naked)]`

folkertdev:
Based on [#t-compiler/help > Does &#96;#&#91;cfg(bootstrap)&#93;&#96; get set in dependencies?](https://rust-lang.zulipchat.com/#narrow/channel/182449-t-compiler.2Fhelp/topic/Does.20.60.23.5Bcfg.28bootstrap.29.5D.60.20get.20set.20in.20dependencies.3F) this should resolve the error we see in https://github.com/rust-lang/rust/pull/139934.

It's probably a good idea to validate that this will actually work, but I'm not sure how to do that.

### DIFF
--- a/compiler-builtins/Cargo.toml
+++ b/compiler-builtins/Cargo.toml
@@ -70,3 +70,6 @@ rustc-dep-of-std = ['compiler-builtins', 'core']
 # This makes certain traits and function specializations public that
 # are not normally public but are required by the `testcrate`
 public-test-deps = []
+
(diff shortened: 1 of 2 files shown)


folkertdev:
I think I got it, I just failed to actually read the code before.

tgross35 (approved):
I haven't had a chance to test this all the way through, but might as well try it

PR #860, opened 2025-04-20
Add cr_log from CORE-Math

tgross35:


awxkee:
If you want 0.5 ULP math [here](https://github.com/awxkee/pxfm) is also Rust source. Not everything comes from CORE-Math but it should be correct for nearest rounding mode and is pretty straightforward to copy.

PR #861, opened 2025-04-20
libm: Add `fmaf16`

tgross35:
Recreated from https://github.com/rust-lang/libm/pull/419

### DIFF
--- a/libm/src/math/fmaf16.rs
+++ b/libm/src/math/fmaf16.rs
@@ -1,0 +1,258 @@
+/* SPDX-License-Identifier: MIT OR Apache-2.0
+ * origin: original implementation, 2026 (TG) */
+
+use crate::support::{CastFrom, Float, Int, unbounded_shr_u64};
+
+/// We use a a U21.43 fixed-point representation when needed.
+const FIXED_FRAC_BITS: u32 = 43;
+
+/// Floating multiply add (f16)
+///
+/// Computes `(x*y)+z`, rounded as one ternary operation (i.e. calculated with infinite precision).
+#[cfg_attr(assert_no_panic, no_panic::no_panic)]
+pub fn fmaf16(x: f16, y: f16, z: f16) -> f16 {
+    let ix = x.to_bits() & !f16::SIGN_MASK;
+    let iy = y.to_bits() & !f16::SIGN_MASK;
+    let iz = z.to_bits() & !f16::SIGN_MASK;
+
+    let xneg = x.is_sign_negative();
+    let yneg = y.is_sign_negative();
+    let zneg = z.is_sign_negative();
+
+    let mneg = xneg ^ yneg;
+
+    if ix == 0 || ix >= f16::EXP_MASK || iy == 0 || iy >= f16::EXP_MASK {
+        // Value will overflow, defer to non-fused operations.
+        return x * y + z;
+    }
+
+    if iz == 0 {
+        // Empty add component means we only need to multiply.
+        return x * y;
+    }
+
+    if iz >= f16::EXP_MASK {
+        // `z` is NaN or infinity, which sets the result.
+        return z;
+    }
+
+    let mut xexp = x.ex();
+    let mut yexp = y.ex();
+    let mut zexp = z.ex();
+
+    let mut xsig = ix & f16::SIG_MASK;
+    let mut ysig = iy & f16::SIG_MASK;
+    let mut zsig = iz & f16::SIG_MASK;
+
+    // If not subnormal, set the implicit bit
+    if xexp != 0 {
+        xsig |= f16::IMPLICIT_BIT;
+    }
+    if yexp != 0 {
+        ysig |= f16::IMPLICIT_BIT;
+    }
+    if zexp != 0 {
+        zsig |= f16::IMPLICIT_BIT;
+    }
+
+    // A biased exponent of 1 (min normal) and 0 (subnormal) have the same real exponent, so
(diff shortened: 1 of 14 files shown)


libm/src/math/fma_wide.rs
```diff
@@ -5,11 +5,10 @@ use super::support::{FpResult, IntTy, Round, Status};
 use super::{CastFrom, CastInto, DFloat, Float, HFloat, MinInt};
 
 // Placeholder so we can have `fmaf16` in the `Float` trait.
```
tmvkrpxl0: Is it really a placeholder if it (now) has implementation?

tgross35: Indeed, updated to a doc comment :)

tgross35:
Baseline at 1909bf7a5ce65782f5448a79d3793c059c0bdf25:

```text
icount::icount_bench_math_group::icount_bench_fmaf16 logspace:(setup_fmaf16())
  Baselines:                   arch_enabled|arch_enabled (old)
  Instructions:                       64543|N/A                  (*********)
  L1 Hits:                            71915|N/A                  (*********)
  LL Hits:                                3|N/A                  (*********)
  RAM Hits:                              18|N/A                  (*********)
  Total read+write:                   71936|N/A                  (*********)
  Estimated Cycles:                   72560|N/A                  (*********)
```

libm/src/math/fmaf16.rs
```diff
@@ -0,0 +1,303 @@
...
+            rneg = false;
+            m64.wrapping_sub(z64)
+        } else {
+            rneg = zneg;
+            z64.wrapping_sub(m64)
+        }
+    } else {
+        rneg = mneg;
+        m64 + z64
+    };
```
quaternic, line 145: At this point we have the result as a sign-magnitude fixed-point value that just needs rounding. The rest of this function might be a useful to factor out, and test independently. Something like
```rust
/// Rounds the value `n * 2^s` to `f16`
fn f16_from_u64_with_scale(n: u64, s: i32) -> f16
```
And this could return
```rust
f16_from_u64_with_scale(r64, -40).copysign(/* from rneg */)
```


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/861/99fdc8a9ccf6f3d07068391e077931143c8cf4aa..1909bf7a5ce65782f5448a79d3793c059c0bdf25)*

libm/src/math/fmaf16.rs
```diff
@@ -0,0 +1,303 @@
...
+        // rounding overflowed to the next power
+        r >>= 1;
+        rexp += 1;
+        println!("round rexp: {rexp}");
+    }
+    if r & (1 << 14) != 0 {
+        // ensure subnormals that rounded up get the exponent of 1
+        rexp = rexp.max(1);
+        println!("sub rexp: {rexp}");
+    }
```
quaternic, line 192: A handy trick for doing the rounding: The effect of rounding a finite magnitude up is always just `+1` to the bits of the float. If the mantissa had its maximum value, it carries into the exponent field but that still gives the correct value because the mantissa is then all zeroes.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/861/99fdc8a9ccf6f3d07068391e077931143c8cf4aa..1909bf7a5ce65782f5448a79d3793c059c0bdf25)*

tgross35:
Scratch: https://rust.godbolt.org/z/T9Tb66q8W

libm/src/math/fmaf16.rs
tgross35: This should be added back now

tgross35:
On x86 and other targets that need to do f16 math via soft f32 conversions, it's probably more efficient to continue with the rest of this routine than pay for an `__extendhfsf2`/`__truncsfhf2` roundtrip for `x * y` early exits. Perhaps even unconditionally since many arches with any `f16` support also have an `fmaf16` instruction so don't need this routine, and there's a good chance that anything without fmaf16 but with cheap f16<->f32 will have cheap f16<->f64 and also won't need it.

I'll investigate this more in a followup.

libm/src/math/fmaf16.rs
tgross35: (I'll get this in my next round)

tgross35:
Okay, I think this should be in a pretty reasonable state.

libm/src/math/fmaf16.rs
```diff
@@ -0,0 +1,258 @@
...
+    let iz = z.to_bits() & !f16::SIGN_MASK;
+
+    let xneg = x.is_sign_negative();
+    let yneg = y.is_sign_negative();
+    let zneg = z.is_sign_negative();
+
+    let mneg = xneg ^ yneg;
+
+    if ix == 0 || ix >= f16::EXP_MASK || iy == 0 || iy >= f16::EXP_MASK {
+        // Value will overflow, defer to non-fused operations.
```
quaternic:
```suggestion
        // The product is either zero or non-finite, defer to non-fused operations.
```

quaternic (approved):
LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/861/4d3ab8695dcf965be926882c92878b99017bf99b..6bc30d943072e76e928819bd3742ab37fcb086fb)*

tgross35:
Thank you!

PR #879, opened 2025-04-21
cleanup: Reuse `MinInt` and `Int` from `libm` in `compiler-builtins`

tgross35:
Since the two crates are now in the same repo, it is easier to share
code. Begin some deduplication with the integer traits.


CONTRIBUTING.md
```diff
@@ -0,0 +1,167 @@
...
+  `core` itself).
+- Only use relative imports within the `math` directory / module, e.g.
+  `use self::fabs::fabs` or `use super::k_cos`. Absolute imports from core are
+  OK, e.g. `use core::u64`.
+- To reinterpret a float as an integer use the `to_bits` method. The MUSL code
+  uses the `GET_FLOAT_WORD` macro, or a union, to do this operation.
+- To reinterpret an integer as a float use the `f32::from_bits` constructor. The
+  MUSL code uses the `SET_FLOAT_WORD` macro, or a union, to do this operation.
+- You may use other methods from core like `f64::is_nan`, etc. as appropriate.
+- Rust does not have hex float literals. This crate provides two `hf16!`,
```
fbstj, line 54:
```suggestion
- Rust does not have hex float literals. This crate provides four `hf16!`,
```
or just drop the word

Issue #883, opened 2025-04-22
build: libm 0.2.12 cannot compile without edition 2024 support


MrCroxx:
When building project in edition 2021 with libm 0.2.12, this compile error pops up:

```bash
> cat Cargo.toml
[package]
name = "t"
version = "0.1.0"
edition = "2021"

[dependencies]
libm = "0.2.12"%
> cargo +nightly-2024-08-30 check
error: failed to download `libm v0.2.12`

Caused by:
  unable to get packages from source

Caused by:
  failed to parse manifest at `/home/mrcroxx/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libm-0.2.12/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.82.0-nightly (8f40fc59f 2024-08-21)).
  Consider trying a more recent nightly release.
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```

I checked the downloaded libm Cargo.toml:

```toml
# THIS FILE IS AUTOMATICALLY GENERATED BY CARGO
#
# When uploading crates to the registry Cargo will automatically
# "normalize" Cargo.toml files for maximal compatibility
# with all versions of Cargo and also rewrite `path` dependencies
# to registry (e.g., crates.io) dependencies.
#
# If you are reading this file be aware that the original Cargo.toml
# will likely look very different (and much more reasonable).
# See Cargo.toml.orig for the original contents.

[package]
edition = "2021"
rust-version = "1.63"
name = "libm"
version = "0.2.12"
authors = ["Jorge Aparicio <jorge@japaric.io>"]
build = "build.rs"
autolib = false
autobins = false
autoexamples = false
autotests = false
autobenches = false
description = "libm in pure Rust"
documentation = "https://docs.rs/libm"
readme = "README.md"
keywords = [
    "libm",
    "math",
]
categories = ["no-std"]
license = "MIT"
repository = "https://github.com/rust-lang/compiler-builtins"
resolver = "3"

[features]
arch = []
default = ["arch"]
force-soft-floats = []
unstable = [
    "unstable-intrinsics",
    "unstable-float",
]
unstable-float = []
unstable-intrinsics = []
unstable-public-internals = []

[lib]
name = "libm"
path = "src/lib.rs"

[dev-dependencies.no-panic]
[...]

PR #884, opened 2025-04-22
fix: Switch to resolver v2

tgross35:
The published crates fail to build with an edition less than 2024 because they are packaged with `resolver = "3"`, which is a 2024-only option. Revert back to resolver v2 to drop this requirement.

Fixes: https://github.com/rust-lang/compiler-builtins/issues/883


MrCroxx (approved):
LGTM, thank you for the quick response!

PR #901, opened 2025-05-01
Enable testing against MPFR on windows-gnu

tgross35:
ci: skip-extensive
ci: test-libm

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -192,26 +192,26 @@ jobs:
         python3 etc/update-api-list.py --check
         cargo test -p update-api-list
 
-    - uses: msys2/setup-msys2@v2
-      if: matrix.os == 'windows-2025'
-      with:
-        msystem: UCRT64
-        update: true
-        install: diffutils m4 make # mingw-w64-rustup
+    # - uses: msys2/setup-msys2@v2
+    #   if: matrix.os == 'windows-2025'
+    #   with:
+    #     msystem: UCRT64
+    #     update: true
+    #     install: diffutils m4 make # mingw-w64-rustup
 
-    - name: Setup windows
-      if: matrix.os == 'windows-2025'
-      shell: msys2 {0}
-      run: |
-        ./ci/run-windows.sh ${{ matrix.target }}
+    # - name: Setup windows
+    #   if: matrix.os == 'windows-2025'
+    #   # shell: msys2 {0}
+    #   run: |
+    #     ./ci/run-windows.sh ${{ matrix.target }}
 
-    - name: Run on Windows
-      if: matrix.os == 'windows-2025'
-      shell: msys2 {0}
-      run: |
-        rustc -vV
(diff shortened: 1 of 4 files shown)


tgross35:
@mati865 any chance you'd be able to give me a hand here if you're familiar with MinGW on GHA? Our more thorough tests need to build [`gmp_mpfr_sys`](https://docs.rs/gmp-mpfr-sys/latest/gmp_mpfr_sys/#building-on-windows) which _should_ work on windows-gnu, but I can't seem to get a setup that has both needed tools (gcc, make, m4, diffutils) and Rust.

mati865:
I can take a look tomorrow.

tgross35:
Feel free to scrap this if you know of a better way than going through msys2/setup-msys2 in GHA. The actual functional part of this diff is ` *windows-gnu*) ;;` -> `*windows-gnu*) mflags+=(--features libm-test/build-mpfr) ;;` in `ci/run.sh`.

mati865:
@tgross35 I think what is missing is just setting up the PATH to contain rustup. Windows has some questionable design decisions for using PATH for binaries and shared objects (dynamic libraries), so MSYS2 by default doesn't include it entirely: https://github.com/msys2/setup-msys2#path-type

So the suggestion from Rustup:
> To get started you may need to restart your current shell.
This would reload its PATH environment variable to include
Cargo's bin directory (%USERPROFILE%\.cargo\bin).

Is not helpful at this time.
To solve it, you can either add Cargo bin dir explictly (`export PATH=$USERPROFILE/.cargo/bin:$PATH`) or setup MSYS2 to inherit whole PATH.

mati865:
So there are 2 solutions:
- add `echo 'export PATH="/c/Users/$USERNAME/.cargo/bin:$PATH"' >> ~/.bash_profile` in `ci/run_windows.sh`
- add `path-type: inherit` in `.github/workflows/main.yaml`

Then Rust installed with rustup is found. I don't have much insight on what happens later, but if you still need help, let me know.

tgross35:
Great catch and thanks for finding it, I didn't get a chance to look into it further. This is a compiler version difference https://gcc.godbolt.org/z/YTd6hnbM3.

I think the code is actually valid as the old school untyped function signatures, it would have to be `void g(void) {}` to actually be a zero-parameter function. I guess I wouldn't be too surprised if gmp is using k&r syntax somewhere.

tgross35:
Looks like they're aware of it https://gmplib.org/list-archives/gmp-bugs/2024-November/005550.html

tgross35:
It has been fixed upstream but that hasn't made it to the Rust version yet. Opened https://gitlab.com/tspiteri/gmp-mpfr-sys/-/issues/39 requesting an update.

tgross35:
Maybe there is a bit of an easier solution - the most recent GCC seems to accept the code with `--std=C17`. If I can get CFLAGS to the right place then I think that might work.

PR #903, opened 2025-05-02
Use rounding instructions on aarch64

TDecking:
This extends the use of dedicated rounding instructions to the remaining rounding related functions such as `trunc`.
This also fixes an incorrect use of the `frintn` instruction, which was used for the implementation of `rint` instead of `roundeven`.

ci: allow-many-extensive

### DIFF
--- a/etc/function-definitions.json
+++ b/etc/function-definitions.json
@@ -98,6 +98,7 @@
     },
     "ceil": {
         "sources": [
+            "libm/src/math/arch/aarch64.rs",
             "libm/src/math/arch/i586.rs",
             "libm/src/math/arch/wasm32.rs",
             "libm/src/math/ceil.rs",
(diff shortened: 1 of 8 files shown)


tgross35:
I had a similar change as part of the original diff for https://github.com/rust-lang/libm/pull/459. The result was that we can omit these since we will have them in `core` soon (https://github.com/rust-lang/rust/pull/138087) and LLVM will do a reasonable lowering from there.

Regarding `rint`: with Rust/LLVM we don't interact with fpenv and also don't make any guarantees about it to C code that calls Rust functions, so `frintn` is the most simple instruction to meet our uses of `rint`.  There is a bit of discussion at https://github.com/rust-lang/libm/pull/459#discussion_r1926780808.

TDecking:
Thanks for the info. Does this mean that this PR is moot and can be closed?

As for`rint`, I based may decision of the fact that LLVM generates `frintx`, so I assumed the use of `frintn`
was an oversight and not a deliberate design decision.

tgross35:
Sorry - yeah, I'll close this since I don't think we need the changes here. I opened https://github.com/rust-lang/compiler-builtins/pull/924 to give us some indication about why we use frintn.

PR #904, opened 2025-05-03
Enable testing against MPFR on windows-gnu

mati865:


mati865:
Guess there is no need for this PR anymore.

PR #905, opened 2025-05-05
Disable aarch64 assembly if `target_abi = "softfloat"`

tgross35:
The target `aarch64-unknown-none-softfloat` still sets the `neon` target feature, so the assembly implementations are incorrectly being enabled. Update gates to differentiate the `softfloat` ABI.


tgross35:
Nevermind, I was reading the `--print cfg` output wrong.

PR #911, opened 2025-05-08
Remove remaining traces of AT&T assembly syntax

TDecking:
The current minimum LLVM version seems to be 18, way larger than required for resolving this FIXME.


compiler-builtins/src/mem/x86_64.rs
```diff
@@ -22,13 +22,12 @@ use core::{intrinsics, mem};
 #[inline(always)]
 #[cfg(target_feature = "ermsb")]
 pub unsafe fn copy_forward(dest: *mut u8, src: *const u8, count: usize) {
-    // FIXME: Use the Intel syntax once we drop LLVM 9 support on rust-lang/rust.
-    core::arch::asm!(
-        "repe movsb (%rsi), (%rdi)",
+    asm!(
+        "repe movsb [rdi], rsi",
```
tgross35, line 26: In the original asm both rsi and rdi are dereferenced, here it’s only rdi. Is this still equivalent for some reason?

TDecking: I don't think it is. I fixed this.

compiler-builtins/src/mem/x86_64.rs
```diff
@@ -83,20 +82,19 @@ pub unsafe fn copy_backward(dest: *mut u8, src: *const u8, count: usize) {
         inout("rdi") dest.add(count - 1) => _,
         inout("rsi") src.add(count - 1) => _,
         // We modify flags, but we restore it afterwards
-        options(att_syntax, nostack, preserves_flags)
+        options(nostack, preserves_flags)
```
quaternic, line 85: This definitely shouldn't have `preserves_flags`. Flags are modified by each of `add`,`sub`,`test`. The comment refers to how `std`/`cld` are used to set/clear the direction flag, but that's separate since it *must* be cleared before exiting the inline assembly.

Actually, what is the point of the `test` instruction here? It only affects flags, and none of the subsequent instructions depend on any flags.

quaternic (changes requested):

compiler-builtins/src/x86.rs
```diff
@@ -28,26 +27,25 @@ intrinsics! {
...
-            "push   (%eax)",        // push return address onto the stack
-            "sub    %esp,%eax",     // restore the original value in eax
+            "3:",
+            "sub    ecx, eax",
+            "test   [ecx], ecx",
+            "lea    eax, [esp + 4]",  // load pointer to the return address into eax
+            "mov    esp, ecx",     // install the new top of stack pointer into esp
+            "mov    ecx, [eax - 4]", // restore ecx
+            "push   [eax]",        // push return address onto the stack
+            "sub    eax, esp",     // restore the original value in eax
```
tgross35: Nit: comments used to be aligned

tgross35:
@quaternic the issues you mentioned seem to be preexisting. Would you mind putting up a PR to fix them? We can do that independent of the syntax swap.

Then @TDecking you won't have to make any changes aside from resolving conflicts. Would you be able to rebase?

compiler-builtins/src/mem/x86_64.rs
tgross35: I think it can be removed: ae557bde4efcd85bf1fc75c488c4370749d969a7 had an early exit but ef37a23d8417afdb9fb6f215ec4c651e18146366 removed it.

tgross35: This diff gave me a similar question about https://github.com/rust-lang/compiler-builtins/blob/e1d8a4965f0d49a738f152377152648d40cc055f/compiler-builtins/src/x86_64.rs#L20-L26. Aren't flags set by `test` overwritten by the `sub` then `cmp` before `ja` checks them?

quaternic:
@tgross35 I can, but I'd much prefer doing that in intel syntax, so merging this with those preexisting issues would be fine.

compiler-builtins/src/mem/x86_64.rs
quaternic: Yeah, that looked weird at first, but I think the `test` there just used as a way to load from memory without modifying any registers (except flags). So the loop is just reading from memory at 4kB intervals for the side-effects only, which is exactly what a `chkstk` should be doing.

compiler-builtins/src/probestack.rs
```diff
@@ -112,14 +112,13 @@ pub unsafe extern "custom" fn __rust_probestack() {
             // for this target, [manually patch for LVI].
             //
             // [manually patch for LVI]: https://software.intel.com/security-software-guidance/insights/deep-dive-load-value-injection#specialinstructions
-            pop %r11
+            pop r11
```
TDecking: Something I've noticed is that LLVM automatically inserts `lfence` instructions on sgx targets. Right now, `lfence` appears twice.

TDecking:
I've rebased and removed the final instances of `att_syntax` in `probestack.rs`.

tgross35 (approved):
LGTM as a 1:1 change, thanks for the updates!

PR #912, opened 2025-05-08
Fix `i256::MAX`

TDecking:


tgross35:
Good catch, thanks. Would you mind updating the version at https://github.com/rust-lang/compiler-builtins/blob/a4c748f72a1dce652cc3e41c3a8425731bd1519a/libm/src/math/support/big.rs#L84-L87 as well? (I'll hopefully be deduplicating these to share the libm version in the near future since it's a bit faster).

I should figure out a way to sanity check our constants somehow (maybe also against `rug` like libm-test/tests/u256.rs).

tgross35 (approved):
Thanks!

PR #917, opened 2025-05-18
fix an if statement that can be collapsed

folkertdev:
This is flagged by clippy in 1.87

This is the simple solution. We can use `let_chains`, but currently the `libm-macros` crate looks like it builds on stable.


tgross35:
Actually, would you mind doing the suggested fix? It might build on stable but there isn't any reason it has to, I'd love start using let chains anyway.

PR #918, opened 2025-05-21
Enable `__powitf2` on MSVC

beetrees:
`__divtf3` is now available on nightly, so CI should pass.


tgross35 (approved):
Thanks! I’ve been meaning to clean this up but never remembered at the right time.

Issue #919, opened 2025-05-22
Float CmpResult seems to be of the wrong type


antoyo:
Hi.
I'm currently trying to make some new f128 tests pass for cg_gcc and I believe I found out the issue: it seems the `CmpResult` type is aliased wrongly to `i32` while it should be aliased to `i64` (or at least `long`) on x86-64.

The type is [defined here](https://github.com/rust-lang/compiler-builtins/blob/f1b9055b0d09370a474ce5e9b9d78ac4758cedf8/compiler-builtins/src/float/cmp.rs#L12) as a `i32` while the comment just above point to [this](https://github.com/llvm/llvm-project/blob/1e6ba3cd2fe96be00b6ed6ba28b3d9f9271d784d/compiler-rt/lib/builtins/fp_compare_impl.inc#L25) that mentions that it should be a `long`.

Could you please double-check that it indeed should be a `long` and not a `i32`?

Thanks.

antoyo:
> Out of curiosity, how did you notice the difference here? I guess if LLVM only reads the lower bits of the return but GCC checks the whole register, that would be visible.

In cg_gcc, I saw the following instruction to check the result:
```asm
test    rax, rax
```
while cg_llvm had:
```asm
test     eax, eax
```

> I tried to double check against GCC but the exact logic is a bit hidden behind the `__libgcc_cmp_return__` mode and I have no idea where that is actually getting set

From a quick glance, this is a target hook that can be overwritten by the backend (see here for [Aarch64](https://github.com/rust-lang/gcc/blob/04ce66d8c918de9273bd7101638ad8724edf5e21/gcc/config/aarch64/aarch64.cc#L31686) for instance).
The default value seems to be defined [here](https://github.com/rust-lang/gcc/blob/04ce66d8c918de9273bd7101638ad8724edf5e21/gcc/targhooks.cc#L245).

Thanks!

tgross35:
Oh, so I guess my searches just didn't find anything because GH doesn't index the 30k line GCC target files :)

Posted a fix in https://github.com/rust-lang/compiler-builtins/pull/920, would you mind reviewing?

antoyo:
> Posted a fix in [#920](https://github.com/rust-lang/compiler-builtins/pull/920), would you mind reviewing?

Yes, I'll test this with cg_gcc when the CI passes.

tgross35:
[...]

PR #920, opened 2025-05-23
Update `CmpResult` to use a pointer-sized return type

tgross35:
As seen at [1], LLVM uses `long long` on LLP64 (to get a 64-bit integer matching pointer size) and `long` on everything else, with exceptions for AArch64 and AVR. Our current logic always uses an `i32`. This happens to work because LLVM uses 32-bit instructions to check the output on x86-64, but the GCC checks the full 64-bit register so garbage in the upper half leads to incorrect results.

Update our return type to be `isize`, with exceptions for AArch64 and AVR.

Fixes: https://github.com/rust-lang/compiler-builtins/issues/919

[1]: https://github.com/llvm/llvm-project/blob/0cf3c437c18ed27d9663d87804a9a15ff6874af2/compiler-rt/lib/builtins/fp_compare_impl.inc#L11-L27

### DIFF
--- a/builtins-test/benches/float_cmp.rs
+++ b/builtins-test/benches/float_cmp.rs
@@ -1,12 +1,23 @@
 #![cfg_attr(f128_enabled, feature(f128))]
 
 use builtins_test::float_bench;
-use compiler_builtins::float::cmp;
+use compiler_builtins::float::cmp::{self, CmpResult};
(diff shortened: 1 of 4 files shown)


antoyo:
Could you please make a release with this fix?
This is blocking the sync of cg_gcc.
Thanks.

PR #921, opened 2025-05-28
Add tooling for josh syncs

tgross35:
Create a justfile that handles pulling from and pushing to
rust-lang/rust. This can be invoked with the following:

    $ just rustc-pull
    $ just rustc-push


tgross35:
Superseded by https://github.com/rust-lang/compiler-builtins/pull/942

PR #931, opened 2025-05-29
ci: Skip extensive tests if `to_test` is empty

tgross35:
Have this show up as skipped in the UI, rather than passed.


tgross35:
Doesn't seem to work unfortunately. Thought I saw this somewhere but I guess I must have confused it.

PR #955, opened 2025-06-14
josh-sync: Replace `#xxxx`-style links in messages

tgross35:
Often our short summaries will pick up a Bors "Auto merge of #xxxx ...` commit message. Replace these with something like `rust-lang/rust#1234` to avoid broken links when going between repositories.

### DIFF
--- a/crates/josh-sync/src/sync.rs
+++ b/crates/josh-sync/src/sync.rs
@@ -1,7 +1,10 @@
+use std::borrow::Cow;
 use std::net::{SocketAddr, TcpStream};
 use std::process::{Command, Stdio, exit};
 use std::time::Duration;
 use std::{env, fs, process, thread};
+
+use regex_lite::Regex;
 
 const JOSH_PORT: u16 = 42042;
 const DEFAULT_PR_BRANCH: &str = "update-builtins";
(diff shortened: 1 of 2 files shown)


crates/josh-sync/src/sync.rs
```diff
@@ -77,6 +80,7 @@ impl GitSync {
             "--depth=1",
         ]);
         let new_summary = check_output(["git", "log", "-1", "--format=%h %s", &new_upstream_base]);
+        let new_summary = replace_references(&new_summary, &self.upstream_repo);
```
Urgau, line 83: If this interest you, in triagebot bot we are using this regex, https://github.com/rust-lang/triagebot/blob/1657187393c4ed9979dfab3901a67b4b832cade0/src/handlers/issue_links.rs#L18-L19

tgross35: Oh that's great, I had no idea `\B` was a thing. Thanks!

Urgau: To avoid false positives (like `#123az`) I would also highly recommend using `\b` at the end.

tgross35: Indeed, missed that. Added, thanks

crates/josh-sync/src/sync.rs
```diff
@@ -297,6 +301,13 @@ fn check_output_cfg(prog: &str, f: impl FnOnce(&mut Command) -> &mut Command) ->
     String::from_utf8(out.stdout.trim_ascii().to_vec()).expect("non-UTF8 output")
 }
 
+/// Replace `#1234`-style issue/PR references with `prefix#1234` to ensure links work across
```
fbstj:
```suggestion
/// Replace `#1234`-style issue/PR references with `repo#1234` to ensure links work across
```
?

tgross35: Thanks!

Issue #123, opened 2016-12-10
Depending on compiler-builtins creates more unresolved symbols.


turnage:
I am building for arm-none-eabihf (no_std) with the following target file:

````
{
    "llvm-target": "arm-none-eabihf",
    "target-endian": "little",
    "target-pointer-width": "32",
    "os": "none",
    "env": "eabi",
    "vendor": "unknown",
    "arch": "arm",

    "data-layout": "e-m:e-p:32:32-i64:64-v128:64:128-a:0:32-n32-S64",
    "executables": true,
    "relocation-model": "static",
    "no-compiler-rt": true
}
````

I build with ````RUST_TARGET_PATH=`pwd` xargo build --target arm-none-eabihf````.

Before adding a dependency on compiler-builtins, I get the following unresolved symbols, found by ````arm-none-eabi-nm target/arm-none-eabihf/debug/libtyran.a | grep "U "````:

````
         U _ZN4core9panicking5panic17h86aded962ecadfccE
         U __aeabi_memclr4
         U __aeabi_memcpy
         U __aeabi_memcpy4
         U __aeabi_memset
         U __aeabi_uidiv
         U __aeabi_uidivmod
         U __aeabi_ul2d
         U __aeabi_ul2f
         U __aeabi_uldivmod
         U __aeabi_unwind_cpp_pr0
         U __aeabi_unwind_cpp_pr1
         U memcmp
         U __mulodi4
         U rust_begin_unwind
         U __sync_val_compare_and_swap_1
         U __sync_val_compare_and_swap_2
         U __sync_val_compare_and_swap_4
````

After adding compiler-builtins to my dependencies, the extern crate statement and corresponding feature in my lib.rs, I get the following undefined symbols found by the same command:

````
         U _ZN4core9panicking5panic17h86aded962ecadfccE
         U __aeabi_ddiv
         U __aeabi_dmul
         U __aeabi_fdiv
         U __aeabi_fmul
         U __aeabi_unwind_cpp_pr0
         U __aeabi_unwind_cpp_pr1
         U memcpy
         U memmove
         U memset
         U rust_eh_personality
         U _Unwind_Resume
         U _ZN4core9panicking5panic17h86aded962ecadfccE
         U __aeabi_memclr4
         U __aeabi_memcpy
         U __aeabi_memcpy4
         U __aeabi_memset
         U __aeabi_uidiv
         U __aeabi_uidivmod
         U __aeabi_ul2d
[...]

PR #956, opened 2025-06-14
Add minimal triagebot config

Urgau:
This PR adds a minimal `triagebot.toml` config to make contributions to this repository respect upstream rust-lang/rust conventions and avoid issues when syncing this subtree.

If you're interested in more triagebot feature, I highly recommand looking at [miri's `triagebot.toml` config](https://github.com/rust-lang/miri/blob/master/triagebot.toml).

Accompanying team PR: https://github.com/rust-lang/team/pull/1876

cf. https://github.com/rust-lang/rust/pull/142489#issuecomment-2972474814
cc @tgross35

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -1,0 +1,21 @@
+## See <https://forge.rust-lang.org/triagebot/index.html> for documentation
+## of these features.
+
+# Warns when a PR contains merge commits
+# Documentation at: https://forge.rust-lang.org/triagebot/no-merge.html
+[no-merges]
+exclude_titles = ["Update from"]
+
+# Canonicalize issue numbers to avoid closing the wrong issue
+# when commits are included in subtrees, as well as warning links in commits.
+# Documentation at: https://forge.rust-lang.org/triagebot/issue-links.html
+[issue-links]
+check-commits = false
+
+# Prevents mentions in commits to avoid users being spammed
+# Documentation at: https://forge.rust-lang.org/triagebot/no-mentions.html
+[no-mentions]
+
+# Enable issue transfers within the org
+# Documentation at: https://forge.rust-lang.org/triagebot/transfer.html
+[transfer]


tgross35:
Adding it LGTM, but I believe this needs `check-commits = false` since we pretty often need to link to r-l/rust issues

Urgau:
Then they should be put in the PR description and not in the commit description, we don't want to spam the issues with references to the commit. It also creates issues with subtrees, especially if someone puts `Fixes #123` in the commit description. It's explained in [the warning message](https://github.com/rust-lang/rust/pull/142478#issuecomment-2971900883).

tgross35:
I don't really follow that logic - if links are in the PR description on a repo that uses merge commits (r-l/r), then the links wind up in those merge commits and still make noise from forks on issues. In fact, that's actually worse because of how Josh does syncs - it (horribly) needs to recreate every totally unrelated merge commit for every merge from the same rollup, e.g. the last sync https://github.com/rust-lang/compiler-builtins/pull/946 wound up mentioned in all these totally unrelated issues:

* https://github.com/rust-lang/cargo/issues/3946
* https://github.com/rust-lang/rust/issues/141619
* https://github.com/rust-lang/rust/pull/134814
* https://github.com/rust-lang/stdarch/pull/1706
* ... many more

And then in a repo that doesn't use merge commits (this one), the links are just lost and you have no useful crossrefs in the log/blame.

It would make sense to disallow `Fixes:`/`Closes:`/etc links that can accidentally close issues in other repos, but not allowing any links in history seems pretty strong.

Urgau:
> if links are in the PR description on a repo that uses merge commits (r-l/r), then the links wind up in those merge commits and still make noise from forks on issues

Unfortunately, we currently can't do much about them. We *unfortunately* can't change GitHub default merge commit description and the one bors generates is parsed by some tooling making it difficult to change.

> And then in a repo that doesn't use merge commits (this one), the links are just lost and you have no useful crossrefs in the log/blame.

That's a good point. When there isn't a merge commit the PR is not cross-referenced in a commit.
There is still a reference on the UI tough.
image

> It would make sense to disallow `Fixes:`/`Closes:`/etc links that can accidentally close issues in other repos, but not allowing any links in history seems pretty strong.

That's what we were doing at the start, but the same issue about spam and wrongly referenced issue exist without the `Fixes`/..., the solution for the wrongly referenced one is to use canonicalized issue links (ie prefixed with the org and repo name), but that won't do anything for the spam issue, furthermore by moving the issue links to the PR description triagebot can automatically update the issue links with canonicalized version, reducing a bit the contributor frustration.

> but not allowing any links in history seems pretty strong.

Technically it's a warning, it's not blocking in anyway.

------

Anyway, added `check-commits = false` to `[issue-links]`.

I will see how we can better accommodate repos without merge commits.

tgross35:
I wish github would just allow you to filter mentions by comment vs. commit and by organization so this wasn't a concern. One of the things I miss most from GitLab 🙃 

Hopefully bors2 will help clean up the merge messages a bit. With any luck it will also be able to use octopus merges for rollups so the merge commits don't get so out of hand...

In any case, thank you for getting this set up! I don't think there is any harm in merging before the team PR merges.

PR #957, opened 2025-06-14
use `extern "custom"` on naked functions with a custom calling convention

folkertdev:
I'm not sure whether the atomic functions in https://github.com/rust-lang/compiler-builtins/blob/master/compiler-builtins/src/aarch64_linux.rs follow the C abi, so I didn't touch them for now.


compiler-builtins/src/arm.rs
tgross35, line 1: The `__aeabi_idiv` extern block should also be updated

tgross35 (commented):
Thanks for putting this together! It was on my todo list but 

> I'm not sure whether the atomic functions in https://github.com/rust-lang/compiler-builtins/blob/master/compiler-builtins/src/aarch64_linux.rs follow the C abi, so I didn't touch them for now.

I believe they are actually the C ABI, or at least https://community.arm.com/arm-community-blogs/b/tools-software-ides-blog/posts/making-the-most-of-the-arm-architecture-in-gcc-10 has an example calling a similar function directly and they seem to obey the input register calling convention.

compiler-builtins/src/arm.rs
```diff
@@ -9,19 +9,16 @@ unsafe extern "C" {
 }
 
 // SAFETY: these are defined in compiler-builtins
-// FIXME(extern_custom), this isn't always the correct ABI
-unsafe extern "aapcs" {
-    // AAPCS is not always the correct ABI for these intrinsics, but we only use this to
-    // forward another `__aeabi_` call so it doesn't matter.
-    fn __aeabi_idiv(a: i32, b: i32) -> i32;
+unsafe extern "custom" {
```
tgross35, line 12:
```suggestion
unsafe extern "custom" {
    // AAPCS is not always the correct ABI for these intrinsics, but we only use this to
    // forward another `__aeabi_` call so it doesn't matter.
```

Issue #852, opened 2025-02-25
musl upstream is down


udoprog:
Seems like https://git.musl-libc.org/git/musl is down (504 Bad Gateway), meaning building this project from git no longer works.

I'm not sure how to contact the musl project, so I'm opening an issue here for now since this is the project which breaks for me.

It might be an idea to host a mirror of the project on GitHub with the commit being depended on.

tgross35:
Could anybody confirm whether or not this is still a problem? I assume Cargo was doing something weird, but now that the musl submodule is no longer within the crate's source directory I'm wondering if it fixed itself.

tgross35:
Since https://github.com/rust-lang/compiler-builtins/commit/0c7a82c63410008b97450fa0acd2bffb25664894 we use a mirror here.

PR #958, opened 2025-06-14
use `is_multiple_of` to check if an addr is aligned

folkertdev:


folkertdev:
Looks like this hit https://github.com/rust-lang/compiler-builtins/issues/852 e.g. here https://github.com/rust-lang/compiler-builtins/actions/runs/15650960447/job/44095645760?pr=958

tgross35 (approved):
Thanks!

PR #961, opened 2025-06-29
libm: Improved integer utilities, implement shifts and bug fixes for i256 and u256

quaternic:
`i256` and `u256`
- operators now use the same overflow convention as primitives
- implement `<<` and `-` (previously just `>>` and `+`)
- implement `Ord` correctly (the previous `PartialOrd` was broken)
- corrected `i256::SIGNED` to `true`

~~The above mentioned operators are now included in the `MinInt`-trait.~~
(Never mind, that would need additional changes in compiler-builtins, so I removed it for now.)

The `Int`-trait is extended with `trailing_zeros`, `carrying_add`, and `borrowing_sub`.

### DIFF
--- a/libm-test/benches/icount.rs
+++ b/libm-test/benches/icount.rs
@@ -120,6 +120,22 @@ fn icount_bench_u256_add(cases: Vec<(u256, u256)>) {
 }
 
 #[library_benchmark]
+#[bench::linspace(setup_u256_add())]
+fn icount_bench_u256_sub(cases: Vec<(u256, u256)>) {
+    for (x, y) in cases.iter().copied() {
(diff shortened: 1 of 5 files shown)


tgross35:
The implementations all look pretty great to me. Would it be possible to extend the fuzz tests in `libm-test/tests/u256.rs` to cover these?

There are also some benchmarks, it would be good to basically copy the `add` and `shr` benchmarks for `sub` and `shl`. https://github.com/rust-lang/compiler-builtins/blob/0bbec7238890242e7754d6ff604c989666328d83/libm-test/benches/icount.rs#L57-L127

quaternic:
I've now added both of those.

Issue #837, opened 2025-01-06
Replace i586 `ceil` and `floor` implementations with assembly to fix +/-0


tgross35:
From https://github.com/rust-lang/libm/pull/404#issuecomment-2572399955:

> i think for `i586` it should just use `frndint` with setting the x87 CW (actually not slow) for `floor`/`ceil`. like: https://github.com/NetBSD/src/blob/20433927938987dd64c8f6aa46904b7aca3fa39e/lib/libm/arch/i387/s_ceil.S

PR #976, opened 2025-07-13
implement `floor` and `ceil` in assembly on `i586`

folkertdev:
fixes https://github.com/rust-lang/compiler-builtins/issues/837

The assembly is based on

- https://github.com/NetBSD/src/blob/20433927938987dd64c8f6aa46904b7aca3fa39e/lib/libm/arch/i387/s_floor.S
- https://github.com/NetBSD/src/blob/20433927938987dd64c8f6aa46904b7aca3fa39e/lib/libm/arch/i387/s_ceil.S

Which both state

```
/*
 * Written by J.T. Conklin <jtc@NetBSD.org>.
 * Public domain.
 */
```

Which I believe means we're good in terms of licensing.

---

Are changes needed here?

https://github.com/rust-lang/compiler-builtins/blob/599f0e6408bef29a69f986afc71fa76e732ddd60/libm-test/src/precision.rs#L118-L124

These lines were added in https://github.com/rust-lang/libm/pull/404, but don't have a direct relation to `floor`/`ceil` so maybe it's unrelated.


libm/src/math/arch/i586.rs
```diff
@@ -1,37 +1,63 @@
...
+        "andw    $0xfbff,%dx",
+        "movw    %dx,-8(%ebp)",
+        // Load modified control word
+        "fldcw   -8(%ebp)",
+        // Round.
+        "fldl    8(%ebp)",
+        "frndint",
+        // Restore original control word.
+        "fldcw   -4(%ebp)",
+        //
```
tgross35: Is this missing a comment?

tgross35: Oh, I guess just a section break. Maybe just say "// Restore esp and ebp and return"

libm/src/math/arch/i586.rs
```diff
@@ -1,37 +1,63 @@
...
-        let truncated = x as i64 as f64;
-        if truncated < x {
-            return truncated + 1.0;
-        } else {
-            return truncated;
-        }
-    } else {
-        return x;
-    }
+/// Based on https://github.com/NetBSD/src/blob/trunk/lib/libm/arch/i387/s_ceil.S
```
tgross35: It doesn't say it's a copyright but it's probably good to preserve the author anyway

```rust
/// Based on https://github.com/NetBSD/src/blob/trunk/lib/libm/arch/i387/s_ceil.S
/// (written by J.T. Conklin <jtc@NetBSD.org>).
```

tgross35:
This is awesome, thank you for implementing it!

There is one problem, our `libm` MSRV unfortunately doesn't support naked functions. I'll be bumping it in the near future but I don't think it will be that high; how hard is this to do with inline asm? Probably adds a few instructions because the loads stack access becomes opaque, but that would still be faster than the slow existing implementation.

tgross35:
Fix for the red CI in https://github.com/rust-lang/compiler-builtins/pull/979 btw

libm-test/src/precision.rs
```diff
@@ -292,16 +280,6 @@ impl MaybeOverride<(f64,)> for SpecialCase {
...
-        if cfg!(x86_no_sse)
-            && (ctx.fn_ident == Identifier::Ceil || ctx.fn_ident == Identifier::Floor)
-            && expected.eq_repr(F::NEG_ZERO)
-            && actual.eq_repr(F::ZERO)
-        {
-            // FIXME: the x87 implementations do not keep the distinction between -0.0 and 0.0.
-            // See https://github.com/rust-lang/libm/pull/404#issuecomment-2572399955
-            return XFAIL("i586 ceil/floor signed zero");
-        }
-
```
folkertdev, line 304: I believe the removal of `eq_repr` here causes 

```
 error: function `f32_to_bits` is never used
   --> libm/src/math/support/float_traits.rs:367:14
    |
367 | pub const fn f32_to_bits(x: f32) -> u32 {
    |              ^^^^^^^^^^^
    |
    = note: `-D dead-code` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(dead_code)]`

error: function `f64_to_bits` is never used
   --> libm/src/math/support/float_traits.rs:381:14
    |
381 | pub const fn f64_to_bits(x: f64) -> u64 {
    |              ^^^^^^^^^^^
```

via biteq. How should we handle that (allow, remove code, something else)?

libm/src/math/arch/i586.rs
folkertdev: Fixed

libm-test/src/precision.rs
tgross35: Nah that's not this PR, `#[allow(dead_code)]` on traits just stopped working recently https://github.com/rust-lang/rust/issues/144060. Fix in https://github.com/rust-lang/compiler-builtins/pull/979

folkertdev:
> how hard is this to do with inline asm?

Well, I don't know this target and its assembly that well. So using `global_asm!` would be much easier (for me, anyway) and more straightforward to upgrade once the MSRV does support naked functions.

I think the only downside is that the symbols might be visible from the outside? Is that a problem?

folkertdev:
actually we can prevent even that with a trick that Björn showed me recently using `sym`

folkertdev:
from what I can find this is a 2024 edition thing

```

error: extern blocks must be unsafe
  --> builtins-shim/../compiler-builtins/src/math/../../../libm/src/math/arch/i586.rs:23:1
   |
23 | / extern "cdecl" {
24 | |     fn ceil_helper(_: f64) -> f64;
25 | |     fn floor_helper(_: f64) -> f64;
26 | | }
   | |_^
```

but rust 1.63 won't compile `unsafe extern` right?

tgross35:
> I think the only downside is that the symbols might be visible from the outside? Is that a problem?

I think it may be, also conflicts (though we could make it weak). But this should be pretty easy to move to inline asm; I think something like this would work https://rust.godbolt.org/z/P98rKEW7P, which is reasonably close to the original (not sure if there's a way to turn pointers to locals into stack offsets, to avoid the `leal` s)

folkertdev:
> not sure if there's a way to turn pointers to locals into stack offsets, to avoid the leal s

That's probably fine, we can fix that properly once naked functions are available

---

we now get errors like this?

```
  stderr ───
    Spaced Musl sinf arg 1/1: 800 iterations (800 total)

    thread 'musl_quickspace_sinf' panicked at libm-test/tests/compare_built_musl.rs:26:50:
    called `Result::unwrap()` on an `Err` value: 
        input:    (-239880100.0,)
        as hex:   (-0x1.c988f4p+27,)
        as bits:  (0xcd64c47a,)
        expected: -0.17352822            -0x1.6362c4p-3 0xbe31b162
        actual:   -0.17353132            -0x1.636464p-3 0xbe31b232
```

quaternic:
You're clobbering the saved control word on the stack with the modified one, so it isn't restored correctly.

I think there's an argument for using a fixed control word for the `frndint`, but I'll have to come back to this later.

libm/src/math/arch/i586.rs
```diff
@@ -1,37 +1,67 @@
...
+    // We save and later restore the FPU control word.
+    let mut cw_orig: u16 = 0;
+    unsafe {
+        core::arch::asm!(
+            "fstcw   ({cw_ptr})",      // Save the cw
+            "movw    ({cw_ptr}), %dx", // ...
+            "orw     $0x0800, %dx",    // Set rounding control to 0b10 (+∞),
+            "andw    $0xfbff, %dx",    // preserving other controls
+            "movw    %dx, ({cw_ptr})", // Apply cw
+            "fldcw    ({cw_ptr})",     // ...
```
tgross35: Sorry, yeah, as @quaternic mentioned this needs a second stack slot. As a microopt this could be `let mut cw_stash = [0u16; 2];` and then these accesses can be `-4({stash_ptr})` so we save a register vs. two different `u16` locals

tgross35: Actually even better, `let mut cw_stash = MaybeUninit::<[u16; 2]>::uninit();` to save the zero init instruction

folkertdev: I got this to work

https://rust.godbolt.org/z/44Tjcx3bb

the `-4` stuff didn't work (also, for `u16`, why `-4`?)

libm/src/math/arch/i586.rs
```diff
@@ -1,37 +1,69 @@
...
-            return truncated;
-        }
-    } else {
-        return x;
+///
+/// Based on https://github.com/NetBSD/src/blob/trunk/lib/libm/arch/i387/s_floor.S
+/// (written by J.T. Conklin <jtc@NetBSD.org>).
+pub fn floor(mut x: f64) -> f64 {
+    // We save and later restore the FPU control word.
+    let mut cw_stash = [core::mem::MaybeUninit::<u16>::uninit(); 2];
```
beetrees:
```suggestion
    let mut cw_temp = core::mem::MaybeUninit::<u16>::uninit();
    let mut stash_temp = core::mem::MaybeUninit::<u16>::uninit();
```

Since this is never used as an array, just as two disjoint variables, it's clearer what's going on if these are separate variables.

folkertdev:
Somehow these changes introduce a panic somewhere??

```
error: linking with `cc` failed: exit status: 1
  |
  = note:  "cc" "-m32" "<1 object files omitted>" "-Wl,--as-needed" "-Wl,-Bstatic" "<sysroot>/lib/rustlib/i586-unknown-linux-gnu/lib/libcompiler_builtins-*.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-L" "/tmp/rustcBJvx8Y/raw-dylibs" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-L" "<sysroot>/lib/rustlib/i586-unknown-linux-gnu/lib" "-o" "/home/folkertdev/rust/compiler-builtins/target/i586-unknown-linux-gnu/release-opt/deps/libm-7098e2e2ad176b85" "-Wl,--gc-sections" "-pie" "-Wl,-z,relro,-z,now" "-Wl,-O1" "-Wl,--strip-debug" "-nodefaultlibs"
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: ld: error: undefined symbol: 
          
          ERROR[no-panic]: detected panic in function `rem_pio2`
          
          >>> referenced by libm.ddf30f7f94fa857e-cgu.0
          >>>               /home/folkertdev/rust/compiler-builtins/target/i586-unknown-linux-gnu/release-opt/deps/libm-7098e2e2ad176b85.libm.ddf30f7f94fa857e-cgu.0.rcgu.o:(core::ops::function::FnOnce::call_once::h239ee4d4be0e829c)
          >>> referenced by libm.ddf30f7f94fa857e-cgu.0
          >>>               /home/folkertdev/rust/compiler-builtins/target/i586-unknown-linux-gnu/release-opt/deps/libm-7098e2e2ad176b85.libm.ddf30f7f94fa857e-cgu.0.rcgu.o:(core::ops::function::FnOnce::call_once::hf87b8ecfbb337da8)
          >>> referenced by libm.ddf30f7f94fa857e-cgu.0
          >>>               /home/folkertdev/rust/compiler-builtins/target/i586-unknown-linux-gnu/release-opt/deps/libm-7098e2e2ad176b85.libm.ddf30f7f94fa857e-cgu.0.rcgu.o:(libm::math::rem_pio2::rem_pio2::h2b1fed0a2d9377e5)
          
          ld: error: undefined symbol: 
          
          ERROR[no-panic]: detected panic in function `j1f`
```

The culprit is `floor`, but eh, wut? Maybe inlining no longer happens (sufficiently) with the inline assembly?

libm/src/math/arch/i586.rs
tgross35: No idea, took that from the original asm without checking :)

tgross35:
That's bizarre. The failure didn't show up with `naked_asm` right? I have no idea what could be different but that seems like an interesting thread to pull...

I'll try to look on Monday but it would be good to trace back where that panic is coming from, maybe an `assert_unchecked` can get rid of it (I'll figure out how to do the MSRV dance around that).

beetrees:
After doing a bit of testing, it appears that any inline ASM block that doesn't have the `pure` option causes `no_panic` to fail.

tgross35:
> After doing a bit of testing, it appears that any inline ASM block that doesn't have the `pure` option causes `no_panic` to fail.

Is this bug in no_panic or is does codegen for non-`pure` asm somehow prevent eliminating an actual panic?

If it's the first case, I think it might be okay to work around this by updating `use_arch_required` at https://github.com/folkertdev/compiler-builtins/blob/8f43b0b5f568c688c3504f89097221c2d16749b4/libm/src/math/floor.rs and `ceil` to include `not(assert_no_panic)`.

folkertdev:
I think it's a codegen/optimization issue. When I move the `floor` and `ceil` implementations into their own binary, and add `#[no_panic]`, it fails in debug mode (unsurprising), but succeeds with `--release`. So it is not the assembly inherently.

https://godbolt.org/z/aad7TYaWf (with `no_panic` turned on locally, godbolt does not have that sadly)

On the other hand in the `compiler-builtins` codebase even a `asm!("nop")` is enough to trigger a panic apparently. Neither `inline(always)` nor `inline(never)` changes that.

As far as I can tell there is no actual panic there (see e.g. the godbolt)

quaternic:
> I think there's an argument for using a fixed control word for the `frndint`, but I'll have to come back to this later.

So, what I had in mind there was that while the implementation from NetBSD is surely more correct in supporting unmasked exceptions, we really don't have that support anywhere in the language (except maybe naked functions) so a chunk of it is unnecessary.

I believe this should work just as well:
(using intel syntax)
```rust
extern "C" fn floor(mut x: f64) -> f64 {
    unsafe {
        core::arch::asm!(
            "fld qword ptr [{x}]",
            "fstcw [{x}]",
            "mov word ptr [{x} + 2], 0x077f",
            "fldcw [{x} + 2]",
            "frndint",
            "fldcw [{x}]",
            "fstp qword ptr [{x}]",
            x = in(reg) &mut x,
            out("st(0)") _, out("st(1)") _,
            out("st(2)") _, out("st(3)") _,
            out("st(4)") _, out("st(5)") _,
            out("st(6)") _, out("st(7)") _,
            options(nostack),
        );
    }
    x
}
```
https://godbolt.org/z/8b6a8Yxn1
It's a bit of a hack to reuse the memory of `x` for the control words, but since the memory operand is needed regardless, might as well.

folkertdev:
That implementation has the right behavior, but still runs into the panic issue.

quaternic:
Yeah, I didn't mean to imply it would do anything for that, and there's currently no way to make x87 inline assembly `pure` since that would require support for either memory operands or x87 register operands.

---

As for the panic, it seems that all the errors trace back to the same source: `rem_pio2_large` contains one [call](https://github.com/rust-lang/compiler-builtins/blob/2cdde03950e8d10d37b8d13e1c1d3ce2e54f36c4/libm/src/math/rem_pio2_large.rs#L293) to `floor`, and inserting `asm!("")` next to that introduces the same no-panic errors.

Since that call looks to be replaceable with `trunc`, that would be a possible workaround. (Testing now: #982 )

tgross35:
Does making the functions `extern "C"` change things? I wonder if that could be the difference with the naked implementation, because of assumptions around unwinding with extern ABIs.

If that doesn't work, I think the change in https://github.com/rust-lang/compiler-builtins/pull/982 is fine if it gets us out of this issue. 

> I think it's a codegen/optimization issue. When I move the `floor` and `ceil` implementations into their own binary, and add `#[no_panic]`, it fails in debug mode (unsurprising), but succeeds with `--release`. So it is not the assembly inherently.

Do you mean putting the definitions in a separate object file but still checking the panic in `rem_pio2`? If so, that could be falling into `no_panic` needing LTO https://github.com/rust-lang/compiler-builtins/blob/2cdde03950e8d10d37b8d13e1c1d3ce2e54f36c4/Cargo.toml#L46-L51

folkertdev:
Locally at least, for `floor` the combination of `extern "C"` with `inline(never)` gets rid of the panics. I'm assuming that is fine (e.g. the naked function version also would not get inlined).

libm/src/math/arch/i586.rs
```diff
@@ -1,37 +1,71 @@
...
-            return truncated + 1.0;
-        } else {
-            return truncated;
-        }
-    } else {
-        return x;
+pub extern "C" fn ceil(mut x: f64) -> f64 {
+    // We save and later restore the FPU control word.
+    let mut cw_stash = core::mem::MaybeUninit::<u16>::uninit();
+    let mut cw_tmp = core::mem::MaybeUninit::<u16>::uninit();
```
tgross35: Could this be imported since it is used a few times?

libm/src/math/arch/i586.rs
```diff
@@ -1,37 +1,71 @@
...
-        let truncated = x as i64 as f64;
-        if truncated > x {
-            return truncated - 1.0;
-        } else {
-            return truncated;
-        }
-    } else {
-        return x;
+// NOTE: this `inline(never)` is load-bearing for making functions that use it panic-free.
+// Without it `rem_pio2_large` (and any function that uses it) will contain panics.
```
tgross35: Do they actually contain panics? Or is it a false positive that the pattern used by `no_panic` doesn't handle

tgross35 (commented):
Few small nits but it looks good to me, @quaternic could you also review?

I do think it is okay to use https://github.com/rust-lang/compiler-builtins/pull/982 as well so we can allow `floor` to be inlined

libm/src/math/arch/i586.rs
folkertdev: I haven't been able to actually pinpoint where the reported panic really is. e.g. https://godbolt.org/z/TcEGeYos9 doesn't show any panics. So there is a chance that it's a limitation in `no_panic`.

quaternic:
> Does making the functions `extern "C"` change things? I wonder if that could be the difference with the naked implementation, because of assumptions around unwinding with extern ABIs.

I did some more testing, and I think this might be an issue in rustc (well, more of a feature request). Functions containing inline asm with an ABI that allows unwinding are considered potentially unwinding (so don't get the `nounwind` attribute applied). Calls to such functions then generate landing pads and cause `no_panic` to fail. This is even if the asm is marked `pure` (as long as the result is used).

https://godbolt.org/z/x9cYPovMz

So my understanding is that in this case:
 - `floor` passes `no_panic` with either ABI, since it doesn't contain any calls.
 - `rem_pio2_large` passes `no_panic` if `floor` is `extern "C"` or gets inlined
 - However, e.g. `rem_pio2f` fails `no_panic` since the call to `rem_pio2_large` isn't inlined (since it's sufficiently complex and also called elsewhere), it is `extern "Rust"` (may unwind), and it contains the asm from the inlined `floor`.

So, to fix this specific case while still allowing inlining in the general case, the simplest fix that I could think of is to wrap the call with`#[inline(never)]` only in `rem_pio2_large` specifically.

---

I've had some doubts that substituting `trunc` there might not be correct in some edge case, so it's probably safer to not touch the original logic without more thought. I've repurposed #982 into the above idea. That's using `#[cfg(x86_no_sse)]` as the toggle, but it should probably match whatever is used to choose this implementation.

---

Other that that, this looks all good by me. Thanks for writing proper comments for the asm!

tgross35:
Thank you for looking into this more. Would you mind writing up a rust-lang/rust issue with your findings? That would be good to link from the fixme in https://github.com/rust-lang/compiler-builtins/pull/982.

tgross35:
> Functions containing inline asm with an ABI that allows unwinding are considered potentially unwinding

Shouldn't `extern "C"` disallow unwinding?

I'm wondering what a solution would look like here. Are you thinking something like a `nounwind` asm option?

quaternic:
Created issue at https://github.com/rust-lang/rust/issues/144518, there's some more context there.

`extern "C"` does disallow unwinding, and it's also currently just UB to unwind out of `asm!`. The problem here happens when the `extern "C"`-function is inlined into something with an ABI that does allow unwinding, and then that is called (and not inlined).

libm/src/math/arch/i586.rs
```diff
@@ -1,37 +1,64 @@
...
+//!
+//! We use an alternative implementation on x86, because the
+//! main implementation fails with the x87 FPU used by
+//! debian i386, probably due to excess precision issues.
+//!
+//! See https://github.com/rust-lang/compiler-builtins/pull/976 for discussion on why these
+//! functions are implemented in this way.
 
-use super::super::fabs;
+// FIXME: when the MSRV allows, use naked functions instead.
```
tgross35: Is there actually any reason to do this now? Being able to inline this assembly should be advantageous, and from a quick look at godbolt it looks like this function is only 13 instructions (compared to 14 from the NetBSD implementations).

tgross35:
Okay! I think this just needs to rebase to drop the first commit, and drop the `extern "C"` (assuming `no_panic` works without it now) then should be all set.

folkertdev:
Cool, I've made those changes and CI is happy.

tgross35:
Thanks for making this happen!

Cc @programmerjake since this was your suggestion a long time ago

tgross35:
The output, for reference:

```asm
.section .text.ceil,"ax",@progbits
        .hidden ceil
        .weak   ceil
        .p2align        4
.type   ceil,@function
ceil:
        .cfi_startproc
        sub esp, 12
        .cfi_def_cfa_offset 16
        fld qword ptr [esp + 16]
        fstp qword ptr [esp]
        mov eax, esp
        #APP

        fld qword ptr [eax]
        wait
        fnstcw word ptr [eax]
        mov word ptr [eax + 2], 2943
        fldcw word ptr [eax + 2]
        frndint
        fldcw word ptr [eax]
        fstp qword ptr [eax]

        #NO_APP
        fld qword ptr [esp]
        add esp, 12
        .cfi_def_cfa_offset 4
        ret

.section .text.floor,"ax",@progbits
        .hidden floor
        .weak   floor
        .p2align        4
.type   floor,@function
floor:
        .cfi_startproc
        sub esp, 12
        .cfi_def_cfa_offset 16
        fld qword ptr [esp + 16]
        fstp qword ptr [esp]
        mov eax, esp
        #APP

        fld qword ptr [eax]
        wait
        fnstcw word ptr [eax]
        mov word ptr [eax + 2], 1919
        fldcw word ptr [eax + 2]
        frndint
        fldcw word ptr [eax]
        fstp qword ptr [eax]

        #NO_APP
        fld qword ptr [esp]
        add esp, 12
        .cfi_def_cfa_offset 4
        ret
```

PR #980, opened 2025-07-18
Change memcmp and bcmp return type to c_int

supersurviveur:
This PR fix the return type of `memcmp` and `bcmp` builtin function on target with a `pointer_width` equals to 16.
Linked issue: https://github.com/rust-lang/rust/issues/144076


compiler-builtins/src/mem/impls.rs
```diff
@@ -384,13 +384,13 @@ pub unsafe fn set_bytes(mut s: *mut u8, c: u8, mut n: usize) {
 }
 
 #[inline(always)]
-pub unsafe fn compare_bytes(s1: *const u8, s2: *const u8, n: usize) -> i32 {
+pub unsafe fn compare_bytes(s1: *const u8, s2: *const u8, n: usize) -> crate::mem::c_int {
```
tgross35: Could you import `core::ffi::c_int` and use `c_int` here instead?

I didn't even realize we had `crate::mem::c_int`, I'll probably remove that to just use `core::ffi` instead.

compiler-builtins/src/mem/impls.rs
```diff
@@ -384,13 +384,13 @@ pub unsafe fn set_bytes(mut s: *mut u8, c: u8, mut n: usize) {
...
     let mut i = 0;
     while i < n {
         let a = *s1.wrapping_add(i);
         let b = *s2.wrapping_add(i);
         if a != b {
-            return a as i32 - b as i32;
+            return a as crate::mem::c_int - b as crate::mem::c_int;
```
beetrees:
```suggestion
            return crate::mem::c_int::from(a) - crate::mem::c_int::from(b);
```

To make it clearer the conversions are lossless.

compiler-builtins/src/mem/impls.rs
tgross35: The changes in `compiler-builtins/src/mem/mod.rs` still need a full path `core::ffi::c_int` though, unfortunately the macro makes imports a bit weird.

tgross35: > I didn't even realize we had `crate::mem::c_int`, I'll probably remove that to just use `core::ffi` instead.

Doing this here https://github.com/rust-lang/compiler-builtins/pull/981

jthillard:
Ouuups pushed too early, I will do the other changes. 
I already changed to `core::ffi::c_int`

jthillard:
Should I revert my change for the `memset` function since @tgross35 did it in another PR ?

tgross35:
No need, it figured things out with a rebase. One nit though, in compiler-builtins/src/mem/impls.rs could you import `core::ffi::c_int`? Since it's used a few times might as well simplify the path.

Sorry, should have noticed before doing the rebase

tgross35 (approved):
Great, thanks for the fix!

PR #982, opened 2025-07-20
Avoid inlining `floor` into `rem_pio2`

quaternic:
Possible workaround for https://github.com/rust-lang/compiler-builtins/pull/976#issuecomment-3085530354

Inline assembly in the body of a function currently causes the compiler to consider that function possibly unwinding, even if said asm originated from inlining an `extern "C"` function. This patch wraps the problematic callsite with `#[inline(never)]`.

### DIFF
--- a/libm/src/math/rem_pio2_large.rs
+++ b/libm/src/math/rem_pio2_large.rs
@@ -11,7 +11,7 @@
  * ====================================================
  */
 
-use super::{floor, scalbn};
+use super::scalbn;
 
 // initial value for jk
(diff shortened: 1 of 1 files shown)


folkertdev:
This passes the test suite (just needs a formatting fix), and with this change the `floor`/`ceil` assembly versions also pass the test suite as expected.

tgross35:
Clippy noise fixed in https://github.com/rust-lang/compiler-builtins/pull/995

tgross35 (approved):

PR #994, opened 2025-07-25
Remove `no-asm` gating when there is no alternative implementation

tgross35:
Assembly-related configuration was added in 1621c6dbf9eb ("Use
`specialized-div-rem` 1.0.0 for division algorithms") to account for
Cranelift not yet supporting assembly. This hasn't been relevant for a
while, so we no longer need to gate `asm!` behind this configuration.
Thus, remove `cfg(not(feature = "no-asm"))` in places where there is no
generic fallback.

There are other cases, however, where setting the `no-asm` configuration
enables testing of generic version of builtins when there are platform-
specific implementations available; these cases are left unchanged. This
could be improved in the future by exposing both versions for testing
rather than using a configuration and running the entire testsuite
twice.

This is the compiler-builtins portion of
https://github.com/rust-lang/rust/pull/144471.

### DIFF
--- a/builtins-shim/Cargo.toml
+++ b/builtins-shim/Cargo.toml
@@ -37,8 +37,9 @@ default = ["compiler-builtins"]
 # implementations and also filling in unimplemented intrinsics
 c = ["dep:cc"]
 
-# Workaround for the Cranelift codegen backend. Disables any implementations
-# which use inline assembly and fall back to pure Rust versions (if available).
+# For implementations where there is both a generic version and a platform-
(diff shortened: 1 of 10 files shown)


tgross35:
On second thought, maybe it is a good thing to provide a way to require the generic implementations when they exist, so there is a way to keep testing them on platforms that have asm versions. We still don't need this for functions that only have an assembly implementation.

I'll update.

Amanieu (approved):

PR #1006, opened 2025-08-07
Remove instances of `allow(improper_ctypes)`

tgross35:
i128/u128 haven't flagged `improper_ctypes` for a while, and this just
made it to stable [1]. Remove the `allow`s as they are no longer needed.

[1]: https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/#i128-and-u128-in-extern-c-functions


niacdoial:
huh!
I was trying to do part of what this PR does, and I only saw you were already doing it because I was looking into how to make local tests work
anywho, thanks!

PR #1007, opened 2025-08-07
Start runnning tests for aarch64-pc-windows-msvc

tgross35:
This target is currently build-only. Switch to the windows-11-arm runner, which allows us to start running tests.

ci: test-libm

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -51,8 +51,7 @@ jobs:
         - target: aarch64-unknown-linux-gnu
           os: ubuntu-24.04-arm
         - target: aarch64-pc-windows-msvc
-          os: windows-2025
-          build_only: 1
+          os: windows-11-arm
         - target: arm-unknown-linux-gnueabi
           os: ubuntu-24.04
         - target: arm-unknown-linux-gnueabihf


tgross35:
@dpaoliello by the way, is f16/f128 support for arm64ec on the near future radar? As far as I know, since LLVM21 it is our only remaining target to still have an LLVM crash trying to use the `f16` type. (`f128` has a bit further to go)

dpaoliello:
> @dpaoliello by the way, is f16/f128 support for arm64ec on the near future radar? As far as I know, since LLVM21 it is our only remaining target to still have an LLVM crash trying to use the `f16` type. (`f128` has a bit further to go)

I'm afraid that I don't know, my team doesn't own defining the ABI/thunks.

@pmsjt @wesleywiser do either of you know if Arm64EC will have a defined way to define thunks for 16- and 128-bit floating point types?

tgross35:
I assumed you may be the one to know since you self-assigned https://github.com/llvm/llvm-project/issues/94434 :)

I would expect at least f16 should be reasonably easy, given aarch64 hardware support and a de facto x86 ABI of passing and returning in xmm (of course that is unfortunately not MSVC-blessed though). That's the type we are much closer to stabilizing anyway.

wesleywiser:
Unfortunately, I don't have any information on that

PR #1002, opened 2025-08-01
Optimize `fmod` with a method using integer multiplication

quaternic:
This is kind of a retry at #898. One of the problems there was that it would have added overhead and regressed performance for typical inputs.

Unlike that PR, this doesn't aim for sub-linear scaling; the cost of evaluating `fmod(x, y)` is still roughly proportional to `log2(|x/y|)`. However, the constant factor is much better. Running the `random`-benchmarks locally, I got walltime reductions of
```text
fmodf16:  -56.9%
fmodf:    -85.0%
fmod:     -95.4%
fmodf128: -98.7%
```

New utilities in `libm::support`:
 - `trait NarrowingDiv` for dividing `u2N / uN` when the quotient fits in `uN`
 - a reasonable implementation of that for up to `u256 / u128`
 - `fn linear_mul_reduction<U>(x: U, mut e: u32, y: U) -> U` computes `(x << e) % y` with the new method

PR #1011, opened 2025-08-12
libm: define and implement `trait NarrowingDiv` for unsigned integer division

quaternic:
New utility in `libm::support`:
 - `trait NarrowingDiv` for dividing `u2N / uN` when the quotient fits in `uN`
 - a reasonable implementation of that for primitives up to `u256 / u128`

This is the inverse operation of unsigned widening multiplication:
```rust
let xy: u256 = u128::widen_mul(x, y);
assert_eq!(xy.checked_narrowing_div_rem(y), Some((x, 0))); // unless y == 0
```

The trait API is based on x86's `div`-instruction: quotient overflow happens exactly when the high half of the dividend is greater or equal to the divisor, which includes division by zero.

Split from rust-lang/compiler-builtins#1002

### DIFF
--- a/libm-test/benches/icount.rs
+++ b/libm-test/benches/icount.rs
@@ -112,6 +112,17 @@ fn icount_bench_u128_widen_mul(cases: Vec<(u128, u128)>) {
 }
 
 #[library_benchmark]
+#[bench::linspace(setup_u128_mul())]
+fn icount_bench_u256_narrowing_div(cases: Vec<(u128, u128)>) {
+    use libm::support::NarrowingDiv;
+    for (x, y) in cases.iter().copied() {
+        let x = black_box(x.widen_hi());
+        let y = black_box(y);
+        black_box(x.checked_narrowing_div_rem(y));
+    }
+}
+
+#[library_benchmark]
 #[bench::linspace(setup_u256_add())]
 fn icount_bench_u256_add(cases: Vec<(u256, u256)>) {
     for (x, y) in cases.iter().copied() {
@@ -145,7 +156,7 @@ fn icount_bench_u256_shr(cases: Vec<(u256, u32)>) {
 
 library_benchmark_group!(
     name = icount_bench_u128_group;
-    benchmarks = icount_bench_u128_widen_mul, icount_bench_u256_add, icount_bench_u256_sub, icount_bench_u256_shl, icount_bench_u256_shr
+    benchmarks = icount_bench_u128_widen_mul, icount_bench_u256_narrowing_div, icount_bench_u256_add, icount_bench_u256_sub, icount_bench_u256_shl, icount_bench_u256_shr
 );
 
 #[library_benchmark]


libm/src/math/support/int_traits/narrowing_div.rs
```diff
@@ -0,0 +1,163 @@
...
+            }
+        }
+    };
+}
+
+impl_narrowing_div_primitive!(u16);
+impl_narrowing_div_primitive!(u32);
+impl_narrowing_div_primitive!(u64);
+impl_narrowing_div_primitive!(u128);
+impl_narrowing_div_recurse!(u256);
```
tgross35, line 80: Since this one is only used once, it can probably just be an impl without the macro. Please no `f256` 😆

libm/src/math/support/int_traits/narrowing_div.rs
```diff
@@ -0,0 +1,163 @@
...
+pub trait NarrowingDiv: DInt + MinInt<Unsigned = Self> {
+    /// Computes `(self / n, self % n))`
+    ///
+    /// # Safety
+    /// The caller must ensure that `self.hi() < n`, or equivalently,
+    /// that the quotient does not overflow.
+    unsafe fn unchecked_narrowing_div_rem(self, n: Self::H) -> (Self::H, Self::H);
+
+    /// Returns `Some((self / n, self % n))` when `self.hi() < n`.
+    fn checked_narrowing_div_rem(self, n: Self::H) -> Option<(Self::H, Self::H)> {
```
tgross35, line 20: Nit: from std's convention, this can just be `narrowing_div_rem`

libm/src/math/support/int_traits/narrowing_div.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+macro_rules! impl_narrowing_div_primitive {
+    ($D:ident) => {
+        impl NarrowingDiv for $D {
+            unsafe fn unchecked_narrowing_div_rem(self, n: Self::H) -> (Self::H, Self::H) {
+                if self.hi() >= n {
+                    unsafe { core::hint::unreachable_unchecked() }
+                }
+                ((self / n as $D) as Self::H, (self % n as $D) as Self::H)
```
tgross35: Could this use the `DInt`/`HInt` traits? Bit more clear than `as`.

`(self / n.widen()).lo(), (self % n.widen()).lo()`

It would be good to add a note that we're not doing anything special here since it optimizes well for primitives

tgross35: With that, I think this could even be a trait impl `impl<D> NarrowingDiv for D where D: ops::Div + ops::Rem`. The macro isn't too bad, though

libm/src/math/support/big/tests.rs
tgross35, line 1: Would it be posisble to add a bench for u256? Similar to https://github.com/rust-lang/compiler-builtins/blob/9c176c24e8b6295e2ba1c35d9713ef9e2d0055fb/libm-test/benches/icount.rs#L114-L144

libm/src/math/support/int_traits/narrowing_div.rs
```diff
@@ -0,0 +1,163 @@
...
+            unsafe fn unchecked_narrowing_div_rem(self, n: Self::H) -> (Self::H, Self::H) {
+                if self.hi() >= n {
+                    unsafe { core::hint::unreachable_unchecked() }
+                }
+
+                // Normalize the divisor by shifting the most significant one
+                // to the leading position. `n != 0` is implied by `self.hi() < n`
+                let lz = n.leading_zeros();
+                let a = self << lz;
+                let b = n << lz;
```
tgross35, line 59: Does it make any sense to check if `a.hi() == 0` and do a normal div in a smaller type in that case?

libm/src/math/support/int_traits/narrowing_div.rs
```diff
@@ -0,0 +1,163 @@
...
+/// # Safety
+/// Requires that `n.leading_zeros() == 0` and `a < n`.
+unsafe fn div_three_digits_by_two<U>(a0: U, a: U::D, n: U::D) -> (U, U::D)
+where
+    U: HInt,
+    U::D: Int + NarrowingDiv,
+{
+    if n.leading_zeros() > 0 || a >= n {
+        debug_assert!(false, "unsafe preconditions not met");
+        unsafe { core::hint::unreachable_unchecked() }
```
tgross35: Mind either adding a `debug_assert` to the other `unreachable_unchecked` cases or removing the one here, just to be consistent? Technically `unreachable_unchecked` has one that works correctly nowadays, but having an explicit case is still nice imo.

libm/src/math/support/int_traits/narrowing_div.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    #[test]
+    fn inverse_mul() {
+        for x in 0..=u8::MAX {
+            for y in 1..=u8::MAX {
+                let xy = x.widen_mul(y);
+                assert_eq!(xy.checked_narrowing_div_rem(y), Some((x, 0)));
+            }
+        }
+    }
```
tgross35: Maybe also check the `x.carrying_mul(y, 1) ` case?

libm/src/math/support/int_traits/narrowing_div.rs
```diff
@@ -0,0 +1,163 @@
...
+impl_narrowing_div_recurse!(u256);
+
+/// Implement `u3N / u2N`-division on top of `u2N / uN`-division.
+///
+/// Returns the quotient and remainder of `(a * R + a0) / n`,
+/// where `R = (1 << U::BITS)` is the digit size.
+///
+/// # Safety
+/// Requires that `n.leading_zeros() == 0` and `a < n`.
+unsafe fn div_three_digits_by_two<U>(a0: U, a: U::D, n: U::D) -> (U, U::D)
```
tgross35, line 89: Maybe `div_three_3x_by_2x` or so? Just thinking that the inputs aren't really digits

libm/src/math/support/int_traits/narrowing_div.rs
```diff
@@ -0,0 +1,163 @@
+use crate::support::{DInt, HInt, Int, MinInt, u256};
```
tgross35: If this is something you authored, could you add `/* SPDX-License-Identifier: MIT OR Apache-2.0 */`?

libm/src/math/support/int_traits/narrowing_div.rs
quaternic: I'll change them to use `(self / n.widen()).cast()`.

> It would be good to add a note that we're not doing anything special here since it optimizes well for primitives

They actually don't optimize as well as they could (at least on x86):
https://rust.godbolt.org/z/q5e8vc3nv
... but that's a separate issue to look into. (LLVM: https://github.com/llvm/llvm-project/issues/115158)
Anyway, it's good enough IMO, and I'll add a note on the macro.

libm/src/math/support/int_traits/narrowing_div.rs
quaternic: Which convention? `Option`-returning methods on e.g. `u32` all seem to be named `checked_*`

libm/src/math/support/int_traits/narrowing_div.rs
quaternic: The macro was convenient for more thorough testing by using the recursive construction all the way from `u16`, and I'd want to do that for any future changes to it.

libm/src/math/support/int_traits/narrowing_div.rs
quaternic: Yes, that could make sense, and it should be even better to check for `self.hi() == 0` before normalizing the divisor. However, I don't think that case would be hit very often. I also like how the current construction only depends on `NarrowingDiv` for the smaller type.

libm/src/math/support/int_traits/narrowing_div.rs
quaternic: They are digits in base `R = 2.pow(U::BITS)`, and this function does division of a 3-digit number by a 2-digit number in that base. Admittedly the signature is rather weird, passing the "high two thirds" of the 3-digit number as one `U::D` and the least significant third as a `U`.

libm/src/math/support/int_traits/narrowing_div.rs
tgross35: For anything where there isn't a panicking version it's usually `foo` and `foo_unchecked`, and most things in this crate try not to panic. But for methods on numbers it is usually panic by default, so 🤷‍♂️ I suppose

PR #1012, opened 2025-08-12
libm: implement accelerated computation of `(x << e) % y`

quaternic:
Builds on rust-lang/compiler-builtins#1011, both were split from rust-lang/compiler-builtins#1002

The computation is implemented in `linear_mul_reduction`.

### DIFF
--- a/libm/src/math/support/modular.rs
+++ b/libm/src/math/support/modular.rs
@@ -1,0 +1,301 @@
+/* SPDX-License-Identifier: MIT OR Apache-2.0 */
+
+//! To keep the equations somewhat concise, the following conventions are used:
+//!  - all integer operations are in the mathematical sense, without overflow
+//!  - concatenation means multiplication: `2xq = 2 * x * q`
+//!  - `R = (1 << U::BITS)` is the modulus of wrapping arithmetic in `U`
+
+use crate::support::int_traits::NarrowingDiv;
+use crate::support::{DInt, HInt, Int};
+
+/// Compute the remainder `(x << e) % y` with unbounded integers.
+/// Requires `x < 2y` and `y.leading_zeros() >= 2`
+#[allow(dead_code)]
+pub fn linear_mul_reduction<U>(x: U, mut e: u32, mut y: U) -> U
+where
+    U: HInt + Int<Unsigned = U>,
+    U::D: NarrowingDiv,
+{
+    assert!(y <= U::MAX >> 2);
+    assert!(x < (y << 1));
+    let _0 = U::ZERO;
+    let _1 = U::ONE;
+
+    // power of two divisors
+    if (y & (y - _1)).is_zero() {
+        if e < U::BITS {
+            // shift and only keep low bits
+            return (x << e) & (y - _1);
+        } else {
+            // would shift out all the bits
+            return _0;
+        }
+    }
+
+    // Use the identity `(x << e) % y == ((x << (e + s)) % (y << s)) >> s`
+    // to shift the divisor so it has exactly two leading zeros to satisfy
+    // the precondition of `Reducer::new`
+    let s = y.leading_zeros() - 2;
+    e += s;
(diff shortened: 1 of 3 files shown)


libm/src/math/support/int_traits.rs
```diff
@@ -296,7 +296,14 @@ int_impl!(i128, u128);
 
 /// Trait for integers twice the bit width of another integer. This is implemented for all
 /// primitives except for `u8`, because there is not a smaller primitive.
-pub trait DInt: MinInt {
+pub trait DInt:
+    MinInt
+    + core::ops::Add<Output = Self>
+    + core::ops::Sub<Output = Self>
+    + core::ops::Shl<u32, Output = Self>
+    + core::ops::Shr<u32, Output = Self>
```
tgross35: `core::ops` -> `ops` since it's already in scope

libm/src/math/support/modular.rs
tgross35, line 1: I think this commit should be able to allow the `#[allow(dead_code)]` on `NarrowingDiv`

libm/src/math/support/modular.rs
```diff
@@ -0,0 +1,195 @@
+use crate::support::int_traits::NarrowingDiv;
+use crate::support::{DInt, HInt, Int};
```
tgross35, line 9: Could you add a module-level doc comment with some of the common names used here? E.g. `R`, `RR` (if that's not `R*R`) `r`, `m`, `q`, `xq`. I'm unfortunately a bit lost :) (but I don't need to understand it in detail)

libm/src/math/support/modular.rs
```diff
@@ -0,0 +1,195 @@
...
+        e -= U::BITS - 1;
+    }
+    m.shift_reduce(e);
+
+    let rem = m.partial_remainder() >> y_shift;
+    rem.checked_sub(y).unwrap_or(rem)
+}
+
+#[cfg(test)]
+mod test {
```
tgross35, line 226: Could this spot check another integer size as well? Just using constants

libm/src/math/support/modular.rs
tgross35: If you've authored this, add `/* SPDX-License-Identifier: MIT OR Apache-2.0 */` as well (or only MIT if it's derived, as appropriate)

quaternic:
I've now rebased and applied the previous feedback.

@tgross35 
> Could you add a module-level doc comment with some of the common names used here? E.g. `R`, `RR` (if that's not `R*R`) `r`, `m`, `q`, `xq`. I'm unfortunately a bit lost :) (but I don't need to understand it in detail)

To make the module easier to follow, I moved the public interface, `fn linear_mul_reduction`, to the start of the file and added more explanatory comments. That should be understandable on its own. The implementation is essentially unchanged from before, other than one minor tweak.

libm/src/math/support/modular.rs
```diff
@@ -0,0 +1,251 @@
...
+        m.word_reduce();
+        e -= U::BITS - 1;
+    }
+    // Finish with the variable shift operation
+    m.shift_reduce(e);
+
+    // The partial remainder is in `[0, 2y)` ...
+    let r = m.partial_remainder();
+    // ... so check and correct, and compensate for the earlier shift.
+    r.checked_sub(y).unwrap_or(r) >> s
```
quaternic, line 57: That minor tweak:
Previously the right shift was done first, and then the `checked_sub` was with the original `y`. Now that's done first with the left-shifted `y`, and the shift is last.

quaternic:
I've gone through and rewritten the comments for the implementation of `Reducer`, which should actually be understandable now. (At least this time I didn't get too much of a headache from it.)

tgross35 (approved):
The comments are super thorough and I was able to (to a reasonable level) understand what's going on, thank you for all the updates!

PR #1019, opened 2025-09-08
ci: Use nextest on PowerPC64LE and s390x

taiki-e:
See https://github.com/taiki-e/install-action/issues/1056 for the context.

~~Depend on https://github.com/taiki-e/install-action/pull/1133.~~ UPDATE: released

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -132,10 +132,7 @@ jobs:
         rustup default "$channel"
         rustup target add "${{ matrix.target }}"
 
-    # Our scripts use nextest if possible. This is skipped on the native ppc
-    # and s390x runners since install-action doesn't support them.
     - uses: taiki-e/install-action@nextest
-      if: "!(matrix.os == 'ubuntu-24.04-ppc64le' || matrix.os == 'ubuntu-24.04-s390x')"
 
     - uses: Swatinem/rust-cache@v2
       with:


.github/workflows/main.yaml
```diff
@@ -132,10 +132,7 @@ jobs:
         rustup default "$channel"
         rustup target add "${{ matrix.target }}"
 
-    # Our scripts use nextest if possible. This is skipped on the native ppc
-    # and s390x runners since install-action doesn't support them.
     - uses: taiki-e/install-action@nextest
-      if: "!(matrix.os == 'ubuntu-24.04-ppc64le' || matrix.os == 'ubuntu-24.04-s390x')"
```
taiki-e, line 138: As I said in https://github.com/taiki-e/install-action/issues/1056#issuecomment-3111674746, it is actually probably better to use cache-cargo-install-action on these platforms.


> That said, considering that it is almost always necessary to do that on these platforms, using [cache-cargo-install-action](https://github.com/taiki-e/cache-cargo-install-action) is definitely better than install-action in terms of performance. (Eventually, the two may be merged, but there is no concrete plan yet: https://github.com/taiki-e/install-action/issues/66)

tgross35: Only on ppc/s390x or on all platforms? I'm happy to use either, just would prefer if the same one could be used for all platforms rather than using the special case.

I could have switched to cache-cargo-install-action here, but the scripts have a fallback to `cargo test` that worked fine.

taiki-e: cache-cargo-install-action is only efficient than install-action on platforms where pre-built binaries is not available, so if using only one of the two, install-action is better here (in this case, cache-cargo-install-action is only efficient for two jobs).

PR #1036, opened 2025-12-06
remove incorrectly placed preserves_flags, and other inline asm fixes

quaternic:
As discussed in rust-lang/compiler-builtins#911

### DIFF
--- a/compiler-builtins/src/mem/x86_64.rs
+++ b/compiler-builtins/src/mem/x86_64.rs
@@ -23,7 +23,7 @@ use core::{intrinsics, mem};
 #[cfg(target_feature = "ermsb")]
 pub unsafe fn copy_forward(dest: *mut u8, src: *const u8, count: usize) {
     asm!(
-        "repe movsb [rdi], [rsi]",
+        "rep movsb [rdi], [rsi]",
         inout("rcx") count => _,
         inout("rdi") dest => _,
         inout("rsi") src => _,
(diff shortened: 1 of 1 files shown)


compiler-builtins/src/mem/x86_64.rs
```diff
@@ -70,7 +70,6 @@ pub unsafe fn copy_backward(dest: *mut u8, src: *const u8, count: usize) {
         "sub rdi, 7",
         "mov rcx, {qword_count:r}",
         "rep movsq",
-        "test {pre_byte_count:e}, {pre_byte_count:e}",
```
quaternic, line 73: This was a leftover from https://github.com/rust-lang/compiler-builtins/commit/ef37a23d8417afdb9fb6f215ec4c651e18146366 which removed an early exit conditional jump but not the test for it.

tgross35 (approved):
Thanks! Needs to wait for https://github.com/rust-lang/compiler-builtins/pull/1002 but I'll merge after.

tgross35:
Yay for small perf improvements, a single instruction saved on memmove forward tests:

```text
 mem_icount::memmove::forward_move large_spread_off_0:setup_forward(Cfg { len : 16, spread : Large, off ...
  Baselines:                      softfloat|softfloat (old)
  Instructions:                         299|300                  (-0.33333%) [-1.00334x]
  L1 Hits:                              420|420                  (No change)
  LL Hits:                                0|1                    (-100.000%) [---inf---]
  RAM Hits:                              12|12                   (No change)
  Total read+write:                     432|433                  (-0.23095%) [-1.00231x]
  Estimated Cycles:                     840|845                  (-0.59172%) [-1.00595x]
```

I think that's the `test` and that I actually have forward and backward swapped at https://github.com/rust-lang/compiler-builtins/blob/8ab6f086310acb696d1d7146744cd4ab2367d018/builtins-test/benches/mem_icount.rs#L387-L396 and https://github.com/rust-lang/compiler-builtins/blob/8ab6f086310acb696d1d7146744cd4ab2367d018/builtins-test/benches/mem_icount.rs#L481-L490...

Issue #1046, opened 2025-12-12
Change to `Ord::clamp` causes build errors by calling through upstream monomorphisations


Dan54:
Following rust-lang/rust#147754, chromium encounters the following build error:
```
error: `compiler_builtins` cannot call functions through upstream monomorphizations; encountered invalid call from `core::fmt::num::<impl core::fmt::Debug for i32>::fmt` to `core::fmt::num::imp::<impl core::fmt::Display for i32>::fmt`
 --> ./../../third_party/rust-toolchain/lib/rustlib/src/rust/library/core/src/fmt/num.rs:79:0
 ::: ./../../third_party/rust-toolchain/lib/rustlib/src/rust/library/core/src/fmt/num.rs:97:1
  |
  = note: in this expansion of `impl_Debug!`
 ::: ./../../third_party/rust-toolchain/lib/rustlib/src/rust/library/core/src/fmt/num.rs:592:1
  |
  = note: in this macro invocation
```
The `clamp()` function is called twice in `libm/src/math/genric/scalbn.rs` (at lines [99](https://github.com/rust-lang/compiler-builtins/blob/fe052cc8b6e6eba8a9f5a034c5ad5e86375d4f6f/libm/src/math/generic/scalbn.rs#L99) and [106](https://github.com/rust-lang/compiler-builtins/blob/fe052cc8b6e6eba8a9f5a034c5ad5e86375d4f6f/libm/src/math/generic/scalbn.rs#L106)). I think the easiest fix would be to replace these calls with a locally defined `clamp()` that doesn't call `Debug::fmt`.

saethlin:
Hey! This is a compiler bug. @rust-lang/libs please try to report these instead of papering over them! Fortunately this was resurfaced by this comment https://github.com/rust-lang/rust/issues/137222#issuecomment-3688225174 or I would never have known about it.

PR #1047, opened 2025-12-14
remove uses of Ord::clamp in scalbn

quaternic:
Avoid using `Ord::clamp` in the `f16`-specific part of the generic `scalbn`.

fixes rust-lang/compiler-builtins#1046

It turned out to be redundant anyway, as both callsites follow a pattern like
```
if n < negative_val {
    let foo = (n + positive_val).clamp(negative_val, positive_val);
}
```
and `n < negative_val < 0` implies `n + positive_val < positive_val`.


zmodem:
Is this ready for review? Do we need to add someone?

tgross35 (approved):
Thank you!

PR #1048, opened 2025-12-16
Rustc pull (with fixes)

quaternic:
This is https://github.com/rust-lang/compiler-builtins/pull/1045 plus a patch for `builtins-test`


tgross35:
Thank you for the fix! I redid the auto job to avoid conflicts, then picked the fix patch to https://github.com/rust-lang/compiler-builtins/pull/1049.

PR #1050, opened 2025-12-27
Implement __sync builtins for thumbv6-none-eabi

taiki-e:
This is a PR for thumbv6-none-eabi (bere-metal Armv6k in Thumb mode) which proposed to be added by https://github.com/rust-lang/rust/pull/150138.

Armv6k supports atomic instructions, but they are unavailable in Thumb mode unless Thumb-2 instructions available (v6t2).

Using Thumb interworking (can be used via `#[instruction_set]`) allows us to use these instructions even from Thumb mode without Thumb-2 instructions, but LLVM does not implement that processing (as of LLVM 21), so this PR implements it in compiler-builtins.

The code around `__sync` builtins is basically copied from `arm_linux.rs` which uses kernel_user_helpers for atomic implementation.
The atomic implementation is a port of my [atomic-maybe-uninit inline assembly code].

This PR has been tested on QEMU 10.2.0 using patched compiler-builtins and core that applied the changes in this PR and https://github.com/rust-lang/rust/pull/150138 and the [portable-atomic no-std test suite] (can be run with `./tools/no-std.sh thumbv6-none-eabi` on that repo) which tests wrappers around `core::sync::atomic`. (Note that the target-spec used in test sets max-atomic-width to 32 and atomic_cas to true, unlike the current https://github.com/rust-lang/rust/pull/150138.) The original atomic-maybe-uninit implementation has been tested on real Arm hardware.

(Note that Armv6k also supports 64-bit atomic instructions, but they are skipped here. This is because there is no corresponding code in `arm_linux.rs` (since the kernel requirements increased in 1.64, it may be possible to implement 64-bit atomics there as well. see also https://github.com/taiki-e/portable-atomic/pull/82), the code becomes more complex than for 32-bit and smaller atomics.)

cc @thejpster (target maintainer)

I'll undraft this PR once the target maintainer approved this approach.

[atomic-maybe-uninit inline assembly code]: https://github.com/taiki-e/atomic-maybe-uninit/blob/HEAD/src/arch/arm.rs
[portable-atomic no-std test suite]: https://github.com/taiki-e/portable-atomic/tree/HEAD/tests/no-std-qemu

### DIFF
--- a/compiler-builtins/src/sync/thumbv6k.rs
+++ b/compiler-builtins/src/sync/thumbv6k.rs
@@ -67,6 +67,9 @@ macro_rules! atomic {
                 let mut out: Self;
                 // SAFETY: the caller must guarantee that the pointer is valid for read and write
                 // and aligned to the element size.
+                //
+                // Instead of the common `fence; ll/sc loop; fence` form, we use the form used by
+                // LLVM, which omits the preceding fence if no write operation is performed.
                 unsafe {
                     asm!(
                             concat!("ldrex", $suffix, " {out}, [{dst}]"),      // atomic { out = *dst; EXCLUSIVE = dst }


compiler-builtins/src/thumbv6k.rs
```diff
@@ -0,0 +1,326 @@
...
+        intrinsics! {
+            pub unsafe extern "C" fn $name(ptr: *mut $ty, oldval: $ty, newval: $ty) -> $ty {
+                // FIXME(safety): preconditions review needed
+                unsafe { <$ty as Atomic>::cmpxchg(ptr, oldval, newval) }
+            }
+        }
+    };
+}
+
+atomic_rmw!(@old __sync_fetch_and_add_1, u8, |a: u8, b: u8| a.wrapping_add(b));
```
taiki-e: The macro calls from here to L319 are the same as in `arm_linux.rs`  (the definitions of the called macros are different), so I think it would be preferable to extract them into a separate file.

taiki-e: Done in af8d4af012a99939f971a1e159fea4b7faa7a68d.

compiler-builtins/src/thumbv6k.rs
```diff
@@ -0,0 +1,200 @@
...
+//
+// Armv6 does not support DMB instruction, so use use special instruction equivalent to it.
+//
+// Refs: https://developer.arm.com/documentation/ddi0360/f/control-coprocessor-cp15/register-descriptions/c7--cache-operations-register
+macro_rules! dmb {
+    () => {
+        "mcr p15, #0, {zero}, c7, c10, #5"
+    };
+}
+macro_rules! asm_use_dmb {
```
thejpster: I think this macro is just an asm block, but with macro expansion (and an extra `{zero}` register which I guess is useful somewhere). But it could probably do with a comment to explain all that because it's really non-obvious. Especially as this macro doesn't have anything to do with `dmb` (other than it's written as a macro that needs to be expanded).

taiki-e: This was originally a helper for [writing code that required no changes other than using the Data Memory Barrier (DMB instruction or equivalent instructions)](https://github.com/taiki-e/atomic-maybe-uninit/blob/f98d8ab43c83e49bc3e030d8c3cb8ea1c5f96cab/src/arch/arm.rs#L187-L203).

In this PR, it is indeed unclear, so it can be deleted.

taiki-e: Removed `asm_use_dmb!` macro and renamed `dmb!` macro to `cp15_barrier!` (to more clearly reflect what it does) in https://github.com/rust-lang/compiler-builtins/pull/1050/commits/a1a540188fd4b499b51fbde9f36599d3fd9089aa.

thejpster:
This looks OK to me, but it's a bit beyond my Arm experience.

compiler-builtins/src/sync/arm_thumb_shared.rs
tgross35, line 1: Nit: could you name this `arm_thumb_sync_builtins` or similar?

compiler-builtins/src/thumbv6k.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+trait Atomic: Copy + Eq {
+    unsafe fn load_relaxed(src: *const Self) -> Self;
+    unsafe fn cmpxchg(dst: *mut Self, current: Self, new: Self) -> Self;
+}
+
+#[rustfmt::skip]
+macro_rules! atomic {
```
tgross35: I don't think rustfmt does any changes here

compiler-builtins/src/sync/thumbv6k.rs
```diff
@@ -0,0 +1,199 @@
...
+// is theoretically possible for the compiler to insert operations that might
+// clear the reservation between LL and SC. See https://github.com/taiki-e/portable-atomic/blob/58ef7f27c9e20da4cc1ef0abf8b8ce9ac5219ec3/src/imp/atomic128/aarch64.rs#L44-L55
+// for more details.
+unsafe fn atomic_rmw<T: Atomic, F: Fn(T) -> T, G: Fn(T, T) -> T>(ptr: *mut T, f: F, g: G) -> T {
+    loop {
+        // FIXME(safety): preconditions review needed
+        let curval = unsafe { T::load_relaxed(ptr) };
+        let newval = f(curval);
+        // FIXME(safety): preconditions review needed
+        if unsafe { T::cmpxchg(ptr, curval, newval) } == curval {
```
tgross35: Would you mind filling these in? Most of them are probably just related to pointer validity.

compiler-builtins/src/sync/arm_thumb_shared.rs
tgross35: Actually, I think we may as well group this by functionality like we have for other areas, something like:

```text
src/sync/
    arm_linux.rs
    thumbv6k.rs
    arm_thumb_shared.rs // this file
```

I'll move the aarch64 builtins there too.

compiler-builtins/src/thumbv6k.rs
```diff
@@ -0,0 +1,199 @@
...
+                            "beq 3f",                                          // if Z == 1 { jump 'success }
+                            concat!("ldrex", $suffix, " {out}, [{dst}]"),      // atomic { out = *dst; EXCLUSIVE = dst }
+                            "cmp {out}, {old}",                                // if out == old { Z = 1 } else { Z = 0 }
+                            "beq 2b",                                          // if Z == 1 { jump 'retry }
+                        "3:", // 'cmp-fail | 'success:
+                            cp15_barrier!(),                                            // fence
+                        dst = in(reg) dst,
+                        // Note: this cast must be a zero-extend since loaded value
+                        // which compared to it is zero-extended.
+                        old = in(reg) old as u32,
```
tgross35: `u32::from(old)`, since this should never truncate

compiler-builtins/src/sync/thumbv6k.rs
```diff
@@ -0,0 +1,199 @@
...
+                            "cmp {out}, {old}",                                // if out == old { Z = 1 } else { Z = 0 }
+                            "bne 3f",                                          // if Z == 0 { jump 'cmp-fail }
+                            cp15_barrier!(),                                            // fence
+                        "2:", // 'retry:
+                            concat!("strex", $suffix, " {r}, {new}, [{dst}]"), // atomic { if EXCLUSIVE == dst { *dst = new; r = 0 } else { r = 1 }; EXCLUSIVE = None }
+                            "cmp {r}, #0",                                     // if r == 0 { Z = 1 } else { Z = 0 }
+                            "beq 3f",                                          // if Z == 1 { jump 'success }
+                            concat!("ldrex", $suffix, " {out}, [{dst}]"),      // atomic { out = *dst; EXCLUSIVE = dst }
+                            "cmp {out}, {old}",                                // if out == old { Z = 1 } else { Z = 0 }
+                            "beq 2b",                                          // if Z == 1 { jump 'retry }
```
tgross35, line 85: For my own understanding: why is the barrier needed between the first ldrex/strex, but not between subsequent l/s pairs?

compiler-builtins/src/sync/thumbv6k.rs
```diff
@@ -0,0 +1,199 @@
...
+                unsafe { <$base as Atomic>::load_relaxed(src.cast::<$base>()) as Self }
+            }
+            #[inline]
+            unsafe fn cmpxchg(dst: *mut Self, old: Self, new: Self) -> Self {
+                // SAFETY: the caller must uphold the safety contract.
+                // casts are okay because $ty and $base implement the same layout.
+                unsafe {
+                    <$base as Atomic>::cmpxchg(dst.cast::<$base>(), old as $base, new as $base)
+                        as Self
+                }
```
tgross35, line 137: `.cast_{un,}signed()` is stable now and would make the `as` casts more clear

compiler-builtins/src/thumbv6k.rs
tgross35: I don't think rustfmt does any changes here

tgross35 (commented):
Looks pretty good to me, thanks for the patch. This should wait until one of the compiler leads approves the target PR, though.

tgross35:
> This PR has been tested on QEMU 10.2.0 using patched compiler-builtins and core that applied the changes in this PR and [rust-lang/rust#150138](https://github.com/rust-lang/rust/pull/150138) and the [portable-atomic no-std test suite](https://github.com/taiki-e/portable-atomic/tree/HEAD/tests/no-std-qemu) (can be run with `./tools/no-std.sh thumbv6-none-eabi` on that repo) which tests wrappers around `core::sync::atomic`.

If you get the chance at some point, I would be rather interested in having some of your atomic test suite here. Especially the part you mention since we don't have any no-std tests set up.

> (Note that Armv6k also supports 64-bit atomic instructions, but they are skipped here. This is because there is no corresponding code in `arm_linux.rs` (since the kernel requirements increased in 1.64, it may be possible to implement 64-bit atomics there as well. see also [taiki-e/portable-atomic#82](https://github.com/taiki-e/portable-atomic/pull/82)), the code becomes more complex than for 32-bit and smaller atomics.)

Assuming this would allow us to set `target_has_atomic = "64"`, I think that would be worth it. Just not pressing, of course.

thejpster:
@davidtwco said “LGTM” on the target but the target is blocked waiting for this PR. But it seems this PR is blocked waiting for the target?

@taiki-e - should we still do this one first?

compiler-builtins/src/sync/thumbv6k.rs
taiki-e: This (omit the preceding fence if no write operation is performed) is what LLVM does (https://godbolt.org/z/G7jGEGz6q), but I can rewrite to always emits two fences.

taiki-e:
@thejpster

> But it seems this PR is blocked waiting for the target?

This PR is fine without the target. (See the CI change using a custom target.)

(However, the target or patched `core` is required to run the test suite not included in this PR.)

compiler-builtins/src/sync/arm_thumb_shared.rs
taiki-e: Done  in 5ccc285.

compiler-builtins/src/thumbv6k.rs
taiki-e: Done in 83d096b.

compiler-builtins/src/sync/thumbv6k.rs
taiki-e: Done in 83d096b.

compiler-builtins/src/thumbv6k.rs
taiki-e: Done in 83d096b.

compiler-builtins/src/sync/thumbv6k.rs
taiki-e: Filled in e420b6b2b3b2816dd574cc047efba73a362fca65.

compiler-builtins/src/sync/thumbv6k.rs
tgross35: I don't know enough about the target to know what is better here, so it's your call. Either way, mind adding a comment?

taiki-e: Added a commit in https://github.com/rust-lang/compiler-builtins/pull/1050/commits/2a6a68c85c09fc51445de5957e2b11a1710a7077.

taiki-e:
> I'd still mildly prefer if [rust-lang/rust#150138](https://github.com/rust-lang/rust/pull/150138) merges first so this PR can test against the target as-added without the back-and-forth. But if a few days go by without any movement there, I'll go ahead and apply this.

I think it is easier to understand/use if the target consistently provides atomics regardless of the compiler version from the user's point of view. (In cases where atomic operations are missing, it may be necessary to enable the [specific optional feature to enable atomic](https://github.com/tokio-rs/bytes?tab=readme-ov-file#no_std-support).)


(Also, to run tests, it's preferable that atomics are enabled for the target (manually calling the sync built-in is annoying), so I guess adding no-std tests will need multiple PRs on the compiler-builtins side anyway.)

tgross35:
Fair enough, applied now. I probably won't do a sync until after https://github.com/rust-lang/compiler-builtins/pull/1061 (unless that takes a long time) since that may have some UB in atomic fixes. (Bit of a weird issue, any idea if it's the sort of thing your testsuite would have caught?)

PR #1053, opened 2026-01-02
Fix a typo in `libm::Libm::roundeven`

the-ssd:
I use `Libm` helper in a macro, to which f32 and f64 is passed. Because of this typo, `Libm::roundeven` can't be used.


the-ssd:
Thank you too!

Issue #1054, opened 2026-01-04
CI: support ci job on riscv64 hardware


yuzibo:
As we have discussed [this]([#t-infra > We're eager to provide the community with RISC-V CI infrastr @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/We're.20eager.20to.20provide.20the.20community.20with.20RISC-V.20CI.20infrastr/near/566193554)) on zulipchat channel, let's focus on some details on how to push this.

yuzibo:
The main steps for the job that I refer to the [main.yml](https://github.com/rust-lang/compiler-builtins/blob/main/.github/workflows/main.yaml).

They are:

* Update/Download Code
* Print runner information
* Configure Rustup(install rustup)
* Download/Cache compiler-rt
* Download musl source
* Verify API list
* Run locally test

Because we have run the Job in Docker, so I run `./ci/run.sh`  instead of `./ci/run-docker.sh` here.

Is it enough for one CI job on riscv64? Maybe it is good to add `benchmarks` test here, but I am not sure about this.

tgross35:
Those look like the important steps! Really `./ci/run.sh` is the important bit, how you get there shouldn't matter too much.

The icount benchmarks would be nice eventually since it lets us make sure we don't regress performance on risc-v. But they're probably a bit trickier to set up so that can come later.

Happy to accept a PR here when you're ready.

tgross35:
> The icount benchmarks would be nice eventually since it lets us make sure we don't regress performance on risc-v. But they're probably a bit trickier to set up so that can come later.

More specifically, they should be pretty easy to actually run since it's just getting the right dependencies then running the script. But doing anything with the results is probably trickier: currently we upload the results as a CI artifact, then download the artifact from the latest run on `main` to compare against and fail the job if there are regressions. That part isn't strictly required, though.

tgross35:
Another note: if you can install nextest on the system like we do in CI, that can speed up test runs quite a bit (though they're not much slower than our current slowest jobs). `run.sh` detects it automatically.

yuzibo:
Many thanks for kindly guiding. I have proposed one commit to support the CI job on riscv64 hardware.

Highlight:

* Self hosted runner fully, no longer interacting with Jenkins;
* Multiple nodes for the CI job.

This is my [experimental](https://github.com/yuzibo/compiler-builtins/actions).

yuzibo:
[...]

PR #1059, opened 2026-01-09
ci: Test on riscv64 using a self-hosted runner

yuzibo:
Closes: https://github.com/rust-lang/compiler-builtins/issues/1054

ci: test-libm

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -44,6 +44,9 @@ jobs:
   test:
     name: Build and test
     timeout-minutes: 60
+    # NOTE: self-hosted riscv64 runners are experimental and may be flaky.
+    # Do not block CI on failures from this platform for now.
+    continue-on-error: ${{ contains(matrix.os, 'self-hosted') }}
     strategy:
       fail-fast: false
       matrix:
@@ -74,6 +77,8 @@ jobs:
           os: ubuntu-24.04
         - target: powerpc64le-unknown-linux-gnu
           os: ubuntu-24.04-ppc64le
+        - target: riscv64gc-unknown-linux-gnu
+          os: ["self-hosted", "linux", "riscv64"]
         - target: riscv64gc-unknown-linux-gnu
           os: ubuntu-24.04
         - target: s390x-unknown-linux-gnu
@@ -173,7 +178,7 @@ jobs:
       shell: bash
 
     - name: Verify API list
-      if: matrix.os == 'ubuntu-24.04'
+      if: matrix.os == 'ubuntu-24.04' || contains(matrix.os, 'self-hosted')
       run: python3 etc/update-api-list.py --check
 
     # Non-linux tests just use our raw script


yuzibo:
To address the [issue](https://github.com/rust-lang/compiler-builtins/issues/1054).

I left some comments on there, please review it also

.github/workflows/main.yaml
```diff
@@ -194,6 +194,52 @@ jobs:
         rm -rf /tmp/.buildx-cache
         mv /tmp/.buildx-cache-new /tmp/.buildx-cache
 
+  test-riscv64gc-native:
```
yuzibo: This is the best solution I can think of for easy maintenance of this separate job. Although I'm not satisfied with the riscv64 job occupying a separate block. It would be better under one block like `Running on self hosted runner` which has the same level with `matrix` block.?

.github/workflows/main.yaml
```diff
@@ -194,6 +194,52 @@ jobs:
...
+    name: Build and test (riscv64gc-unknown-linux-gnu self-hosted, experimental)
```
yuzibo: Let us focus on riscv64gc-unknown-linux-gnu first and other target triple we will to also support, like rva23 triple.

The experimental tag as it is.

.github/workflows/main.yaml
```diff
@@ -194,6 +194,52 @@ jobs:
...
+        shell: bash
+        run: |
+          sed -i 's|https://github.com|https://community-ci.openruyi.cn|g' ./ci/download-compiler-rt.sh
+          sed -i 's|llvm-project-rustc-${rust_llvm_version}|llvm-project|g' ./ci/download-compiler-rt.sh
+          ./ci/download-compiler-rt.sh
+          echo "RUST_COMPILER_RT_ROOT=$(realpath ./compiler-rt)" >> "$GITHUB_ENV"
+
+    - name: Download musl source
+        run: |
+          sed -i 's|https://github.com/kraj/musl.git|https://community-ci.openruyi.cn/others/kraj-musl.git|g' ./ci/update-musl.sh
```
yuzibo: I hope these `sed` usages can be accepted in order to speed up the whole workflow. And this clearly demonstrates what we did.

.github/workflows/main.yaml
tgross35: I think you could probably add this under the `test` block pretty easily since a lot of the jobs are shared. Setting `os: [self-hosted, linux, riscv64]` should work, and for any step that doesn't need to run you can add/update `if: !contains(matrix.os, "self-hosted")`

.github/workflows/main.yaml
tgross35: Just starting with one is great 👍

.github/workflows/main.yaml
tgross35: Is the speed difference that significant? I'd rather not use the mirrors if possible to reduce the possible points of failure or things that could get out of sync.

yuzibo: Yes, this is one of the feasible solutions we can implement.

On the self-hosted riscv64 runners, outbound access to some upstream services can be intermittently unstable due to external network factors, which may occasionally affect CI reliability before the actual build or tests start from my [experimental](https://github.com/yuzibo/compiler-builtins/actions/runs/20879960063/job/59995883014). In fact,localized mirror services were included as an initial goal of this project to ensure the stability of CI.

Regarding the risk of mirrors getting out of sync: these mirrors are actively maintained and have a daily synchronization mechanism to track upstream changes, along with basic health checks to ensure they remain usable. This helps reduce the risk of silent divergence.

We do not publish our mirror service in case other users abuse it to affect the stability of our CI, if necessary, we can publish it after our internal assessment.

Using these [Prepare mirrored sources](https://github.com/rust-lang/compiler-builtins/pull/1059/files#diff-71cabc4177e41ea8f15a89eb65398fa8187739f68a2762706d84885cd8b2ddc3R160) with `sed` is a more transparent approach than using a cached solution, which increases complexity. But for r-l/r, we have to depend on lots of cache strategy, If we get to that point.

.github/workflows/main.yaml
```diff
@@ -102,7 +106,9 @@ jobs:
         - target: x86_64-pc-windows-gnu
           os: windows-2025
           channel: nightly-x86_64-gnu
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.os == 'self-hosted-riscv64' && fromJSON('["self-hosted","linux","riscv64"]')
+               || matrix.os
+              }}
```
tgross35: I believe you could set `os: ["self-hosted","linux","riscv64"]`, then use `if: contains(matrix.os, "self-hosted")` in the below jobs to slightly simplify things and eliminate the `fromJSON`

yuzibo: ok, will do.

.github/workflows/main.yaml
tgross35: Confirmed with infra on Zulip (starting around here [#t-infra > RISC-V runner in compiler builtins @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/RISC-V.20runner.20in.20compiler.20builtins/near/580000812)) that we don't want to use these mirrors. It's fine to keep around while you're still testing, but please drop them at some point.

If there's some weird circumstances making these mirrors work better than GH, bring it up on Zulip so we can see if there's something we can do.

yuzibo:
We can add riscv64 runners first then merge this PR if no problem.

tgross35:
Mind rebasing so we get a fresh run with the more recent CI changes?

PR #1060, opened 2026-01-11
Remove `no-f16-f128`

tgross35:


tgross35:
This will go via upstream https://github.com/rust-lang/rust/pull/150968

PR #1061, opened 2026-01-12
Fix generated assembly for the `outline-atomics` feature in Apple AArch64 targets 

vini-fda:
**Context**

See https://github.com/rust-lang/rust/issues/149634#issuecomment-3727134152.

Most of the credit goes to https://github.com/tgross35/rust/commit/76a4adc3a28352912188622cfade5ae0b672bcbd.

**Summary**

The main fixes are (as per [Trevor's comment on the issue in rust-lang/rust](https://github.com/rust-lang/rust/issues/149634#issuecomment-3611270146)):

- Replacing semicolons (`;`) in the concatenated macros by newlines (`\n`)
- Properly accounting for Mach-O vs ELF AArch64 relocation specifier syntax
  - As a side note, while learning about these differences, I found [this blog post](https://maskray.me/blog/2025-03-16-relocation-generation-in-assemblers) by [MaskRay](https://github.com/MaskRay) quite useful
  
Additionally, I added some doc comments to explain the differences between Apple/Mach-O vs Linux/ELF syntax wrt to the relocation types.

### DIFF
--- a/compiler-builtins/src/aarch64_outline_atomics.rs
+++ b/compiler-builtins/src/aarch64_outline_atomics.rs
@@ -34,6 +34,19 @@ intrinsics! {
     }
 }
 
+/// Function to enable/disable LSE. To be used only for testing purposes.
+#[cfg(feature = "mangled-names")]
+pub unsafe fn set_have_lse_atomics(has_lse: bool) {
+    let lse_flag = if has_lse { 1 } else { 0 };
+    HAVE_LSE_ATOMICS.store(lse_flag, Ordering::Relaxed);
+}
+
+/// Function to obtain whether LSE is enabled or not. To be used only for testing purposes.
+#[cfg(feature = "mangled-names")]
+pub fn get_have_lse_atomics() -> bool {
+    HAVE_LSE_ATOMICS.load(Ordering::Relaxed) != 0
+}
+
 /// Translate a byte size to a Rust type.
 #[rustfmt::skip]
 macro_rules! int_ty {
--- a/builtins-test/tests/lse.rs
+++ b/builtins-test/tests/lse.rs
@@ -2,6 +2,45 @@
 #![feature(macro_metavar_expr_concat)]
 #![cfg(target_arch = "aarch64")]
 
+use std::sync::Mutex;
+
+use compiler_builtins::aarch64_outline_atomics::{get_have_lse_atomics, set_have_lse_atomics};
+use compiler_builtins::int::{Int, MinInt};
+use compiler_builtins::{foreach_bytes, foreach_ordering};
+
+#[track_caller]
+fn with_maybe_lse_atomics(use_lse: bool, f: impl FnOnce()) {
+    // Ensure tests run in parallel don't interleave global settings
+    static LOCK: Mutex<()> = Mutex::new(());
+    let _g = LOCK.lock().unwrap();
+    let old = get_have_lse_atomics();
+    // safety: as the caller of the unsafe fn `set_have_lse_atomics`, we
+    // have to ensure the CPU supports LSE. This is why we make this assertion.
+    if use_lse || old {
+        assert!(std::arch::is_aarch64_feature_detected!("lse"));
+    }
+    unsafe { set_have_lse_atomics(use_lse) };
+    f();
+    unsafe { set_have_lse_atomics(old) };
+}
+
+pub fn run_fuzz_tests_with_lse_variants<I: Int, F: Fn(I, I) + Copy>(n: u32, f: F)
+where
+    <I as MinInt>::Unsigned: Int,
+{
+    // We use `fuzz_2` because our subject function `f` requires two inputs
+    let test_fn = || {
+        builtins_test::fuzz_2(n, f);
+    };
+    // Always run without LSE
+    with_maybe_lse_atomics(false, test_fn);
+
+    // Conditionally run with LSE
+    if std::arch::is_aarch64_feature_detected!("lse") {
+        with_maybe_lse_atomics(true, test_fn);
+    }
+}
+
 /// Translate a byte size to a Rust type.
 macro int_ty {
     (1) => { u8 },
@@ -15,16 +54,17 @@ mod cas {
     pub(super) macro test($_ordering:ident, $bytes:tt, $name:ident) {
         #[test]
         fn $name() {
-            builtins_test::fuzz_2(10000, |expected: super::int_ty!($bytes), new| {
+            crate::run_fuzz_tests_with_lse_variants(10000, |expected: super::int_ty!($bytes), new| {
                 let mut target = expected.wrapping_add(10);
+                let ret: super::int_ty!($bytes) = unsafe {
+                    compiler_builtins::aarch64_outline_atomics::$name::$name(
+                        expected,
+                        new,
+                        &mut target,
+                    )
+                };
                 assert_eq!(
-                    unsafe {
-                        compiler_builtins::aarch64_outline_atomics::$name::$name(
-                            expected,
-                            new,
-                            &mut target,
-                        )
-                    },
+                    ret,
                     expected.wrapping_add(10),
                     "return value should always be the previous value",
                 );
@@ -35,15 +75,17 @@ mod cas {
                 );
 
                 target = expected;
+                let ret: super::int_ty!($bytes) = unsafe {
+                    compiler_builtins::aarch64_outline_atomics::$name::$name(
+                        expected,
+                        new,
+                        &mut target,
+                    )
+                };
                 assert_eq!(
-                    unsafe {
-                        compiler_builtins::aarch64_outline_atomics::$name::$name(
-                            expected,
-                            new,
-                            &mut target,
-                        )
-                    },
-                    expected
+                    ret,
+                    expected,
+                    "the new return value should always be the previous value (i.e. the first parameter passed to the function)",
                 );
                 assert_eq!(target, new, "should have updated target");
             });
@@ -59,16 +101,21 @@ mod swap {
     pub(super) macro test($_ordering:ident, $bytes:tt, $name:ident) {
         #[test]
         fn $name() {
-            builtins_test::fuzz_2(10000, |left: super::int_ty!($bytes), mut right| {
-                let orig_right = right;
-                assert_eq!(
-                    unsafe {
-                        compiler_builtins::aarch64_outline_atomics::$name::$name(left, &mut right)
-                    },
-                    orig_right
-                );
-                assert_eq!(left, right);
-            });
+            crate::run_fuzz_tests_with_lse_variants(
+                10000,
+                |left: super::int_ty!($bytes), mut right| {
+                    let orig_right = right;
+                    assert_eq!(
+                        unsafe {
+                            compiler_builtins::aarch64_outline_atomics::$name::$name(
+                                left, &mut right,
+                            )
+                        },
+                        orig_right
+                    );
+                    assert_eq!(left, right);
+                },
+            );
         }
     }
 }
@@ -80,7 +127,7 @@ macro_rules! test_op {
                 ($_ordering:ident, $bytes:tt, $name:ident) => {
                     #[test]
                     fn $name() {
-                        builtins_test::fuzz_2(10000, |old, val| {
+                        crate::run_fuzz_tests_with_lse_variants(10000, |old, val| {
                             let mut target = old;
                             let op: fn(super::int_ty!($bytes), super::int_ty!($bytes)) -> _ = $($op)*;
                             let expected = op(old, val);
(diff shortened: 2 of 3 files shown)


vini-fda:
Self-note: the tests in `lse.rs` are now up and running but many tests are failing. I'll check out the failing tests and try to fix as I go.

compiler-builtins/src/aarch64_outline_atomics.rs
```diff
@@ -135,18 +135,73 @@ macro_rules! stxp {
...
+        concat!($sym, "@PAGE")
+    };
+}
+
+#[cfg(target_vendor = "apple")]
+macro_rules! sym_off {
+    ($sym:literal) => {
+        concat!($sym, "@PAGEOFF")
+    };
+}
```
tgross35, line 204: This matches my understanding, but @madsmtm mind taking a second look here?

vini-fda:
@tgross35 I think the bug has to do with sign extensions. Take a look at this minimal `main.rs` file that reproduces the issue:

```rust
use std::sync::atomic::AtomicU8;
/// non-zero if the host supports LSE atomics.
/// This is hardcorded to 1 because Apple M-series macs have this enabled.
static HAVE_LSE_ATOMICS: AtomicU8 = AtomicU8::new(1);

/// Implementation of Compare-And-Swap for i8 for Apple aarch64 platforms, with "Acquire" ordering semantics.
///
/// Stores the `new` value into the atomic integer given by `ptr` if the currently stored value is the same as the value of `current`.
/// The return value is always the previously stored value. If it is equal to `current`, then the value was updated.
#[unsafe(naked)]
pub unsafe extern "C" fn compare_and_swap_acq(current: i8, new: i8, ptr: *mut i8) -> i8 {
    core::arch::naked_asm! {
        ".arch_extension lse",
        "adrp    x16, {have_lse}@PAGE",
        "ldrb    w16, [x16, {have_lse}@PAGEOFF]",
        "cbz     w16, 0f",
        "casab   w0, w1, [x2]",
        "ret",
        "0:",
        "uxtb    w16, w0",
        "1:",
        "ldaxrb  w0, [x2]",
        "cmp     w0, w16",
        "bne    2f",
        "stxrb   w17, w1, [x2]",
        "cbnz    w17, 0b",
        "2:",
        "ret",
        have_lse = sym HAVE_LSE_ATOMICS,
    };
}

fn main() {
    let expected: i8 = -1;
    let new: i8 = 0;
    let mut target = expected.wrapping_add(10);
    // begin test
    let _ = unsafe { compare_and_swap_acq(expected, new, &mut target) };
    target = expected;
    let ret: i8 = unsafe { compare_and_swap_acq(expected, new, &mut target) };
    assert_eq!(
        ret, expected,
        "the new return value should always be the previous value (i.e. the first parameter passed to the function), ret = {ret:?}, expected = {expected:?}, new = {new:?}"
    );
}
```

Executing it on a M-series mac with `cargo run --release`, it runs into the same issue:

```
assertion `left == right` failed
  left: -1
 right: -1
 ```
 Probably because the byte load instructions (`ldrb`, `ldaxrb`) and the LSE atomic instruction (`casab`) zero-extend the loaded byte to fill the 32-bit register `w0` (you can confirm this in the ARM docs for [ldarb](https://developer.arm.com/documentation/dui0801/l/A64-Data-Transfer-Instructions/LDRB--immediate-), [ldaxrb](https://developer.arm.com/documentation/ddi0602/2025-12/Base-Instructions/LDAXRB--Load-acquire-exclusive-register-byte-?lang=en) and [casab](https://developer.arm.com/documentation/111181/2025-09_ASL1/Base-Instructions/CASB--CASAB--CASALB--CASLB--Compare-and-swap-byte-in-memory-?lang=en)). So, when it loads `-1` (binary `0xFF`):
- **Expected** (for signed `i8`): `0xFFFFFFFF` (sign-extended)
- **Actual**: `0x000000FF` (zero-extended, equals 255 unsigned)

If so, I think the fix is to sign extend with [sxtb](https://developer.arm.com/documentation/dui0801/l/A64-General-Instructions/SXTB--A64-):
 
```diff
 #[unsafe(naked)]
 pub unsafe extern "C" fn compare_and_swap_acq(current: i8, new: i8, ptr: *mut i8) -> i8 {
     core::arch::naked_asm! {
         ".arch_extension lse",
         "adrp    x16, {have_lse}@PAGE",
         "ldrb    w16, [x16, {have_lse}@PAGEOFF]",
         "cbz     w16, 0f",
         "casab   w0, w1, [x2]",
+        "sxtb    w0, w0",  // Sign-extend byte to 32-bit for correct i8 return value
         "ret",
         "0:",
         "uxtb    w16, w0",
         "1:",
         "ldaxrb  w0, [x2]",
         "cmp     w0, w16",
         "bne    2f",
         "stxrb   w17, w1, [x2]",
         "cbnz    w17, 0b",
         "2:",
+        "sxtb    w0, w0",  // Sign-extend byte to 32-bit for correct i8 return value
         "ret",
         have_lse = sym HAVE_LSE_ATOMICS,
     };
 }
```

I just pushed a change to see if this fixes it. Locally on my laptop it works, but I'll wait on the CI results to be sure.

vini-fda:
I just noticed I'd forgotten to add the sign extension to the LSE path as well. The new commit [45f21ff](https://github.com/rust-lang/compiler-builtins/pull/1061/commits/45f21ff88775892ab3d83a986650a07fe10f2708) adds that fix. Concerningly, before [45f21ff](https://github.com/rust-lang/compiler-builtins/pull/1061/commits/45f21ff88775892ab3d83a986650a07fe10f2708) the CI checks passed (even though the corresponding `main.rs` code from the previous comment errors out on my mac without the LSE path fix). It would be nice if the CI checks covered that path as well...

compiler-builtins/src/aarch64_outline_atomics.rs
```diff
@@ -135,17 +135,92 @@ macro_rules! stxp {
...
+    (1) => {
+        concat!("sxtb ", reg!(1, 0), ", ", reg!(1, 0))
+    };
+    (2) => {
+        concat!("sxth ", reg!(2, 0), ", ", reg!(2, 0))
+    };
+    (4) => { "" };
+    (8) => { "" };
+    (16) => { "" };
+}
```
tgross35: To make this more flexible for the future (also more obvious to read), could you update these to take a second `$num:literal` param that gets passed to `reg`? Rather than always using 0.

compiler-builtins/src/aarch64_outline_atomics.rs
```diff
@@ -135,17 +135,92 @@ macro_rules! stxp {
...
+            concat!("ldrb    w16, [x16, ", sym_off!("{have_lse}"), "]\n"),
+            "cbz     w16, 8f\n",
             // LSE_OP  s(reg),* [$mem]
-            concat!(lse!($op, $ordering, $bytes), $( " ", reg!($bytes, $reg), ", " ,)* "[", stringify!($mem), "]; ",),
-            "ret; ",
+            concat!(lse!($op, $ordering, $bytes), $( " ", reg!($bytes, $reg), ", " ,)* "[", stringify!($mem), "]\n",),
+            "ret\n",
+            // SXTB s(0), s(0)
+            concat!(sign_extend!($bytes), "\n"),
             "8:"
```
tgross35: Should this go before the `ret`?

tgross35:
Amanieu knows all of this much better than I do, requested a review.

I wonder if the missing sign extension might happen to be the issue mentioned at https://github.com/rust-lang/rust/pull/144938#issuecomment-3781057414. @zmodem do you have more info about the crash?

LLVM doesn't seem to sign extend either, may be worth opening an issue there https://github.com/llvm/llvm-project/blob/524589119afd59525ab1f124a1538f0311d608de/compiler-rt/lib/builtins/aarch64/lse.S.

> It would be nice if the CI checks covered that path as well...

Would this be possible to add?

tgross35:
I'm still not sure I understand the difference here; any idea why the failure only showed up on MacOS?

> > It would be nice if the CI checks covered that path as well...
> 
> Would this be possible to add?

Thinking about it more, there could probably be a hook for this that's only used for testing:

```rust
#[cfg(feature = "mangled-names")]
pub unsafe fn set_have_lse_atomics(has_lse: bool) { /* ... */ }
#[cfg(feature = "mangled-names")]
pub fn get_have_lse_atomics() -> bool { /* ... */ }
```

Then in the test file:

```rust
#[track_caller]
fn with_maybe_lse_atomics(use_lse: bool, f: impl FnOnce()) {
    // Ensure tests run in parallel don't interleave global settings
    static LOCK: Mutex<()> = Mutex::new(());
    let _g = LOCK.lock().unwrap();
    let old = get_have_lse_atomics();
    if use_lse || old { assert!(is_aarch64_feature_enabled("lse"); }
    unsafe { set_have_lse_atomics(use_lse) };
    f();
    unsafe { set_have_lse_atomics(old) };
}

zmodem:
> I wonder if the missing sign extension might happen to be the issue mentioned at rust-lang/rust#144938 (comment). @zmodem do you have more info about the crash?

Thanks for the cc. I'm afraid I don't have much details yet. I filed https://github.com/rust-lang/rust/issues/151486 to track it.

taiki-e:
> any idea why the failure only showed up on MacOS?

I guess that is because Apple's calling conventions about argument passing differ from the standard ABI. From [Apple docs](https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Pass-arguments-to-functions-correctly):

> The caller of a function is responsible for signing or zero-extending any argument with fewer than 32 bits. The standard ABI expects the callee to sign or zero-extend those arguments.

(Although it's not explicitly stated regarding the return value, I guess extension of the return value is callee's responsibility  because it allows to avoid extension when returning the value as-is and using it.)

compiler-builtins/src/aarch64_outline_atomics.rs
vini-fda: Yes! thanks for catching that. The new commit should have this fixed.

compiler-builtins/src/aarch64_outline_atomics.rs
vini-fda: Commit https://github.com/rust-lang/compiler-builtins/pull/1061/commits/8285fe118bf05e4543fa3e66cf2eb7f413f87713 should improve on this.

vini-fda:
@taiki-e I'm not sure I understand your explanation. From the Apple quote:

> The caller of a function is responsible for signing or zero-extending any argument with fewer than 32 bits. The standard ABI expects the callee to sign or zero-extend those arguments.

Let's take the minimal example, where the `main` function is the caller and the `compare_and_swap_acq` is the callee.

- In the Apple convention, the caller (`main`) should be responsible for sign-extending, not the callee (the custom naked function `compare_and_swap_acq`)
- In the standard convention, it's the other way around: the callee (our custom function) should perform that operation

So from that quote, I'd imagine the tests would at best only _pass_ in Apple platforms (or not pass in any AArch64 platform), not the other way around. But in reality it is the other way around! Or maybe I'm missing something?

taiki-e:
@vini-fda


> So from that quote, I'd imagine the tests would at best only _pass_ in Apple platforms (or not pass in any AArch64 platform), not the other way around. But in reality it is the other way around! Or maybe I'm missing something?

Under my understanding, in our usage, it doesn't matter which state the arguments are passed in. Atomic, bit-op, and wrapping arithmetic instructions will work regardless of whether extension is applied. Comparisons also work thanks to zero extension.

The issue lies in how the ABI expects the return value to be handled.
- Although [not explicitly stated in the standard ABI](https://github.com/ARM-software/abi-aa/blob/2025Q1/aapcs64/aapcs64.rst#parameter-passing), extensions should not be expected. Otherwise, functions that just return input values would also require adding extension processing.
- In Apple's calling conventions, argument extension is handled by the caller, so the return value is properly extended without additional processing when just returning input value. If return value extension is not callee's responsibility, the caller must handle extension for both arguments and return values, which is inefficient. Therefore, (although [not explicitly stated) return value extension is likely the callee's responsibility.

IIRC values obtained from atomic instructions are usually zero-extended, so Apple targets would likely require sign extension like this PR does. (And the absence of sign extension should be irrelevant for non-Apple targets.)

Amanieu:
Always sign extending isn't correct: this should only be performed if the integer type being passed as an argument or returned is a signed integer. I believe that for all of these intrinsics, the integer types are unsigned, in which case we just need to ensure that the result is zero-extended. This is already the case from the `ldrb`/`ldrh`, so no additional instruction is needed.

Amanieu:
I believe the correct fix is to change the function declarations to use unsigned integer types instead of signed ones, since that's what LLVM's backend is expecting when calling these.

vini-fda:
@Amanieu thanks for the feedback, I wasn't aware these functions expected unsigned types. Commit https://github.com/rust-lang/compiler-builtins/pull/1061/commits/66b43e2d92c27072019e8d7b319acf1cd3754780 should have fixed that. I reverted the sign-extension shenanigans as well.

@tgross35 thanks for the idea on enabling LSE for the tests. Commit https://github.com/rust-lang/compiler-builtins/commit/b1019f58fc5527dc0028499fdc312c79adab3634 now makes the tests more comprehensive. Hopefully this works :)

tgross35:
The changes here LGTM but I'd like @Amanieu and/or @taiki-e to approve. I mentioned on zulip but this should get a bit of history cleanup as well.

Amanieu (approved):

vini-fda:
@tgross35 I did a cleanup of the git history, should be good now.

Also I made a mistake of merging main by clicking on the button in the GItHub UI and Rustbot screamed at me for that :laughing: but I think I managed to fix it.

PR #1062, opened 2026-01-23
Set `codegen-units=1` for benchmarks

quaternic:
This should remove some of the nondeterminism in benchmark results observed in rust-lang/compiler-builtins#935.

ci: allow-regressions


tgross35:
Thanks!

tgross35:
@quaternic mentioned that there is some kind of inlining with the default CGU count that was acting unusually

PR #1063, opened 2026-01-23
WIP: Use BTI c in aarch64_outline_atomic.rs

taiki-e:
Potential fix for https://github.com/rust-lang/rust/issues/151486.

TODO: 
- Does this approach work in naked_asm?
- How to test this on CI? Both [linux](https://github.com/taiki-e/checkout-action/actions/runs/21269456117/job/61216212986#step:2:19) and [macos](https://github.com/taiki-e/checkout-action/actions/runs/21269456117/job/61216212983#step:3:51) runner by GitHub doesn't seem to support BTI.

### DIFF
--- a/compiler-builtins/src/aarch64_outline_atomics.rs
+++ b/compiler-builtins/src/aarch64_outline_atomics.rs
@@ -401,3 +401,28 @@ foreach_ldadd!(add);
 foreach_ldclr!(and);
 foreach_ldeor!(xor);
 foreach_ldset!(or);
+
+core::arch::global_asm!(
+    "
+    .pushsection .note.gnu.property, \"a\"
+    .balign 8
+    .word 4
+    .word 0x10
+    .word 0x5 // NT_GNU_PROPERTY_TYPE_0
+    .asciz \"GNU\"
+    .word 0xc0000000 // GNU_PROPERTY_AARCH64_FEATURE_1_AND
+    .word 4
+    ",
+    #[cfg(all(branch_protection = "bti", branch_protection = "pac-ret"))]
+    ".word 3",
+    #[cfg(all(not(branch_protection = "bti"), branch_protection = "pac-ret"))]
+    ".word 2",
+    #[cfg(all(branch_protection = "bti", not(branch_protection = "pac-ret")))]
+    ".word 1",
+    #[cfg(all(not(branch_protection = "bti"), not(branch_protection = "pac-ret")))]
+    ".word 0",
+    "
+    .word 0
+    .popsection
+    "
+);
--- a/compiler-builtins/src/aarch64_outline_atomics.rs
+++ b/compiler-builtins/src/aarch64_outline_atomics.rs
@@ -181,6 +181,7 @@ macro_rules! compare_and_swap {
             ) -> int_ty!($bytes) {
                 // We can't use `AtomicI8::compare_and_swap`; we *are* compare_and_swap.
                 core::arch::naked_asm! {
+                    "hint #34", // BTI c
                     // CAS    s(0), s(1), [x2]; if LSE supported.
                     try_lse_op!("cas", $ordering, $bytes, 0, 1, [x2]),
                     // UXT    s(tmp0), s(0)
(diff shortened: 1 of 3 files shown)


tgross35:
Thanks for jumping on this

> How to test this on CI? Both [linux](https://github.com/taiki-e/checkout-action/actions/runs/21269456117/job/61216212986#step:2:19) and [macos](https://github.com/taiki-e/checkout-action/actions/runs/21269456117/job/61216212983#step:3:51) runner by GitHub doesn't seem to support BTI.

@rust-lang/project-exploit-mitigations any chance you have some testing ideas?

Does BTI work in qemu-user by chance?

zmodem:
grep'ing compiler-builtins, I don't see any mention of bti at all. Maybe instead of adding `hint #34` to each function manually, there should be some general facility which adds BTI landingpads to manually generated functions when the feature is enabled.

taiki-e:
For normal Rust functions, the BTI c should be inserted by the compiler as needed. For global assemblies, the BTI c should be explicitly written within the assembly as needed.

However, it is unclear whose responsibility it should be for naked functions (what we use for outline_atomic). At least currently, the naked function does nothing, but -Z branch-protection is unstable so we can change its behavior.
cc @Amanieu @folkertdev

compiler-builtins/src/aarch64_outline_atomics.rs
```diff
@@ -397,3 +401,28 @@ foreach_ldadd!(add);
...
+    .word 4
+    ",
+    #[cfg(all(branch_protection = "bti", branch_protection = "pac-ret"))]
+    ".word 3",
+    #[cfg(all(not(branch_protection = "bti"), branch_protection = "pac-ret"))]
+    ".word 2",
+    #[cfg(all(branch_protection = "bti", not(branch_protection = "pac-ret")))]
+    ".word 1",
+    #[cfg(all(not(branch_protection = "bti"), not(branch_protection = "pac-ret")))]
+    ".word 0",
```
folkertdev, line 423: I think `cfg_select` would be nicer here

```suggestion
    cfg_select! (
        all(branch_protection = "bti", branch_protection = "pac-ret") => ".word 3",
        branch_protection = "pac-ret" => ".word 2",
        branch_protection = "bti" => ".word 1",
        _ => ".word 0",
    ),
```

folkertdev:
I'm not that familiar with how this works, but generally the aim is to make naked functions "just work". So I'd say naked functions should support the flag.

zmodem:
There was some discussion about naked functions vs. bti on the llvm side: https://github.com/llvm/llvm-project/issues/56369 but it didn't conclude. I'm not actually sure what the current behavior is.

folkertdev:
Rust naked functions are entirely distinct from LLVM naked functions. We lower them to a block of global assembly + an `extern` block defining the rust symbol ourselves. So, we are free to make our own decisions here. So far, this flag was not considered, but it seems like we should.

Amanieu:
I would actually prefer *not* having a BTI automatically inserted at the start of naked functions. The main purpose of naked functions is to support custom calling conventions which cannot be otherwise supported by the Rust compiler. In those cases we don't want to mess with what the user is intending for these functions to contain.

folkertdev:
Ah, I misunderstood: this actually adds instructions. Then I agree that we should not add that. And then `-Z branch-protection` should provide a way (via `cfg` maybe?) for the program to know what mode it's being compiled in.

Amanieu:
Is there any reason these functions need to be naked at all? They don't use a custom calling convention and could just be implemented using plain `asm!`. This would avoid any issues related to branch protection.

tgross35:
Did a small sanity check and I think inline asm should be reasonably feasible https://rust.godbolt.org/z/s38z6a6jP. Might be a better way to still keep the early `ret` rather than jumping?

I was originally a bit concerned about paying for frame pointers, which used to be a bit of a problem with fma. But it looks like that has been resolved with https://github.com/rust-lang/rust/pull/141800.

Amanieu:
You could write the LSE check in normal rust code and have 2 separate asm blocks in different if branches.

tgross35:
@taiki-e are you still interested in working on this switching to an inline asm version? No pressure of course.

taiki-e:
I'm sorry, but I don't have the bandwidth to work on this further at this time.

Closing in favor of asm! approach.

PR #1066, opened 2026-01-31
hexagon: Make `fma` label local to avoid symbol collision

androm3da:
The `fma:` label in dffma.s was being exported as a global symbol causing a "symbol 'fma' is already defined" error when linking with libm's `fma` function.

Unfortunately rust-lang/compiler-builtins#682 removed `.global fma` but didn't address the implicit global export of the label itself.

    --- old.txt     2026-01-30 20:31:37.265844316 -0600
    +++ new.txt     2026-01-30 20:31:46.531950264 -0600
    @@ -1,4 +1,3 @@
    -00000000 t fma
     00000000 T __hexagon_fmadf4
     00000000 T __hexagon_fmadf5
     00000000 T __qdsp_fmadf5

### DIFF
--- a/compiler-builtins/src/hexagon/dffma.s
+++ b/compiler-builtins/src/hexagon/dffma.s
@@ -7,7 +7,7 @@
  .p2align 5
 __hexagon_fmadf4:
 __hexagon_fmadf5:
-fma:
+.Lfma:
  {
   p0 = dfclass(r1:0,#2)
   p0 = dfclass(r3:2,#2)
@@ -400,7 +400,7 @@ fma:
   r3:2 = insert(r11:10,#63,#0)
   r1 -= asl(r28,#20)
  }
- jump fma
+ jump .Lfma
 
 .Lfma_ab_tiny:
  r9:8 = combine(##0x00100000,#0)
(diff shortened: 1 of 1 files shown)


tgross35 (approved):
I've run into this with build-std on the target before and meant to ask but never actually did. Thanks!

androm3da:
> I've run into this with build-std on the target before and meant to ask but never actually did. Thanks!

I feel like I probably encountered it and was so convinced of my prowess w/ the original fix I just told myself "hmm - some echoes of that old bug.  it'll come out when the crate gets updated ... "  `:facepalm:`

tgross35:
(pushed with a change to canonicalize the issue number in the commit message, since triagebot will yell about that soon https://github.com/rust-lang/compiler-builtins/pull/1065)

androm3da:
@tgross35 how do I propagate this fix to mitigate the CI issue we see in stdarch?  c-b needs to be updated in rust-src?  Is that something I can do or something I request from some project role?

tgross35:
I have a sync in progress (unfortunately they can be a bit tricky), will try to wrap it up today

androm3da:
> I have a sync in progress (unfortunately they can be a bit tricky), will try to wrap it up today

Oh thanks so much.  No hurry on my account, just wanted to know if I could shepherd it through.  Good to hear you're on it.

tgross35:
That took longer than expected but this was synced back in https://github.com/rust-lang/rust/pull/152420

PR #1070, opened 2026-02-02
libm: fix acosh & acoshf for negative inputs

quaternic:
The acosh-functions were incorrectly returning finite values for some negative inputs (should be NaN for any `x < 1.0`)

The bug was inherited when originally ported from musl, and this patch follows their fix for single-precision acoshf in https://git.musl-libc.org/cgit/musl/commit/?id=c4c38e6364323b6d83ba3428464e19987b981d7a

### DIFF
--- a/libm-test/src/precision.rs
+++ b/libm-test/src/precision.rs
@@ -266,6 +266,15 @@ impl MaybeOverride<(f64,)> for SpecialCase {
             return XFAIL_NOCHECK;
         }
 
+        if ctx.base_name == BaseName::Acosh
+            && input.0 < 1.0
+            && actual.is_nan()
+            && ctx.basis == CheckBasis::Musl
+        {
+            // Musl sometimes evaluates acosh(negative) to a numeric value
+            return XFAIL_NOCHECK;
+        }
+
         // maybe_check_nan_bits(actual, expected, ctx)
         unop_common(input, actual, expected, ctx)
     }
@@ -295,19 +304,6 @@ fn unop_common<F1: Float, F2: Float>(
     expected: F2,
     ctx: &CheckCtx,
 ) -> CheckAction {
-    if ctx.base_name == BaseName::Acosh
-        && input.0 < F1::NEG_ONE
-        && !(expected.is_nan() && actual.is_nan())
-    {
-        // acoshf is undefined for x <= 1.0, but we return a random result at lower values.
-
-        if ctx.basis == CheckBasis::Musl {
-            return XFAIL_NOCHECK;
-        }
-
-        return XFAIL("acoshf undefined");
-    }
-
     if (ctx.base_name == BaseName::Lgamma || ctx.base_name == BaseName::LgammaR)
         && input.0 < F1::ZERO
         && !input.0.is_infinite()
--- a/libm/src/math/acosh.rs
+++ b/libm/src/math/acosh.rs
@@ -1,4 +1,4 @@
-use super::{log, log1p, sqrt};
+use super::{Float, log, log1p, sqrt};
 
 const LN2: f64 = 0.693147180559945309417232121458176568; /* 0x3fe62e42,  0xfefa39ef*/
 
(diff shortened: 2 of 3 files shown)


libm-test/src/precision.rs
```diff
@@ -401,14 +388,6 @@ fn binop_common<F1: Float, F2: Float>(
...
     }
 
-    // FIXME(#939): this should not be skipped, there is a bug in our implementationi.
-    if ctx.base_name == BaseName::FmaximumNum
-        && ctx.basis == CheckBasis::Mpfr
-        && ((input.0.is_nan() && actual.is_nan() && expected.is_nan()) || input.1.is_nan())
-    {
-        return XFAIL_NOCHECK;
-    }
-
```
quaternic, line 411: This is unrelated to acoshf, but was just a leftover from https://github.com/rust-lang/compiler-builtins/pull/939/ which is already merged.

libm/src/math/acosh.rs
```diff
@@ -9,19 +9,20 @@ const LN2: f64 = 0.693147180559945309417232121458176568; /* 0x3fe62e42,  0xfefa3
...
+    let e = u & 0x7ff;
 
     /* x < 1 domain error is handled in the called functions */
 
     if e < 0x3ff + 1 {
         /* |x| < 2, up to 2ulp error in [1,1.125] */
-        return log1p(x - 1.0 + sqrt((x - 1.0) * (x - 1.0) + 2.0 * (x - 1.0)));
+        let x_1 = x - 1.0;
+        log1p(x_1 + sqrt(x_1 * x_1 + 2.0 * x_1))
+    } else if u < 0x3ff + 26 {
```
quaternic: Unfortunately the nonessential changes ended up hiding the actual bug fix in the diff, which is simply replacing
```rust
if e < 0x3ff + 26 {
```
with
```rust
if u < 0x3ff + 26 {
```
The difference is that `u` hasn't had the sign bit cleared, so negative inputs no longer take that branch. The fix for `acoshf` is equivalent.

tgross35:
To clarify; musl fixed their acoshf but not acosh? And this PR fixes both for us?

libm/src/math/acosh.rs
```diff
@@ -9,19 +9,20 @@ const LN2: f64 = 0.693147180559945309417232121458176568; /* 0x3fe62e42,  0xfefa3
 /// `x` must be a number greater than or equal to 1.
 #[cfg_attr(assert_no_panic, no_panic::no_panic)]
 pub fn acosh(x: f64) -> f64 {
-    let u = x.to_bits();
-    let e = ((u >> 52) as usize) & 0x7ff;
+    let u = x.to_bits() >> 52;
+    let e = u & 0x7ff;
```
tgross35: I'm assuming the `as usize` cast gave slightly better codegen on 32-bit systems, was it problematic?

quaternic:
> To clarify; musl fixed their acoshf but not acosh? And this PR fixes both for us?

Yes, exactly.

libm/src/math/acosh.rs
quaternic: Oh, that would explain the cast. I didn't consider that. Even casting to `u16` should work just as well, and I would hope the optimizations don't need that. (Tested on godbolt: Some insignificant differences on 32-bit.)

tgross35: > Tested on godbolt: Some insignificant differences on 32-bit.

Do you mean 32-bit codegen didn't regress? If so then I suppose it's fine to keep. Otherwise, `.ex()` can be used https://github.com/rust-lang/compiler-builtins/blob/0e5f78cd1963b89bf0b715c810919917e3fdce6b/libm/src/math/support/float_traits.rs#L138-L141

libm/src/math/acosh.rs
```diff
@@ -9,19 +9,20 @@ const LN2: f64 = 0.693147180559945309417232121458176568; /* 0x3fe62e42,  0xfefa3
...
     /* x < 1 domain error is handled in the called functions */
 
     if e < 0x3ff + 1 {
         /* |x| < 2, up to 2ulp error in [1,1.125] */
-        return log1p(x - 1.0 + sqrt((x - 1.0) * (x - 1.0) + 2.0 * (x - 1.0)));
+        let x_1 = x - 1.0;
+        log1p(x_1 + sqrt(x_1 * x_1 + 2.0 * x_1))
+    } else if u < 0x3ff + 26 {
+        /* 2 <= x < 0x1p26 */
+        log(2.0 * x - 1.0 / (x + sqrt(x * x - 1.0)))
```
tgross35: `u` is usually the value of `x.to_bits()` so I'd slightly prefer to just use `e` here if that works, or give it a different name if not.

libm/src/math/acosh.rs
quaternic: Rewriting in terms of `.ex()` regressed icounts by ~7% because the semantically clear `x.ex() < FOO && x.is_sign_positive()` didn't optimize that well. The current version just does unsigned comparisons on `ux = x.to_bits()`, and seems to be good enough.

quaternic:
> I squashed the last two commits since they're related and added some more context to the commit messages.

Thanks!

PR #1073, opened 2026-02-03
Check for symbols with default visibility in symbol-check

zmodem:
to ensure that compiler-builtins doesn't expose any symbols with default visibility. See discussion on https://github.com/rust-lang/rust/pull/151998

### DIFF
--- a/crates/symbol-check/src/main.rs
+++ b/crates/symbol-check/src/main.rs
@@ -58,6 +58,7 @@ fn main() {
         "The binaries will not be checked for executable stacks. Used for embedded targets which \
         don't set `.note.GNU-stack` since there is no protection.",
     );
+    opts.optflag("", "no-visibility", "Don't check visibility.");
 
     let print_usage_and_exit = |code: i32| -> ! {
         eprintln!("{}", opts.usage(USAGE));
@@ -74,6 +75,7 @@ fn main() {
     }
 
     let no_os_target = m.opt_present("no-os");
+    let check_visibility = !m.opt_present("no-visibility");
     let free_args = m.free.iter().map(String::as_str).collect::<Vec<_>>();
     for arg in &free_args {
         assert!(
@@ -85,18 +87,18 @@ fn main() {
     if m.opt_present("build-and-check") {
         let target = m.opt_str("target").unwrap_or(env!("HOST").to_string());
         let paths = exec_cargo_with_args(&target, &free_args);
-        check_paths(&paths, no_os_target);
+        check_paths(&paths, no_os_target, check_visibility);
     } else if m.opt_present("check") {
         if free_args.is_empty() {
             print_usage_and_exit(1);
         }
-        check_paths(&free_args, no_os_target);
+        check_paths(&free_args, no_os_target, check_visibility);
     } else {
         print_usage_and_exit(1);
     }
 }
 
-fn check_paths<P: AsRef<Path>>(paths: &[P], no_os_target: bool) {
+fn check_paths<P: AsRef<Path>>(paths: &[P], no_os_target: bool, check_visibility: bool) {
     for path in paths {
         let path = path.as_ref();
         println!("Checking {}", path.display());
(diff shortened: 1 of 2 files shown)


zmodem:
Lots of red on the CI, but that means the patch is working :-)

It's flagging the outline atomics targeted by the Rust-side PR. Also `__aeabi_idivmod` etc. which I'm assuming are similarly defined as naked_asm functions. Also `__rust_probestack`.

`RSQRT_TAB` is a variable though, so I guess that will need another patch to hide.

tgross35:
Looking through the failures, all the naked functions fail as expected. But it seems like [`static RSQRT_TAB`](https://github.com/rust-lang/compiler-builtins/blob/0e5f78cd1963b89bf0b715c810919917e3fdce6b/libm/src/math/generic/sqrt.rs#L346-L367) is also exposed? Any idea what's happening there?

tgross35:
Oh sorry, you mentioned that in a comment already and I missed it :) Is that something that could also be fixed in https://github.com/rust-lang/rust/pull/151998?

zmodem:
Yes, I've updated that PR.

crates/symbol-check/src/main.rs
```diff
@@ -299,6 +300,36 @@ fn verify_core_symbols(archive: &BinFile) {
...
+
+        let sym = SymInfo::new(&symbol, obj, member);
+        if sym.scope == SymbolScope::Dynamic {
+            visible.push(sym);
+        }
+
+        found_any = true
+    });
+
+    assert!(found_any, "no symbols found");
```
tgross35: To match up with 2db897ffe717de7fc10738fa2aced1439bdc5e90 this will need

```suggestion
    if archive.has_symbol_tables() {
        assert!(found_any, "no symbols found");
    }
```

tgross35:
I recently added some tests for symcheck, would you be able to add one to `crates/symbol-check/tests/all.rs` as well? It should look a bit like https://github.com/rust-lang/compiler-builtins/blob/a7b1d8e697ab64cac849c07b80f356ed3655a468/crates/symbol-check/tests/all.rs#L65-L76.

zmodem:
> I recently added some tests for symcheck,

Hmm, those tests don't play well with my visibility check :-)

The main issue is that the test is building things which are not compiler-builtins, such as executables and rust libraries, and those will have plenty of visible symbols.

I can avoid some of it by ignoring executables in my check, but for example the `good` function in good_lib.rs will be visible, and I can't fix that with `-Zdefault-visibility=hidden` because that's not available with stable rustc.

(I tried adding `#![compiler_builtins]` to good_lib.rs, but rustc didn't like that either.)

crates/symbol-check/src/main.rs
zmodem: Done.

zmodem:
I added a `--no-visibility` flag which the tests that don't care about visibility checking can pass. Does that sound like a reasonable solution?

tgross35:
Ah sorry yeah, the whole symcheck thing and its tests are unfortunately a bit rough around the edges. It should be possible to add a C/C++ file using `__attribute__ ((visibility ("hidden")))` and build it (like `test_exe_gnu_stack_section` does) if that works. 

Alternatively I think it should always be using nightly rustc in CI? Not worth chasing things down too much though if it's not easy to get working.

zmodem:
> Ah sorry yeah, the whole symcheck thing and its tests are unfortunately a bit rough around the edges. It should be possible to add a C/C++ file using `__attribute__ ((visibility ("hidden")))` and build it (like `test_exe_gnu_stack_section` does) if that works.

The linker will still insert a couple of visible symbols in the executable (`__bss_start`, `_edata`, `_end` I think).

I think having a flag is a reasonable fix? The default use case is checking for visible symbols in compiler-builtins libraries, and for other files -- as produced by some of these tests -- we can pass `--no-visibility`.

With https://github.com/rust-lang/rust/pull/151998 in nightly, the tests are all green :-)

PR #1075, opened 2026-02-04
fix tests for lgamma

quaternic:
The tests were using `rug::ln_gamma` as a reference for `libm::lgamma`, which actually computes the natural logarithm *of the absolute value* of the Gamma function.

This changes the range of inputs used for the icount-benchmarks of these functions, which causes false regressions in https://github.com/rust-lang/compiler-builtins/actions/runs/21788698368/job/62864230903?pr=1075#step:7:2710
ci: allow-regressions

### DIFF
--- a/libm-test/src/domain.rs
+++ b/libm-test/src/domain.rs
@@ -207,7 +207,7 @@ impl<F: Float, I: Int> EitherPrim<Domain<F>, Domain<I>> {
     .into_prim_float()];
 
     /// Domain for `loggamma`
-    const LGAMMA: [Self; 1] = Self::STRICTLY_POSITIVE;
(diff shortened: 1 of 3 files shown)


tgross35:
I think you'll need to add this function to the skip list here https://github.com/rust-lang/compiler-builtins/blob/3571e6ee328802f49b175661615129d84b17bbad/libm-test/benches/random.rs#L128-L140

Tbh I'm not sure there is much utility in the benchmarks against musl anymore, I never look at the results.

quaternic:
Changing the domain to unbounded revealed the relative inaccuracy that these functions have in the negative inputs, when the result is close to zero. I added some targeted `XFAIL_NOCHECK` to weed out those cases but still properly test most inputs.

tgross35 (approved):

PR #1085, opened 2026-02-07
libm: Fix `_status` status outputs on null-mantissa inputs

tgross35:
Values like 0.5 that have an exponent but not a mantissa are currently being reported as OK rather than INEXACT for ceil, floor, and trunc. Fix this and clean up the test cases.

ci: allow-regressions

### DIFF
--- a/libm/src/math/generic/trunc.rs
+++ b/libm/src/math/generic/trunc.rs
@@ -10,7 +10,7 @@ pub fn trunc<F: Float>(x: F) -> F {
 
 #[inline]
 pub fn trunc_status<F: Float>(x: F) -> FpResult<F> {
-    let mut xi: F::Int = x.to_bits();
+    let xi: F::Int = x.to_bits();
     let e: i32 = x.exp_unbiased();
 
     // The represented value has no fractional part, so no truncation is needed
@@ -18,24 +18,26 @@ pub fn trunc_status<F: Float>(x: F) -> FpResult<F> {
         return FpResult::ok(x);
     }
 
-    let mask = if e < 0 {
-        // If the exponent is negative, the result will be zero so we mask out everything
+    let clear_mask = if e < 0 {
+        // If the exponent is negative, the result will be zero so we clear everything
         // except the sign.
-        F::SIGN_MASK
+        !F::SIGN_MASK
     } else {
-        // Otherwise, we mask out the last `e` bits of the significand.
-        !(F::SIG_MASK >> e.unsigned())
+        // Otherwise, we keep `e` fractional bits and clear the rest.
+        F::SIG_MASK >> e.unsigned()
     };
 
-    // If the to-be-masked-out portion is already zero, we have an exact result
-    if (xi & !mask) == IntTy::<F>::ZERO {
-        return FpResult::ok(x);
-    }
+    let cleared = xi & clear_mask;
+    let status = if cleared == IntTy::<F>::ZERO {
+        // If the to-be-zeroed portion is already zero, we have an exact result.
+        Status::OK
+    } else {
+        // Otherwise the result is inexact and we will truncate, so indicate `FE_INEXACT`.
+        Status::INEXACT
+    };
 
-    // Otherwise the result is inexact and we will truncate. Raise `FE_INEXACT`, mask the
-    // result, and return.
-    xi &= mask;
-    FpResult::new(F::from_bits(xi), Status::INEXACT)
+    // Now zero the bits we need to truncate and return.
+    FpResult::new(F::from_bits(xi ^ cleared), status)
 }
 
 #[cfg(test)]


libm/src/math/support/env.rs
```diff
@@ -74,22 +74,22 @@ impl Status {
     /// result is -inf.
     /// `x / y` when `x != 0.0` and `y == 0.0`,
     #[cfg_attr(not(feature = "unstable-public-internals"), allow(dead_code))]
-    pub const DIVIDE_BY_ZERO: Self = Self(1 << 2);
+    pub const DIVIDE_BY_ZERO: Self = Self(1 << 1);
```
quaternic, line 81: I'm guessing the reason why this skipped a bit is that they used the same bit positions as the x86 status registers: [x86-64 MXCSR Register](https://www.lns.cornell.edu/~cesrulib/documentation/totalview/html/Reference_Guides/wwhelp/wwhimpl/common/html/wwhelp.htm#href=Intelx86MXSCRRegister.html&single=true) as well as the separate x87 status word. Those use `0x2` is for the non-standard Denormal Operand Exception.

The patched contiguous labeling matches Arm, but RISC-V has yet another layout (same low 5 bits but in reverse order).

libm/src/math/generic/trunc.rs
```diff
@@ -13,136 +13,113 @@ pub fn trunc_status<F: Float>(x: F) -> FpResult<F> {
...
     // result, and return.
-
-    let status = if xi & F::SIG_MASK == F::Int::ZERO {
-        Status::OK
-    } else {
-        Status::INEXACT
-    };
     xi &= mask;
-    FpResult::new(F::from_bits(xi), status)
+    FpResult::new(F::from_bits(xi), Status::INEXACT)
```
quaternic: If `xi & !mask == 0`, then `xi & mask == xi`, so this can just return `F::from_bits(xi & mask)` and avoid the branch.

Also, it might be slightly better to use the negated mask ("bits to clear") from the start.
```rust
let cleared = xi & clear_mask;
let status = if cleared == 0 { Status::OK } else { Status::INEXACT };
FpResult::new(F::from_bits(xi ^ cleared), status)
```

libm/src/math/generic/floor.rs
```diff
@@ -35,123 +34,113 @@ pub fn floor_status<F: Float>(x: F) -> FpResult<F> {
...
         F::from_bits(ix)
     } else {
-        // |x| < 1.0, raise an inexact exception since truncation will happen.
-        if ix & F::SIG_MASK == F::Int::ZERO {
-            status = Status::OK;
-        } else {
-            status = Status::INEXACT;
+        // |x| < 1.0, zero or inexact with truncation
+
+        if (ix & !F::SIGN_MASK) == F::Int::ZERO {
```
quaternic, line 46: The performance regression didn't make sense to me, so I looked at the assembly and then LLVM-IR, and it seems that LLVM is being "smart" by turning this into a floating point comparison `x == 0.0_f16`, and that generates the equivalent of `x as f32 == 0.0_f32` with a call to `__extendhfsf2`.

tgross35: Thanks for digging into all of this. I wrote up an issue here https://github.com/llvm/llvm-project/issues/180630, wonder what else might be hurting from it.

libm/src/math/support/env.rs
tgross35: Great catch. Is there any reason to be consistent with these? I can't think of much purpose to that unless we use host floats with the environment via asm and load the register back to a `Status`.

quaternic: Yeah, and doing that would require one of
 - mapping to and from the architecture specific layout at that interface
 - making these constants architecture specific

But we don't use it like that, so there's no need. Now that the `Debug` impl prints only the names, and since the UNKNOWN bits should not be settable, the actual bit-values are private to this module.

libm/src/math/generic/floor.rs
quaternic: On x86-64, these contain one or more calls to `__extendhfsf2` (f16 as f32):
```text
    fmaximumf16
    fminimumf16
    fmaximum_numf16
    fminimum_numf16
    fmaxf16
    fminf16
    fdimf16
    sqrtf16
    rintf16
    roundf16
    fmodf16
    ldexpf16
```
The min/max -functions are doing floating point comparisons, but might benefit from inspecting the bits instead. The others do float arithmetic and also contains calls to  `__truncsfhf2` (f32 as f16). So none of them are explicit bit-operations being turned into float ops.

Most other libcalls are in various `f128`-functions.

libm/src/math/support/env.rs
tgross35: Added a note for this here https://github.com/rust-lang/compiler-builtins/compare/32417939f6558916252215c25a7c48b33c41b7df..5c981cfffd98bfb5135395712409811b6d280903

libm/src/math/generic/trunc.rs
tgross35: Thanks, this came with some nice small speedups

```text
[660](https://github.com/rust-lang/compiler-builtins/actions/runs/21939477624/job/63361314470?pr=1085#step:7:2661)
icount::icount_bench_trunc_group::icount_bench_trunc logspace:setup_trunc()
  Baselines:                      softfloat|softfloat (old)
  Instructions:                        7057|7579                 (-6.88745%) [-1.07397x]
  L1 Hits:                             7674|8197                 (-6.38038%) [-1.06815x]
  LL Hits:                                2|1                    (+100.000%) [+2.00000x]
  RAM Hits:                               3|3                    (No change)
  Total read+write:                    7679|8201                 (-6.36508%) [-1.06798x]
  Estimated Cycles:                    7789|8307                 (-6.23570%) [-1.06650x]
icount::icount_bench_truncf128_group::icount_bench_truncf128 logspace:setup_truncf128()
  Baselines:                      softfloat|softfloat (old)
  Instructions:                       12263|13271                (-7.59551%) [-1.08220x]
  L1 Hits:                            14883|15889                (-6.33142%) [-1.06759x]
  LL Hits:                                3|5                    (-40.0000%) [-1.66667x]
  RAM Hits:                               6|6                    (No change)
  Total read+write:                   14892|15900                (-6.33962%) [-1.06769x]
  Estimated Cycles:                   15108|16124                (-6.30117%) [-1.06725x]
icount::icount_bench_truncf16_group::icount_bench_truncf16 logspace:setup_truncf16()
  Baselines:                      softfloat|softfloat (old)
  Instructions:                        7952|8726                 (-8.87004%) [-1.09733x]
  L1 Hits:                             8520|9294                 (-8.32795%) [-1.09085x]
  LL Hits:                                1|1                    (No change)
  RAM Hits:                               5|5                    (No change)
  Total read+write:                    8526|9300                 (-8.32258%) [-1.09078x]
  Estimated Cycles:                    8700|9474                 (-8.16973%) [-1.08897x]
icount::icount_bench_truncf_group::icount_bench_truncf logspace:setup_truncf()
  Baselines:                      softfloat|softfloat (old)
  Instructions:                        7281|7865                 (-7.42530%) [-1.08021x]
  L1 Hits:                             7896|8479                 (-6.87581%) [-1.07383x]
  LL Hits:                                2|3                    (-33.3333%) [-1.50000x]
  RAM Hits:                               5|5                    (No change)
  Total read+write:                    7903|8487                 (-6.88111%) [-1.07390x]
  Estimated Cycles:                    8081|8669                 (-6.78279%) [-1.07276x]
```

libm/src/math/generic/floor.rs
tgross35: Given the benchmarks primarily affect f16 and f128 and that we can't really easily work around LLVM, I'm just going to accept the regressions here.

libm/src/math/generic/trunc.rs
```diff
@@ -10,139 +10,118 @@ pub fn trunc<F: Float>(x: F) -> F {
...
-        // C2: If the exponent is negative, the result will be zero so we mask out everything
+    let clear_mask = if e < 0 {
+        // If the exponent is negative, the result will be zero so we clear everything
         // except the sign.
-        F::SIGN_MASK
+        !F::SIGN_MASK
     } else {
-        // C3: Otherwise, we mask out the last `e` bits of the significand.
-        !(F::SIG_MASK >> e.unsigned())
+        // Otherwise, we clear the last `e` bits of the significand.
```
quaternic:
```suggestion
        // Otherwise, we keep `e` fractional bits and clear the rest.
```
(The old comment was incorrect too)

Issue #835, opened 2024-10-07
powerpc64le SIGILL


tgross35:
Just making a note of this for my future reference: it seems like the musl implementations have crashes on PPC LE. I don't know if this is because of a bug in musl or qemu, but based on the below it seems like just a qemu thing.

Qemu execution:

```sh
$ LD_LIBRARY_PATH='/target/powerpc64le-unknown-linux-gnu/de
bug/build/musl-math-sys-1082aab9c9d13406/out:/target/powerpc64le-unknown-linux-gnu/debug/deps:/target/powerpc64le-unknown-linux-gnu/debug:/root/.rustup/toolchains/nightly-aarch64-unknown-linux-gnu/lib/rustlib/powerpc64le-unknown-linux-gnu/lib:/root/.rustup/toolchains/nightly-aarch64-unknown-linux-gnu/lib' OUT_DIR=/target/powerpc64le-unknown-linux-gnu/debug/build/libm-test-625573e9becfdb68/out qemu-ppc64le-static -g 1234 /target/powerpc64le-unknown-linux-gnu/debug/deps/compare_to_musl-cf318f8dd1aeaaa8 --exact tan
```

GDB session

```text
$ gdb-multiarch
(gdb) target remote :1234
Remote debugging using :1234
Reading /target/powerpc64le-unknown-linux-gnu/debug/deps/compare_to_musl-cf318f8dd1aeaaa8 from remote target...
warning: File transfers from remote targets can be slow. Use "set sysroot" to access files locally instead.
Reading /target/powerpc64le-unknown-linux-gnu/debug/deps/compare_to_musl-cf318f8dd1aeaaa8 from remote target...
...
(No debugging symbols found in target:/lib64/ld64.so.2)
Failed to read a valid object file image from memory.
0x0000100000837a80 in ?? () from target:/lib64/ld64.so.2
(gdb) set arch powerpc:common64
The target architecture is set to "powerpc:common64".
(gdb) set endian little
The target is set to little endian.
(gdb) c
Continuing.
Reading /lib/libc.so.6 from remote target...
Reading /lib/libm.so.6 from remote target...
Reading /lib/libgcc_s.so.1 from remote target...
Error while mapping shared library sections:
`target:/lib/libgcc_s.so.1': not in executable format: file format not recognized
Reading /usr/lib/debug/.build-id/9b/d190627d3463a2d2cf11d9250fb5619e7711ca.debug from remote target...
...
[New Thread 8887.8899]

[...]

PR #1096, opened 2026-02-14
libm: Reenable should_panic tests on ppc64le

Gelbpunkt:
The tests pass successfully for me on powerpc64le-unknown-linux-gnu with nightly 1.95.

Closes rust-lang/compiler-builtins#835

### DIFF
--- a/libm/src/math/support/big/tests.rs
+++ b/libm/src/math/support/big/tests.rs
@@ -261,8 +261,6 @@ fn shr_u256() {
 #[test]
 #[should_panic]
 #[cfg(debug_assertions)]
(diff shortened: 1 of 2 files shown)


tgross35:
I merged 3fa5a8cba55d4e8a0fce06897fd477d66b41da48 a while ago, probably should have pinged you on it (sorry about that). I _think_ I tried disabling these tests there but they still failed, so I wonder if I'm misremembering or if we got a fix since then.

I'd be curious to hear your thoughts on that patch in retrospect anyway about the baseline and SIGILLs, in case I misrepresented something there.

Gelbpunkt:
> Any idea what the issue was?

I think it was a QEMU bug, but I may be wrong...

PR #1098, opened 2026-02-16
fix testing at lgammaf overflow threshold

quaternic:
The current implementation of lgammaf evaluates to f32::MAX at the first input that would overflow if correctly rounded. This is still well within allowed error, so special case it.

### DIFF
--- a/libm-test/src/precision.rs
+++ b/libm-test/src/precision.rs
@@ -222,6 +222,15 @@ impl MaybeOverride<(f32,)> for SpecialCase {
             return XFAIL_NOCHECK;
         }
 
+        // the testing infrastructure doesn't account for allowed ulp in the case of overflow
+        if matches!(ctx.base_name, BaseName::Lgamma | BaseName::LgammaR)
+            && input.0 == 4.0850034e36
+            && expected.is_infinite()
+            && actual == F::MAX
+        {
+            return XFAIL_NOCHECK;
+        }
+
         // FIXME(correctness): lgammaf has high relative inaccuracy near its zeroes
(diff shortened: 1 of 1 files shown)


quaternic:
@tgross35
The exhaustive tests for lgammaf pass locally with this, so it should fix the CI-issue in #1093. I've also checked that the f64-variant doesn't have the same problem.

tgross35:
Would be better to adjust the tests to allow ULP to overlap the max finite / infinite boundary? That's about what I was planning to do but didn't get around to it.

Adding a way to run custom extensive tests and will fix the indicatif CI issue in https://github.com/rust-lang/compiler-builtins/pull/1099.

quaternic:
> Would be better to adjust the tests to allow ULP to overlap the max finite / infinite boundary? That's about what I was planning to do but didn't get around to it.

Probably, yeah. But it might not be a simple fix to do properly, since it should account for what the exact result is. That is, knowing that the rounded result should be `inf` is not enough to decide whether to accept a large but finite value.

tgross35:
> Probably, yeah. But it might not be a simple fix to do properly, since it should account for what the exact result is. That is, knowing that the rounded result should be `inf` is not enough to decide whether to accept a large but finite value.

Ah right, we can't differentiate between barely saturating to infinity and having a truly infinite result. Makes sense, it hasn't been a problem anyway.

tgross35:
Extensive passed already at https://github.com/rust-lang/compiler-builtins/actions/runs/22056086180/job/63724403806 so I dropped it

PR #1100, opened 2026-02-16
meta: Add a root Cargo.lock file

tgross35:
In order to improve upon some inconsistencies seen across CI runs and locally, add a Cargo.lock file for the main workspace.

ci: test-libm

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -590,6 +590,7 @@ dependencies = [
  "console",
  "portable-atomic",
  "unit-prefix",
+ "web-time",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


bjorn3:
If you do this, you will also have to exclude compiler-builtins from the library workspace in the rust repo. A crate may only be in one workspace at a time.

tgross35:
There isn't any overlap currently; the compiler builtins Cargo.toml is duplicated, one for use in r-l/r and one for use here. That's because of the `core` dep, not this though https://github.com/rust-lang/compiler-builtins/blob/bba024d20464f2e50144384199960e7b6d5da920/builtins-shim/Cargo.toml#L1-L12.

PR #1104, opened 2026-03-06
min,max: Add tests for signaling NaNs and update documentation

tgross35:
We do handle signaling NaNs properly, with the exception of raising
exceptions as IEEE 754 requires. Add tests to this effect for
`fmin`, `fminimum`, `fminimum_num`, and the max variants.

ci: allow-many-extensive
ci: allow-regressions (hex float printing)

### DIFF
--- a/libm/src/math/generic/fmaximum.rs
+++ b/libm/src/math/generic/fmaximum.rs
@@ -7,7 +7,8 @@
 //! - +0.0 if x and y are zero with opposite signs
 //! - qNaN if either operation is NaN
 //!
-//! Excluded from our implementation is sNaN handling.
+//! Note that the IEEE 754-2019 specifies that a sNaN in either argument should signal invalid,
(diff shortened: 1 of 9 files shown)


tgross35:
Cc @RalfJung

libm/src/math/generic/fmaximum.rs
```diff
@@ -7,7 +7,8 @@
 //! - +0.0 if x and y are zero with opposite signs
 //! - qNaN if either operation is NaN
 //!
-//! Excluded from our implementation is sNaN handling.
+//! Note that the IEEE 754-2019 specifies that a sNaN in eitt her argument should signal invalid,
```
RalfJung:
```suggestion
//! Note that the IEEE 754-2019 specifies that a sNaN in either argument should signal invalid,
```
happens a few times

PR #1109, opened 2026-03-14
export `__mulsi3` on `m68k`

folkertdev:
Fix for https://github.com/rust-lang/rust/issues/153772. 

This is completely untested. That is consistent with `avr` and honestly I'm not sure how I'd test this really.


compiler-builtins/src/int/mul.rs
tgross35, line 158: Would you mind just moving the riscv __muldi3 too? Probably no point in having the file just for the one algorithm. Context for why it's special: https://github.com/rust-lang/compiler-builtins/pull/461

(Can also put __mulsi3 at the top of this block so it's sorted small->large)

folkertdev: Took a bit of fighting with the `intrinsics!` macro but I think I got it? `cfg_select!` is weeks away which would help a bit.

PR #1110, opened 2026-03-15
ci: Add zizmor for Github Actions static analysis

yuzibo:


yuzibo:
This is one proposal to address concerns from T-infra's [view]([#t-infra > RISC-V runner in compiler builtins @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/RISC-V.20runner.20in.20compiler.20builtins/near/579302948)) regarding external CI resources. 

The result is flaky now, but we can easily add it to `success -> needs` [section](https://github.com/rust-lang/compiler-builtins/blob/main/.github/workflows/main.yaml#L375C3-L376C11) to enable it.

.github/workflows/main.yaml
```diff
@@ -211,6 +211,23 @@ jobs:
       run: ./ci/update-musl.sh
     - run: cargo clippy --workspace --all-targets
 
+  zizmor:
+    name: Zizmor (Static analysis for GitHub Actions)
+    runs-on: ubuntu-24.04
+    permissions:
+      security-events: write
+      contents: read
+      actions: read
```
tgross35: Per https://github.com/zizmorcore/zizmor-action?tab=readme-ov-file#usage-with-github-advanced-security-recommended these are only needed for private/internal repos

.github/workflows/main.yaml
```diff
@@ -211,6 +211,23 @@ jobs:
...
+    timeout-minutes: 10
+    steps:
+      - name: Checkout zizmor repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
```
tgross35:
```suggestion
```

Should no longer be needed with v6 to pass the audit

.github/workflows/main.yaml
```diff
@@ -211,6 +211,23 @@ jobs:
...
+    name: Zizmor (Static analysis for GitHub Actions)
+    runs-on: ubuntu-24.04
+    permissions:
+      security-events: write
+      contents: read
+      actions: read
+    timeout-minutes: 10
+    steps:
+      - name: Checkout zizmor repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
tgross35:
```suggestion
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```

It's checking out this repo and not zizimor, right?

tgross35:
> The result is flaky now, but we can easily add it to `success -> needs` [section](https://github.com/rust-lang/compiler-builtins/blob/main/.github/workflows/main.yaml#L375C3-L376C11) to enable it.

The job itself succeeds even though it flags some things right, and will just report them in the security tab? If so, I think it's fine to add to `success` now.

.github/workflows/main.yaml
yuzibo: hmm, right, my bad

.github/workflows/main.yaml
```diff
@@ -211,6 +211,19 @@
...
 
+  zizmor:
+    name: Zizmor (Static analysis for GitHub Actions)
+    runs-on: ubuntu-24.04
+    permissions:
+      security-events: write
+    timeout-minutes: 10
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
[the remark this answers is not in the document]
yuzibo: Interesting. this will left one notice here.

marcoieni:
An alternative to the security tab is setting `advanced-security: false` to just fail the job is there's a security issue. 

https://github.com/rust-lang/crates-io-auth-action/blob/12221606e175a865016c7e1dfc0f045a6ea18eec/.github/workflows/ci.yml#L96

Feel free to decide the approach you prefer, Trevor 👍

tgross35:
> An alternative to the security tab is setting `advanced-security: false` to just fail the job is there's a security issue.

I think this is fine since it's a bit less invasive to anyone's PR CI if zizmor happens to add new lints, as long as there isn't any issue with that. Thanks for pointing out the option.

I have https://github.com/rust-lang/compiler-builtins/pull/1113 to pin things, I'll land this PR after that and https://github.com/rust-lang/compiler-builtins/pull/1112 merge so we don't get a ton of alerts right away.

marcoieni:
> I think this is fine since it's a bit less invasive to anyone's PR CI if zizmor happens to add new lints

If you pin the zizmor github actions version, this won't happen.

tgross35:
> If you pin the zizmor github actions version, this won't happen.

Oh, I didn't realize that was an option. It would be nice to get the new versions automatically, is that something renovate can update?

tgross35:
I'll merge this for now in any case, happy to get things off the ground here

PR #1112, opened 2026-03-16
publish via trusted publishing

marcoieni:
The environment was created in https://github.com/rust-lang/team/pull/2322:
image

- [x] blocked by https://github.com/rust-lang/team/pull/2326

PR #1113, opened 2026-03-16
ci: Resolve audit issues reported by zizmore

tgross35:
Resolve the following lints:

* `unpinned-uses`: Pin action versions using `zizmor --fix`
* `excessive-permissions`: Add `permissions: {}` to all workflows
* `template-injection`: This one wasn't too concerning since `${{ ... }}` was only used for `matrix` and `needs` access, but it is easy enough to resolve by storing in an environment variable.

There is a remaining `secrets-outside-env` lint that will be resolved by https://github.com/rust-lang/compiler-builtins/pull/1112.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,4 +1,5 @@
 name: CI
+permissions: {}
 on:
   push: { branches: [main] }
   pull_request:
@@ -28,7 +29,7 @@ jobs:
       extensive_matrix: ${{ steps.script.outputs.extensive_matrix }}
       may_skip_libm_ci: ${{ steps.script.outputs.may_skip_libm_ci }}
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           fetch-depth: 500
       - name: Fetch pull request ref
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yaml
```diff
@@ -132,25 +133,25 @@ jobs:
         rustup default "$channel"
         rustup target add "${{ matrix.target }}"
 
-    - uses: taiki-e/install-action@nextest
+    - uses: taiki-e/install-action@e8a33405c3d4f114664bbedf539fcbe45b51e05d # nextest
```
taiki-e: install action's shorthand syntax (@<tool_name>) isn't compatible with hash-pin because it's not versioned.

You have to switch to @\<version> + with.tool syntax first, then switch to hash-pin.

tgross35: It actually appears to somehow work; that hash is https://github.com/taiki-e/install-action/commit/e8a3340 and per https://github.com/rust-lang/compiler-builtins/actions/runs/23162118045/job/67292427567?pr=1113 it gets things right:

```text
Run taiki-e/install-action@e8a33405c3d4f114664bbedf539fcbe45b51e05d
  with:
    tool: nextest
    checksum: true
    fallback: cargo-binstall
```

But using `with.tool` is more clear so I'll switch to that. Thanks for finding this.

taiki-e: It will be broken or at least rejected by zizmor impostor-commit rule when update is available. As documented in the readme, it's a syntax for use cases that do not pin version.

.github/workflows/main.yaml
```diff
@@ -28,7 +29,7 @@ jobs:
       extensive_matrix: ${{ steps.script.outputs.extensive_matrix }}
       may_skip_libm_ci: ${{ steps.script.outputs.may_skip_libm_ci }}
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```
marcoieni, line 32: renovate does the first transition from the tag to the commit, too 👍

PR #1114, opened 2026-03-16
ci: Add configuration for renovatebot

tgross35:
Based on the configuration for bors [1].

[1]: https://github.com/rust-lang/bors/blob/d89bfc24e8e7add3b03af3ab7307cd3b2a1af879/.github/renovate.json5

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -3,7 +3,7 @@
   extends: [
     "config:recommended",
     ":maintainLockFilesMonthly",
-    "helpers:pinGitHubActionDigests"
+    "helpers:pinGitHubActionDigestsToSemver"
   ],
   packageRules: [
     {


tgross35:
Cc @marcoieni

marcoieni (approved):
looks great. You will also get dependencies updates for rust crates. If it's too noisy and you want to only keep dependencies updates for github actions, remove the rust category 👍
As soon as you merge this, renovate should start running

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1114/b6f4b889aa6393b8c390d0c7bc21b5f8298c8c1d..2b4561181966adb82f85eb93cf66520bb9777c3e)*

.github/renovate.json5
```diff
@@ -0,0 +1,37 @@
+{
+  $schema: "https://docs.renovatebot.com/renovate-schema.json",
+  extends: [
+    "config:recommended",
+    ":maintainLockFilesMonthly",
+    "helpers:pinGitHubActionDigests"
```
marcoieni:
```suggestion
    "helpers:pinGitHubActionDigestsToSemver"
```
This will pin the exact tag of the github actions. Like this:

Image

Notice `# v6.0.2` at the end 👍
Instead with `pinGitHubActionDigests` you only see `# v6`.

The caveat is that it only works well with GitHub actions that use semver, i.e. tags like `vx.y.z`

tgross35: Awesome, thanks!

PR #1123, opened 2026-03-22
support: Rename `signed` to `cast_signed` to match `core`

tgross35:
Rename to match the methods that were stabilized in 1.87. Also applies to `unsigned` and `cast_unsigned`.


tgross35:
Clippy gives a lot of false positives due to https://github.com/rust-lang/rust-clippy/issues/12280. I don't think it's worth ignoring the lint in so many places or globally, can revisit this change later.

PR #1152, opened 2026-03-30
trunc: Use an assembly implementation on i586

tgross35:
The `trunc` implementation uses integer operations so currently works
fine on i586. However, we already have the other three easy operations
based on `frndint`, so add `trunc` and complete the set.

ci: skip-extensive

### DIFF
--- a/etc/function-definitions.json
+++ b/etc/function-definitions.json
@@ -1044,6 +1044,7 @@
     },
     "trunc": {
         "sources": [
+            "libm/src/math/arch/i586/rounding.rs",
             "libm/src/math/arch/wasm32/rounding.rs",
             "libm/src/math/generic/trunc.rs",
             "libm/src/math/trunc.rs"
--- a/libm/src/math/arch/i586/mod.rs
+++ b/libm/src/math/arch/i586/mod.rs
@@ -11,4 +11,4 @@ mod exp_all;
 mod rounding;
 
 pub use exp_all::{x87_exp, x87_exp2, x87_exp2f, x87_exp10, x87_exp10f, x87_expf};
-pub use rounding::{ceil, floor, rint};
+pub use rounding::{ceil, floor, rint, trunc};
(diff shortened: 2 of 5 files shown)


tgross35:
Based on https://github.com/rust-lang/compiler-builtins/pull/1142 to avoid conflicts.

quaternic:
I don't think we should have this just for the sake of consistency. It has strictly worse performance.

tgross35:
> I don't think we should have this just for the sake of consistency. It has strictly worse performance.

What makes this slow - is frndint that much slower than soft ops? Looking at https://rust.godbolt.org/z/s6GjGsf6E I have no idea whether the latency of 100 is correct or just a worst case estimate since I can't find it cited anywhere.

Any idea why LLVM inserts the `wait` and why `eax` gets pushed? Thought that should be caller-saved.

quaternic:
> What makes this slow - is frndint that much slower than soft ops? Looking at https://rust.godbolt.org/z/s6GjGsf6E I have no idea whether the latency of 100 is correct or just a worst case estimate since I can't find it cited anywhere.

I'll admit I only tested by comparing against the x86-64 implementation. The 32-bit code does look more complex.

Measuring `frndint` latency (on https://en.wikipedia.org/wiki/Nehalem_(microarchitecture)):
 - |x| in [0, 2^63) -> ~21 cycles
 - |x| in [2^63, Inf) -> ~40 cycles 
 - Inf -> ~230 cycles 
 - NaN -> ~250 cycles
 
 

 
Looks like Agner Fog does provide measurements for x87 instructions too, see `4. Instruction tables` in:
https://www.agner.org/optimize/#manuals

For Nehalem the listed latency for `frndint` is 22, so I'll assume the measurements don't consider potential input dependence, but AMD seems to have had a faster implementation since 2011. Intel does not.

tgross35:
@quaternic sorry for not following up, what were your final thoughts here? If it's similar or faster then I think there's not much downside in having it.

PR #1187, opened 2026-04-07
build: Fix c-b builds always being marked dirty

tgross35:


DKLoehr:
We've started seeing failures when building chromium after this commit, because we pass `MANIFEST_CARGO_DIR="."` which  doesn't end in a filename, thus panicking on `unwrap`. Would it be possible to relax the check to something like `matches!(cfg.manifest_dir.file_name(), Some("builtins-shim"))` instead, or would that cause problems?

tgross35:
Sure, happy to take a patch there!

> because we pass `MANIFEST_CARGO_DIR="."`

CARGO_MANIFEST_DIR right?

PR #1189, opened 2026-04-08
hexagon: Implement `__clear_cache` using inline assembly

androm3da:
Implement `__clear_cache` for Hexagon targets in Rust. Hexagon has separate instruction and data caches, so this flushes dirty data cache lines with `dccleaninva`, invalidates stale instruction cache lines with `icinva`, then issues an `isync` barrier.

Based on the compiler-rt implementation from
llvm/llvm-project#188411.

### DIFF
--- a/compiler-builtins/src/hexagon.rs
+++ b/compiler-builtins/src/hexagon.rs
@@ -1,5 +1,49 @@
 use core::arch::global_asm;
 
+// Hexagon L1 cache line size in bytes (Hexagon PRM sections 5.10.3-5.10.4).
+const CACHE_LINE_SIZE: usize = 32;
+
+intrinsics! {
+    pub unsafe extern "C" fn __clear_cache(start: *mut u8, end: *mut u8) {
+        // Hexagon has separate instruction and data caches.
+        let mask = !(CACHE_LINE_SIZE - 1);
+        let start_line = start.addr() & mask;
+        let end_addr = end.addr();
+
+        // Clean and invalidate data cache to push new code to memory and
+        // invalidate stale lines in the L2 cache.
+        let mut addr = start_line;
+        while addr < end_addr {
+            unsafe {
+                core::arch::asm!(
+                    "dccleaninva({addr})",
+                    addr = in(reg) addr,
+                    options(nostack, preserves_flags),
+                );
+            }
+            addr += CACHE_LINE_SIZE;
+        }
+
+        // Invalidate instruction cache so it re-fetches from memory.
+        addr = start_line;
+        while addr < end_addr {
+            unsafe {
+                core::arch::asm!(
+                    "icinva({addr})",
+                    addr = in(reg) addr,
+                    options(nostack, preserves_flags),
+                );
+            }
+            addr += CACHE_LINE_SIZE;
+        }
+
+        // Instruction sync barrier ensures subsequent fetches see the new code.
+        unsafe {
+            core::arch::asm!("isync", options(nostack, preserves_flags));
+        }
+    }
+}
+
 global_asm!(include_str!("hexagon/func_macro.s"), options(raw));
 
 global_asm!(


androm3da:
cc @quic-mliebel

compiler-builtins/src/hexagon.rs
```diff
@@ -1,5 +1,49 @@
...
 
+// Hexagon L1 cache line size in bytes (Hexagon PRM sections 5.10.3-5.10.4).
+const CACHE_LINE_SIZE: usize = 32;
+
+intrinsics! {
+    pub unsafe extern "C" fn __clear_cache(start: *mut u8, end: *mut u8) {
+        // Hexagon has separate instruction and data caches.
+        let mask = !(CACHE_LINE_SIZE - 1);
+        let start_line = start as usize & mask;
+        let end_addr = end as usize;
```
tgross35: Prefer `.addr()` over `as usize` nowadays, more obvious with the strict provenance API

compiler-builtins/src/hexagon.rs
```diff
@@ -1,5 +1,49 @@
...
+        while addr < end_addr {
+            unsafe {
+                core::arch::asm!(
+                    "icinva({addr})",
+                    addr = in(reg) addr,
+                    options(nostack, preserves_flags),
+                );
+            }
+            addr += CACHE_LINE_SIZE;
+        }
```
tgross35, line 38: Do all `dccleaninva` have to complete before `icinva`? I assume so but just checking that these can't be combined

tgross35 (commented):
One nit and one question then lgtm. I'll wait for @quic-mliebel to review, though.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1189/cb4a87f2dcd053f9627190b0ab68ac10065aa3d0..32a99016b9b81f9f4792f59634f629afef1addb5)*

compiler-builtins/src/hexagon.rs
androm3da: AFAICT yes

> Separate, non-coherent caches (Section 5.11):                                                                                              
> "Data memory accesses and program memory accesses are treated separately and held in separate caches.  Software should ensure coherency between data and program code if necessary."


> Prescribed ordering for modified code (Section 5.11):
> The software must explicitly force modified data cache lines to memory (either by using a write-through policy, or through explicit cache clean instructions). Use a barrier instruction to ensure completion of the stores. Finally, invalidate relevant instruction cache contents so the new instructions can be refetched.

...

>  Cache hierarchy asymmetry (Section 5.10.4)
>  The data cache coherency operations (including clean, invalidate, and clean and invalidate) affect both the L1 and L2 caches

So an `icinva` on line N could trigger an I-cache refetch from L2 before `dccleaninva` on that same line has flushed the new data there.

compiler-builtins/src/hexagon.rs
androm3da: Fixed.  Thanks for the suggestion.

tgross35 (approved):
This is straightforward enough so I'll just merge, thanks!

As a note if you're interested, a test like https://github.com/llvm/llvm-project/pull/188411#issuecomment-4208875204 could be added to https://github.com/rust-lang/compiler-builtins/blob/8d38cdf62bdb9b32b2d7c1da86153099ce1d3afe/builtins-test-intrinsics/src/main.rs. That gets built and run with nothing other than our c-b so anything we're missing shows up.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1189/cb4a87f2dcd053f9627190b0ab68ac10065aa3d0..c2b7f724afb00449dcf86d790dc05bd9cb1114ca)*

androm3da:
> This is straightforward enough so I'll just merge, thanks!

Nice, thanks!  Hey - we landed it here before it arrived in llvm's `compiler-rt`, that's an interesting milestone. :)

> As a note if you're interested, a test like [llvm/llvm-project#188411 (comment)](https://github.com/llvm/llvm-project/pull/188411#issuecomment-4208875204) could be added to https://github.com/rust-lang/compiler-builtins/blob/8d38cdf62bdb9b32b2d7c1da86153099ce1d3afe/builtins-test-intrinsics/src/main.rs. That gets built and run with nothing other than our c-b so anything we're missing shows up.

Only `hexagon-sim` from the hexagon SDK models the cache with this much fidelity.  QEMU hexagon wouldn't exhibit this difference from the baseline, so if `__clear_cache` became a nop it wouldn't be detected.  The Hexagon SDK's a bit less convenient to integrate into CI pipelines, so I'll just cross my fingers for now.  In practice I think compiler-builtins tend to be fairly stable.

PR #1206, opened 2026-04-20
support: Update vendored cfg-if to upstream for `true` and `false` support

tgross35:
Link: https://github.com/rust-lang/cfg-if/commit/15aec4a67e633254e726bf477b8b86c65687bfc6


bjorn3:
Maybe we could switch to `core::cfg_select!`?

tgross35:
libm uses this so we can't yet because of MSRV. But maybe there's a polyfill for `cfg_select` that we could switch to so at least the syntax is consistent (I'm actually surprised docs.rs/cfg-select doesn't exist)

PR #1212, opened 2026-05-01
Fix some typos in comments

Hawk777:


tgross35 (approved):
For future reference we don't usually take small fixes to internal comments since it's usually more effort than it's worth all around. But no harm since you've already done it, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1212/17d267af6bdc47d16d9c9b430fc22365ac7324f8..bed5ff3208c9a1c907b300995eea49b3a6b13cac)*

PR #1213, opened 2026-05-01
c-b: Avoid a build script panic from `.unwrap()`

DKLoehr:
The current code expects CARGO_MANIFEST_DIR to always end in a directory name, which prevents passing something like ".". We can continue to support that by unwrapping using `if let` first.

Followup to rust-lang/compiler-builtins#1187.

### DIFF
--- a/compiler-builtins/build.rs
+++ b/compiler-builtins/build.rs
@@ -12,6 +12,7 @@ fn main() {
     // Work around building as part of `builtins-shim`: if only `build.rs` is used, Cargo always
     // considers the build dirty because `builtins-shim/build.rs` does not exist. If only
     // `../c-b/build.rs` is used, the same may happen if not built in the workspace.
+    // Note: this may be `None` if CARGO_MANIFEST_DIR doesn't end in a directory name (e.g. ".")
     if let Some(dir) = cfg.manifest_dir.file_name()
         && dir == "builtins-shim"
     {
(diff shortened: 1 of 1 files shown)


compiler-builtins/build.rs
```diff
@@ -12,7 +12,9 @@ fn main() {
     // Work around building as part of `builtins-shim`: if only `build.rs` is used, Cargo always
     // considers the build dirty because `builtins-shim/build.rs` does not exist. If only
     // `../c-b/build.rs` is used, the same may happen if not built in the workspace.
-    if cfg.manifest_dir.file_name().unwrap() == "builtins-shim" {
+    if let Some(dir) = cfg.manifest_dir.file_name()
+       && dir == "builtins-shim"
```
tgross35: Could you add a note that `file_name` may return `None` if `CARGO_MANIFEST_DIR` is a relative path?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1213/d529f15f670e9e8b3cc2d7306bf77e6e50d4e584..85f7a2dc3f2bf6b953845007b9fa2e8ef923f19f)*

DKLoehr: Added, though the issue isn't relative vs. absolute but whether or not it ends with a filename. The only useful example I can think of is `CARGO_MANIFEST_DIR = "."`.

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1213/d529f15f670e9e8b3cc2d7306bf77e6e50d4e584..24dd24c7860518433393c4586fbf2d17250d6218)*

zmodem:
@tgross35 is there a process for getting this into rust-lang/rust/ via a subtree update? (Does it happen automatically at some regular interval, is it by-request, or can we do it ourselves?)

tgross35:
Just a poke should be needed if there's something you are waiting on, sorry I missed this one. I usually only do bumps as-needed.

Updating in https://github.com/rust-lang/rust/pull/157037

PR #1215, opened 2026-05-17
symbol-check: Only ELFv1 ppc64 doesn't set .note.GNU-stack

Gelbpunkt:
`powerpc64-unknown-linux-gnu` currently is an ELFv1 target, while `powerpc64-unknown-linux-musl` is an ELFv2 target.

Big-endian and little-endian ELFv2 targets both behave normally: they emit `.note.GNU-stack`. Therefore, currently the tests would fail on big-endian powerpc64 with ELFv2 ABI.

To determine whether we need to special-case powerpc64, we should check the ABI instead of the endianness. The problem here is that the `e_flags` part of the ELF header is actually `0` in the output of `cc -O0 -ffunction-sections -fdata-sections -fPIC -m64 -mabi=elfv2 -Wall -Wextra -o missing_gnu_stack_section.o -c missing_gnu_stack_section.S`, the output of that command is bit-for-bit identical on ELFv1 and ELFv2 hosts. In order to know when to allow an unset `.note.GNU-stack` we therefore must set `.abiversion 2` to be able to tell them apart from the ELF header.

This makes all tests pass on `powerpc64-unknown-linux-musl`.

### DIFF
--- a/crates/symbol-check/src/main.rs
+++ b/crates/symbol-check/src/main.rs
@@ -13,8 +13,8 @@ use std::{env, fs};
 
 use object::read::archive::ArchiveFile;
 use object::{
-    Architecture, BinaryFormat, Endianness, File as ObjFile, Object, ObjectSection, ObjectSymbol,
-    Result as ObjResult, SectionFlags, Symbol, SymbolKind, SymbolScope, U32, elf,
+    Architecture, BinaryFormat, Endianness, File as ObjFile, FileFlags, Object, ObjectSection,
+    ObjectSymbol, Result as ObjResult, SectionFlags, Symbol, SymbolKind, SymbolScope, U32, elf,
 };
 use regex::Regex;
 use serde_json::Value;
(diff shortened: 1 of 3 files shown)


Gelbpunkt:
Successful powerpc64-unknown-linux-musl CI run: https://github.com/Gelbpunkt/compiler-builtins/actions/runs/25989194981/job/76392149895

crates/symbol-check/src/main.rs
```diff
@@ -526,12 +526,15 @@ fn check_elf_exe_stack(obj: &ObjFile) -> Result<(), ExeStack> {
...
+    };
+
     // If there is no `.note.GNU-stack` and no executable sections, behavior differs by platform.
     match obj.architecture() {
         // PPC64 doesn't set `.note.GNU-stack` since GNU nested functions don't need a trampoline,
-        // <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=21098>. From experimentation, it seems
-        // like this only applies to big endian.
-        Architecture::PowerPc64 if obj.endianness() == Endianness::Big => Ok(()),
+        // <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=21098>. This only applies to ELFv1.
+        Architecture::PowerPc64 if e_flags != 2 => Ok(()),
```
tgross35: I assume the upper bits aren't currently used but should this be masked by `EF_PPC64_ABI`?

tgross35 (commented):
Oh awesome, I'm very glad you were able to add some context because I was struggling to find it. One question then LGTM.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1215/3eb4dfa828bdafadf835bae9e53a165d9bc806dd..4e40e1cb1678e5188621b979e4807772a99fe0c5)*

crates/symbol-check/src/main.rs
Gelbpunkt: The ABI specification does not mention it, but both glibc and musl define it to `3` since the lower 2 bits are used, so yeah, doesn't hurt. I'll update it

tgross35 (approved):
Awesome, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1215/3eb4dfa828bdafadf835bae9e53a165d9bc806dd..4c5cdb7832f4dc760e89263c36c2123a9ccc93c7)*

PR #1222, opened 2026-06-23
int: Use leading_zeros_condset on LoongArch and RISC-V

heiher:
Rename `leading_zeros_riscv` to `leading_zeros_condset` and use it for LoongArch targets in addition to RISC-V. The new name reflects the implementation strategy rather than a specific architecture.

Like RISC-V, LoongArch provides set-if-less-than instructions that allow the algorithm to be implemented efficiently without branches, making the same approach beneficial on both architectures.

### DIFF
--- a/compiler-builtins/src/int/leading_zeros.rs
+++ b/compiler-builtins/src/int/leading_zeros.rs
@@ -4,9 +4,9 @@
 // Compilers will insert the check for zero in cases where it is needed.
 
 #[cfg(feature = "unstable-public-internals")]
-pub use implementation::{leading_zeros_default, leading_zeros_riscv};
+pub use implementation::{leading_zeros_condset, leading_zeros_default};
 #[cfg(not(feature = "unstable-public-internals"))]
-pub(crate) use implementation::{leading_zeros_default, leading_zeros_riscv};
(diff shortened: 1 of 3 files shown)


compiler-builtins/src/int/leading_zeros.rs
tgross35: Maybe want to update this line? And/or put in the doc comment the instructions missing/needed to make `_condset` more efficient than `_default`?

Otherwise LGTM

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1222/fb48f81544cf29e6ff7fb4468c8b2e1dacd42b79..c554989c7483138e9f54cd23899110e555da54f0)*

heiher: Thanks. Updated.

`_condset` is more efficient than `_default` here because it generates a more branchless sequence on RISC-V and LoongArch.

tgross35 (approved):
LGTM, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1222/fb48f81544cf29e6ff7fb4468c8b2e1dacd42b79..463549250a124f4560ddf70616bbf130b8666b70)*

PR #1223, opened 2026-06-23
int: Detect CLZ availability on LoongArch

heiher:
LoongArch provides the CLZ instruction as part of the standard `32s` feature set. Use `target_feature = "32s"` to determine whether CLZ instructions are available when selecting specialized division and remainder implementations.


tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1223/d746cbe15ad627be9f711d9f421ce168334e7632..75bc876dad240f8d757694545c30525d5a54631e)*

PR #1224, opened 2026-06-24
replace cfg_if with cfg_select

hkBst:
Now that cfg_select is stable, we can replace cfg_if. One difference is that cfg_select needs one branch to match. Therefore an empty default branch is sometimes needed.

EDIT: now using a local impl of cfg_select (thanks @folkertdev [1]) so as not to have to increase msrv.

[1]:[#t-libs > compiler-builtins msrv policy? @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/compiler-builtins.20msrv.20policy.3F/near/606320965)

### DIFF
--- a/libm/src/math/support/macros.rs
+++ b/libm/src/math/support/macros.rs
@@ -1,67 +1,25 @@
-/// `libm` cannot have dependencies, so this is vendored directly from the `cfg-if` crate
-/// (with some comments stripped for compactness).
-macro_rules! cfg_if {
+// Because `cfg_select` requires rust verion 1.95 and we support older versions,
+// we use the following replacement until that is no longer true.
+// FIXME(msrv)
+#[macro_export]
+macro_rules! cfg_select {
+    ({ $($tt:tt)* }) => {{
+        $crate::cfg_select! { $($tt)* }
+    }};
+    (_ => { $($output:tt)* }) => {
+        $($output)*
+    };
     (
-        if #[cfg( $($i_meta:tt)+ )] { $( $i_tokens:tt )* }
+        $cfg:meta => $output:tt
+        $($( $rest:tt )+)?
+    ) => {
+        #[cfg($cfg)]
+        cfg_select! { _ => $output }
         $(
-            else if #[cfg( $($ei_meta:tt)+ )] { $( $ei_tokens:tt )* }
-        )*
-        $(
-            else { $( $e_tokens:tt )* }
+            #[cfg(not($cfg))]
+            cfg_select! { $($rest)+ }
         )?
-    ) => {
-        cfg_if! {
-            @__items () ;
-            (( $($i_meta)+ ) ( $( $i_tokens )* )),
-            $(
-                (( $($ei_meta)+ ) ( $( $ei_tokens )* )),
-            )*
-            $(
-                (() ( $( $e_tokens )* )),
-            )?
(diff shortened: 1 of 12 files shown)


tgross35 (changes requested):
Thank you for the interest, but I don't think there is any need to do this. We can switch to `cfg_select` as minimum versions across the repo are raised, but the status quo works sufficiently well that there's no reason for `cfg_select` to be the driving factor.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1224/7eacd24b2fd65a7a3095b711c4e0dcae8a004ede..18ee5829d7a9aa4e2fef535fea3426be55c8a5ca)*

hkBst:
@tgross35 this now no longer needs an msrv bump due to a local impl, but should make switching to the real cfg_select trivial when msrv reaches 1.95+. What do you think?

tgross35:
This looks doable, thanks! I'm going to hold off until the patch for the formatting bug lands.

folkertdev:
What formatting bug? https://github.com/rust-lang/rust/issues/158371 is specific to the builtin `cfg_select` and should not affect a `macro_rules!` macro? or if it is broken, that fix won't fix it.

tgross35:
I thought this used the real `cfg_select` in some cases, but I guess that isn't the case.

libm/src/math/support/macros.rs
```diff
@@ -1,67 +1,22 @@
...
-            $(
-                (( $($ei_meta)+ ) ( $( $ei_tokens )* )),
-            )*
-            $(
-                (() ( $( $e_tokens )* )),
-            )?
-        }
+#[macro_export]
+macro_rules! cfg_select {
+    ({ $($tt:tt)* }) => {{
```
tgross35, line 6: Could you add a note about why we have this? Can just be a `// FIXME(msrv):`

tgross35 (changes requested):
One request then LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1224/7eacd24b2fd65a7a3095b711c4e0dcae8a004ede..fcf6ccc1fec5b6533ecce9b4455702f8eafc163c)*

libm/src/math/support/macros.rs
hkBst: Done!

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1224/40c1a7c88b959ca3bb8681c440fe15fd35aa35b0..73c5710f51c51909681b190515a0a7b66fba9faa)*

libm/src/math/support/macros.rs
```diff
@@ -1,67 +1,25 @@
...
-            )*
-            $(
-                (() ( $( $e_tokens )* )),
-            )?
-        }
+// Because `cfg_select` requires rust verion 1.95 and we support older versions,
+// we use the following replacement until that is no longer true.
+// FIXME(msrv)
+#[macro_export]
+macro_rules! cfg_select {
```
tgross35, line 5: This made the macro part of libm's public API. Fix in progress

PR #1225, opened 2026-06-28
tests: add tests for fdim(f)

dotacow:
* This PR aims to increase test coverage, taking a report generated by llvm-cov as a baseline reference.
* a comparison of the test coverage before and after:

```bash
> cargo +nightly llvm-cov -p compiler_builtins -p libm | grep fdim 1> baseline_cov.txt && cat baseline_cov.txt
math/fdim.rs                                      10                10     0.00%           2                 2     0.00%           6                 6     0.00%           0                 0         -
math/generic/fdim.rs                               5                 5     0.00%           1                 1     0.00%           3                 3     0.00%           0                 0         -
```
```bash
› cargo +nightly llvm-cov -p compiler_builtins -p libm | grep fdim 1> changed_cov.txt && cat changed_cov.txt
math/fdim.rs                                      78                 0   100.00%           7                 0   100.00%          38                 0   100.00%           0                 0         -
math/generic/fdim.rs                               5                 0   100.00%           1                 0   100.00%           3                 0   100.00%           0                 0         -

```

Sidenote, there are many branches where test coverage is low or nonexistent, I was thinking of opening an issue to track this and hopefully encourage others to add their own tests on branches lacking coverage, is such an issue welcome? I wouldn't want to clutter the project.

### DIFF
--- a/libm/src/math/fdim.rs
+++ b/libm/src/math/fdim.rs
@@ -2,8 +2,8 @@
 ///
 /// Determines the positive difference between arguments, returning:
 /// * x - y if x > y, or
-/// * +0    if x <= y, or
-/// * NAN   if either argument is NAN.
+/// * +0 if x <= y, or
+/// * NAN if either argument is NAN.
 ///
 /// A range error may occur.
 #[cfg(f16_enabled)]
@@ -16,8 +16,8 @@ pub fn fdimf16(x: f16, y: f16) -> f16 {
 ///
 /// Determines the positive difference between arguments, returning:
 /// * x - y if x > y, or
-/// * +0    if x <= y, or
-/// * NAN   if either argument is NAN.
+/// * +0 if x <= y, or
+/// * NAN if either argument is NAN.
 ///
 /// A range error may occur.
 #[cfg_attr(assert_no_panic, no_panic::no_panic)]
@@ -29,8 +29,8 @@ pub fn fdimf(x: f32, y: f32) -> f32 {
 ///
 /// Determines the positive difference between arguments, returning:
 /// * x - y if x > y, or
-/// * +0    if x <= y, or
-/// * NAN   if either argument is NAN.
+/// * +0 if x <= y, or
+/// * NAN if either argument is NAN.
 ///
 /// A range error may occur.
 #[cfg_attr(assert_no_panic, no_panic::no_panic)]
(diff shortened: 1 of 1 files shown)


libm/src/math/fdim.rs
```diff
@@ -42,12 +42,85 @@ pub fn fdim(x: f64, y: f64) -> f64 {
...
+        assert_eq!(fdimf128(5.0f128, 3.0f128), 2.0f128);
+        assert_eq!(fdimf128(3.0f128, 5.0f128), 0.0f128);
+    }
+
+    #[test]
+    #[cfg(f128_enabled)]
+    fn spec_tests_f128() {
+        spec_test::<f128>(fdimf128);
+    }
+}
```
tgross35, line 151: Use the macro form like at https://github.com/rust-lang/compiler-builtins/blob/7eacd24b2fd65a7a3095b711c4e0dcae8a004ede/libm/src/math/fma.rs#L57-L161, rather than the split spec/sanity tests. It makes things easier to read and I've been slowly moving things over.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1225/7eacd24b2fd65a7a3095b711c4e0dcae8a004ede..3fdf4ae640fc9fc845457ed0c5bafc81792dfb16)*

tgross35:
> Sidenote, there are many branches where test coverage is low or nonexistent, I was thinking of opening an issue to track this and hopefully encourage others to add their own tests on branches lacking coverage, is such an issue welcome? I wouldn't want to clutter the project.

We also have the extensive/exhaustive tests - did you happen to evaluate coverage via those? They should be much higher.

I'm generally happy to take test improvements like this one, but I don't think we need improving branch coverage to be a goal with an issue tracking it. Coverage isn't a great metric for this kind of code; it's easy to get to 100% coverage and still have major bugs because the output is so value-dependent. Simple tests like these that check some basic values and edge cases like infinity/nan are helpful, though, and the integration tests should get the rest.

tgross35 (changes requested):
(updating status)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1225/7eacd24b2fd65a7a3095b711c4e0dcae8a004ede..3fdf4ae640fc9fc845457ed0c5bafc81792dfb16)*

dotacow:
> > Sidenote, there are many branches where test coverage is low or nonexistent, I was thinking of opening an issue to track this and hopefully encourage others to add their own tests on branches lacking coverage, is such an issue welcome? I wouldn't want to clutter the project.
> 
> We also have the extensive/exhaustive tests - did you happen to evaluate coverage via those? They should be much higher.
> 
> I'm generally happy to take test improvements like this one, but I don't think we need improving branch coverage to be a goal with an issue tracking it. Coverage isn't a great metric for this kind of code; it's easy to get to 100% coverage and still have major bugs because the output is so value-dependent. Simple tests like these that check some basic values and edge cases like infinity/nan are helpful, though, and the integration tests should get the rest.

I tried running the coverage script on the extensive set and couldn't really come up with a meaningful metric, it is obvious to me now that a coverage metric is arbitrary in this case. Nonetheless, this is why I opened the (relatively trivial) PR rather than opening a tracking issue.
I'll get around to doing the requested changes and re-ping. thanks for your review.

libm/src/math/fdim.rs
```diff
@@ -42,12 +42,84 @@ pub fn fdim(x: f64, y: f64) -> f64 {
...
+        ($f:ty) => {
+            [
+                // Sanity checks
+                (5.0, 3.0, 2.0),
+                (3.0, 5.0, 0.0),
+                // Spec tests
+                (1.0, 0.0, 1.0),
+                (0.0, -1.0, 1.0),
+                (<$f>::INFINITY, 1.0, <$f>::INFINITY),
+                (1.0, <$f>::NEG_INFINITY, <$f>::INFINITY),
```
tgross35: Could you also check the inverses? (1, inf) and (-inf, 1)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1225/7eacd24b2fd65a7a3095b711c4e0dcae8a004ede..1af08259c58b25318c5f0188f0f0cc5a1d335f56)*

libm/src/math/fdim.rs
```diff
@@ -42,12 +42,84 @@ pub fn fdim(x: f64, y: f64) -> f64 {
...
+                (0.0, 1.0, 0.0),
+                (-1.0, 0.0, 0.0),
+                (<$f>::NEG_INFINITY, 0.0, 0.0),
+                (0.0, 0.0, 0.0),
+                (-0.0, 0.0, 0.0),
+                (0.0, -0.0, 0.0),
```
tgross35, line 89: Can you add (-0, -0)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1225/7eacd24b2fd65a7a3095b711c4e0dcae8a004ede..1af08259c58b25318c5f0188f0f0cc5a1d335f56)*

libm/src/math/fdim.rs
```diff
@@ -42,12 +42,84 @@ pub fn fdim(x: f64, y: f64) -> f64 {
...
+                (1.0, 1.0, 0.0),
+                (<$f>::INFINITY, <$f>::INFINITY, 0.0),
+                (<$f>::NEG_INFINITY, <$f>::NEG_INFINITY, 0.0),
+                // NaN inputs
+                (<$f>::NAN, 1.0, <$f>::NAN),
+                (1.0, <$f>::NAN, <$f>::NAN),
+                (<$f>::NAN, <$f>::NAN, <$f>::NAN),
```
tgross35: What's the sorting pattern here :) This may as well check the full matrix, i.e.:

```
(-inf, -1),
(-inf, -0),
(-inf, 0),
(-inf, 1),
(-inf, inf),
(-inf, nan),
(-1, -1),
(-1, -0),
(-1, 0),
(-1, 1),
(-1, inf),
(-1, nan),
(-0, -1),
(-0, -0),
(-0, 0),
(-0, 1),
(-0, inf),
(-0, nan),
... repeat for 0, 1, inf, nan on the LHS
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1225/7eacd24b2fd65a7a3095b711c4e0dcae8a004ede..1af08259c58b25318c5f0188f0f0cc5a1d335f56)*

libm/src/math/fdim.rs
```diff
@@ -42,12 +42,84 @@ pub fn fdim(x: f64, y: f64) -> f64 {
...
+    fn check<F: Float>(f: fn(F, F) -> F, cases: &[(F, F, F)]) {
+        for &(x, y, exp_res) in cases {
+            let val = f(x, y);
+            if exp_res.is_nan() {
+                assert!(
+                    val.is_nan(),
+                    "fdim({x:?}, {y:?}) expected NaN, got {val:?} ({} {})",
+                    Hex(x),
+                    Hex(y)
+                );
```
tgross35: Are the NANs getting modified, such that `assert_biteq!` doesn't work?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1225/7eacd24b2fd65a7a3095b711c4e0dcae8a004ede..1af08259c58b25318c5f0188f0f0cc5a1d335f56)*

libm/src/math/fdim.rs
tgross35: Whoops, covered in the other comment

libm/src/math/fdim.rs
tgross35: Whoops, meant to remove

libm/src/math/fdim.rs
dotacow: let me know if I am wrong, but the way I understand it ([per this link](https://doc.rust-lang.org/std/primitive.f32.html#associatedconstant.NAN:~:text=This%20constant%20is,and%20target%20platforms.)) 2 NaNs may not be equal-bitwise, so best to just use `assert!` in combination with `is_nan()`, if I am wrong let me know and I'll change it to `assert_biteq`
>...the quiet/signaling bit being set to 1 indicates a quiet NaN). Beyond that, nothing is guaranteed about the specific bit pattern chosen here: both payload and sign are arbitrary. The concrete bit pattern may change across Rust versions and target platforms.

tgross35: It's true that the result of 0.0/0.0 is allowed to change, though I'd never expect this in practice, and doing `is_nan` is the right thing to do most places. But this library does trickier things with NaNs than most places :) and generally we have to be able to assume that even if std gave us an uncommon NaN, it would still roundtrip though `libm`'s APIs. (There are exceptions but I don't think anything that applies here.)

libm/src/math/fdim.rs
dotacow: I added the rest of the cases (AI disclosure: I did get an AI model to generate the permutations for me)

libm/src/math/fdim.rs
dotacow: sounds good to me. I pushed the refactor.

libm/src/math/fdim.rs
```diff
@@ -42,12 +42,120 @@ pub fn fdim(x: f64, y: f64) -> f64 {
...
+                assert_biteq!(
+                    val,
+                    exp_res,
+                    "fdim({x:?}, {y:?}) expected NaN, got {val:?} ({} {})",
+                    Hex(x),
+                    Hex(y)
+                );
+            } else {
+                assert_biteq!(val, exp_res, "fdim({x:?}, {y:?}) ({} {})", Hex(x), Hex(y));
+            }
```
tgross35: These can just be merged, `Hex` makes it clear that they're NaNs (and also shows their payload, e.g. https://github.com/rust-lang/compiler-builtins/blob/1521862f2e486ff58b4ce8c92bfadffbc8736f35/libm/src/math/support/hex_float.rs#L1189-L1204)

Otherwise LGTM assuming checks pass

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1225/7eacd24b2fd65a7a3095b711c4e0dcae8a004ede..bcc731db25c4752d1e93f01864c39c6d187107f9)*

dotacow: CI is green, Thanks for your review!

tgross35: ^^ Sorry for not being clear, could you delete the `if`/`else`? They can both just use the `assert_biteq!(val, exp_res, "fdim({x:?}, {y:?}) ({} {})", Hex(x), Hex(y));`

dotacow: Oh that's so obvious, thanks for pointing it out 😅

dotacow: @tgross35 quiet ping (=

tgross35 (approved):
Looks great, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1225/5ef80b32f957e516a3f6656ace81fdf37ce22220..a003ae290dbac21cb8fe5e04deae10ceb1c38e46)*

PR #1227, opened 2026-07-02
 ci: build and test with s390x-resolute runner

marcoieni:
We'd like to this add new `s390x` runner to `compiler-builtins`. 

This new runner is provided by Canonical. After some iteration alongside with Canonical folks, the `large` runner offers hardware spec similar to the existing s390x provided by IBM and [delivers similar build times](https://github.com/rust-lang/compiler-builtins/actions/runs/31516313575/job/93862267593?pr=1227).

Moreover, it runs on ubuntu-26.04 rather than ubuntu-24.04, and it features a Github integration more friendly to `t-infra`, since the related Github App requires less permissions to run. 

We don't need to remove the s390x IBM runners right now. We propose having both s390x runners running side by side for a while and circle back after a few PRs, sticking with the Canonical one afterwards if everything goes well.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -93,6 +93,8 @@ jobs:
         - target: riscv64gc-unknown-linux-gnu
           os: ubuntu-26.04
         - target: s390x-unknown-linux-gnu
+          os: self-hosted-linux-s390x-resolute-large-rust # resolute == ubuntu-26.04
+        - target: s390x-unknown-linux-gnu
           os: ubuntu-24.04-s390x
         - target: thumbv6m-none-eabi
           os: ubuntu-26.04
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -92,6 +92,8 @@ jobs:
         - target: riscv64gc-unknown-linux-gnu
           os: ubuntu-24.04
         - target: s390x-unknown-linux-gnu
+          os: self-hosted-linux-s390x-resolute-medium-rust
+        - target: s390x-unknown-linux-gnu
           os: ubuntu-24.04-s390x
         - target: thumbv6m-none-eabi
           os: ubuntu-24.04


tgross35:
(rebased to pick up 5ef80b32f957e516a3f6656ace81fdf37ce22220 and see where it's getting stuck)

.github/workflows/main.yaml
```diff
@@ -91,6 +91,8 @@ jobs:
         #   os: ["self-hosted", "linux", "riscv64"]
         - target: riscv64gc-unknown-linux-gnu
           os: ubuntu-24.04
+        - target: s390x-resolute-medium-rust
```
tgross35: Oh - this needs to be the actual rust target

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1227/5ef80b32f957e516a3f6656ace81fdf37ce22220..19be4b45afe377a5786104278d64c636adb797a9)*

marcoieni:
the new runner might be slow due to the cache miss, I want to try to push a new commit to check how it behaves with the cache hit.

tgross35:
It's building nextest from source since there's no binary for this arch, which is taking extra long too (though seems like the ubuntu-24.04-s390x image is reasonably fast?). If trying again with cache doesn't speed things up much, it might be worth an `if:` to disable on this runner at https://github.com/rust-lang/compiler-builtins/blob/5ef80b32f957e516a3f6656ace81fdf37ce22220/.github/workflows/main.yaml#L152-L154.

(edit: install-action actually has `fallback: none`)

marcoieni:
we might ask Canonical if they can install cargo next test on the runner for us so that it is always available. Although we might not be able to pin the version 🤔

tgross35:
I'm fine without an exact version as long as it's relatively recent, the pinning is mostly for Zizmor rather than needing specific features

marcoieni:
Actually if we want to control the version, we could cache it 👀

Btw, there wasn't a cache hit even in the second run because I think the cache is populated only from the main branch. 

This Canonical runner seems slower than the IBM one because to run `cargo install` of cargo nextest, it takes almost four times more.
So I assume there's no reason to merge this PR, since the s390x architecture is already covered by the ibm runner, right?

See: 
image
image

tgross35:
> Actually if we want to control the version, we could cache it 👀

Just building once then putting it on ci-mirrors? That does sound nice 🤔. The nextest maintainers have been pretty open to adding arches if they get the CI resources, maybe we should ask if they'd apply to the IBM program.

> Btw, there wasn't a cache hit even in the second run because I think the cache is populated only from the main branch.

I would have expected this to work, but the inner workings of GHA cache are a complete mystery to me 🙂 

> This Canonical runner seems slower than the IBM one because to run `cargo install` of cargo nextest, it takes almost four times more. So I assume there's no reason to merge this PR, since the s390x architecture is already covered by the ibm runner, right?

Up to you for what you're trying to get. If you wanted a persistent job to keep testing then we can figure out what to do with nextest, but the existing s390x job covers things well so it's also fine to close.

marcoieni:
> Just building once then putting it on ci-mirrors? That does sound nice 🤔

I was thinking more of caching the artifacts in github actions cache. It's easier and requires less maintenance. Similar to how we cache cargo artifacts now.



> The nextest maintainers have been pretty open to adding arches if they get the CI resources, maybe we should ask if they'd apply to the IBM program.

this is also an option!

tgross35:
I thought rust-cache takes care of those since it should cache the bin dir (I think?), but it also looks like the other s390x job had to rebuild. Hmm

marcoieni:
I think the issue is the order of the steps, since the rust-cache action runs after the install action.

ubiratansoares:
Aside caching, I found an interesting data point from [this workflow run](https://github.com/rust-lang/compiler-builtins/actions/runs/29012471561) : it seems the IBM Runners have a significantly bigger CPU compared with the Canonical `resolute-medium` ones : 8 cores (vs 4 cores), bigger CPU caches, etc.  I got this information by expanding logs from [this step](https://github.com/rust-lang/compiler-builtins/actions/runs/29012471561/job/86099617776#step:2:27) in both jobs.

I'd not be surprised if the IBM runner has bigger numbers for other hardware parameters like RAM/Disk sizes, IO perf, etc. That could explain the difference between the two jobs executions, at least partially.

tgross35 (approved):
Looks like time is comparable to the other s390 job now. Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1227/7f4c82e1ef14ad14945bf4a2794dfeda80b08226..1e4d4b4752e152b36366de37dac3f3f2bc7aee59)*

tgross35:
Actually I guess that one was fast because the cache was available, for this cold build it's taking a very long time. I think I'd like to wait for https://github.com/rust-lang/compiler-builtins/pull/1253 which should get rid of the time spent building nextest (cc @Gelbpunkt)

ubiratansoares:
@tgross35 since https://github.com/rust-lang/compiler-builtins/pull/1253 is merged, are we good to merge this one?

tgross35 (approved):
Looks like the cold build time is now ~11 minutes which is pretty tolerable. Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1227/5f0cf29dad55ae06a68303179700903a9f7465c0..1c4dc911c8a52247a084bd225f4f836089e57dd7)*

PR #1228, opened 2026-07-03
libm: Add LoongArch implementations of fma, rint, and sqrt

heiher:
Add LoongArch-specific implementations of `fma`, `rint`, and `sqrt` using hardware floating-point instructions and enable architecture dispatch for these operations.

### DIFF
--- a/etc/function-definitions.json
+++ b/etc/function-definitions.json
@@ -343,6 +343,7 @@
     "fma": {
         "sources": [
             "libm/src/math/arch/aarch64/fma.rs",
+            "libm/src/math/arch/loongarch/fma.rs",
             "libm/src/math/arch/x86/fma.rs",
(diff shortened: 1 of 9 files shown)


tgross35 (approved):
I'm not familiar with loongarch asm syntax but this all looks reasonable. Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1228/14155c5c282f2d65f1e2336b1fb3f7473ecf537c..0ddc11ca84c64c2c600ee3758d590e3fe026c8df)*

tgross35:
Does loongarch have f16 instructions by the way?

heiher:
> Does loongarch have f16 instructions by the way?

The LoongArch only has `f16` and `f32` conversion instructions.

Issue #836, opened 2025-01-06
Specific test failures


tgross35:
Just making a list of failures that should be turned into xfails or fixed.

My loose plan is to put all of these into a single test file and assert that they fail, then if they get fixed they can be converted to test cases.

---

i586-unknown-linux-gnu from on https://github.com/rust-lang/libm/pull/379, musl's result is more accurate.

```text
thread 'musl_random_asinhf' panicked at crates/libm-test/tests/compare_built_musl.rs:43:51:
called `Result::unwrap()` on an `Err` value:
    input:    (-0.37330312,) (0xbebf2196,)
    expected: -0.36513534            0xbebaf305
    actual:   -0.36513525            0xbebaf302
```

---

i686-unknown-linux-gnu from on https://github.com/rust-lang/libm/pull/379, musl has the incorrect result here. Not sure why this would be, is compiler-rt incorrect?

```text
 thread 'musl_random_fdim' panicked at crates/libm-test/tests/compare_built_musl.rs:43:51:
called `Result::unwrap()` on an `Err` value: 
    input:    (5.1720626866622296e88, -5.1749184040104636e103) (0x5259ffdf64f0d2f8, 0xd5771addd21280b5)
    expected: 5.174918404010469e103  0x55771addd21280bc
    actual:   5.174918404010468e103  0x55771addd21280bb

Caused by:
    ulp 1 > 0
```

tgross35:
i586 at https://github.com/rust-lang/libm/actions/runs/12639597612/job/35218274466?pr=404

```text
---- mp_random_fdim stdout ----
Random Mpfr fdim arg 1/2: 100 iterations (10000 total) using `LIBM_SEED=xcGcTkg83bx6IACwhIJGBEnglUPE38Ik`
Random Mpfr fdim arg 2/2: 100 iterations (10000 total) using `LIBM_SEED=xcGcTkg83bx6IACwhIJGBEnglUPE38Ik`

thread 'mp_random_fdim' panicked at crates/libm-test/tests/multiprecision.rs:41:49:
called `Result::unwrap()` on an `Err` value: 
    input:    (2.562129982717898e-287, -1.64014759986637e-294) (0x046f3600eb4ad0e0, 0x82f0c29b2b40023d)
    expected: 2.562130146732658e-287 0x046f36010cd00737
    actual:   2.5621301467326578e-287 0x046f36010cd00736

Caused by:
    ulp 1 > 0
```
(test case added)

[...]

PR #1231, opened 2026-07-04
Revert "Avoid inlining `floor` into `rem_pio2`"

tgross35:
Starting with LLVM version 22 `nounwind` is now propagated from inline
assembly to functions containing it, meaning this workaround can now be
removed.

This reverts commit c4966f9b0d4323183933e98dea4d0a77f52c0e0a.

Link: https://github.com/llvm/llvm-project/pull/150958
Link: https://github.com/rust-lang/rust/issues/144518

### DIFF
--- a/libm/src/math/rem_pio2_large.rs
+++ b/libm/src/math/rem_pio2_large.rs
@@ -13,7 +13,7 @@
  * ====================================================
  */
 
-use super::scalbn;
+use super::{floor, scalbn};
 
 // initial value for jk
 const INIT_JK: [usize; 4] = [3, 4, 4, 6];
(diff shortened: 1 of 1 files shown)


tgross35:
Cc @quaternic as you authored c4966f9b0d4323183933e98dea4d0a77f52c0e0a

quaternic:
Should this wait at least until rustc drops support for LLVM 21? Extrapolating from the last few version bumps, I'd guess that might be around Rust 1.100

-----

The CI failure in a randomized test is another example of https://github.com/rust-lang/compiler-builtins/issues/836#issuecomment-4123516959 (`powi(~1.0, large)` can be inaccurate), and is unrelated to this change.

tgross35:
Good point about the LLVM version, I'll hold off.

I wish I had a better way to skip crazy values on that test without skipping unconditionally, but I have no idea how to do that without knowing a mathematical upper bound...

PR #1233, opened 2026-07-09
ci: Move thumbv6-non-eabi to a regular test job

tgross35:


tgross35:
Non-dist target, I don't want to deal with build-std

PR #1234, opened 2026-07-11
 arm_linux: Add missing memory barrier to atomic_cmpxchg 

taiki-e:
This PR contains three commits. The main fix is the last one, and the first two address minor issues such as documentation that I noticed when writing the main fix. I would suggest reviewing by commit, but I can split them into separate PRs if necessary. 

---

Currently, atomic_cmpxchg returns the value obtained from the first relaxed load as-is without calling the kuser helper when comparison fails:

https://github.com/rust-lang/compiler-builtins/blob/5af7a65c0342d3f5ac1d250c7902aaabbfc2bd49/compiler-builtins/src/sync/arm_linux.rs#L130-L134

This is equivalent to `compare_exchange(SeqCst, Relaxed)`, but the `__sync` builtins are expected to have SeqCst semantics in both the success and failure paths, so this is unsound.

This PR fixes this by adding a memory barrier to failure path:

https://github.com/rust-lang/compiler-builtins/blob/c0cdefbe886807089b9a38b3e21a80060caa0472/compiler-builtins/src/sync/arm_linux.rs#L134-L141

Context and Disclosure: This bug had been observed for several years as a rare stress test failure in portable-atomic's CI when testing pre-v6 ARM Linux on ARM hardware (https://gist.github.com/taiki-e/31962f6efe8cc9bc06aa6869bbceb656). In my past investigations, I suspected the bug was with the algorithm used in portable-atomic, so I was unable to identify the cause, but I recently identified the cause by investigating a wider scope with the help of LLMs. (All changes in this PR were written by me.)

### DIFF
--- a/compiler-builtins/src/sync/arm_linux.rs
+++ b/compiler-builtins/src/sync/arm_linux.rs
@@ -5,15 +5,17 @@ use core::{arch, mem};
 // https://www.kernel.org/doc/Documentation/arm/kernel_user_helpers.txt
 unsafe fn __kuser_cmpxchg(oldval: u32, newval: u32, ptr: *mut u32) -> bool {
     // FIXME(volatile): the third parameter is a volatile pointer
-    // SAFETY: kernel docs specify a known address with the given signature
+    // SAFETY: kernel docs specify a known address with the given signature.
+    // And the caller must guarantee that the pointer is valid for read and write
+    // and aligned to the element size.
     let f = unsafe {
         mem::transmute::<_, extern "C" fn(u32, u32, *mut u32) -> u32>(0xffff0fc0usize as *const ())
     };
     f(oldval, newval, ptr) == 0
 }
 
-unsafe fn __kuser_memory_barrier() {
-    // SAFETY: kernel docs specify a known address with the given signature
+fn __kuser_memory_barrier() {
+    // SAFETY: kernel docs specify a known address with the given signature.
     let f = unsafe { mem::transmute::<_, extern "C" fn()>(0xffff0fa0usize as *const ()) };
     f();
 }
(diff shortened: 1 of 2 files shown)


tgross35 (approved):
This looks reasonable to me, couple questions for my understanding:

1. This only shows up when multiple atomics are involved, right (hence seqcst) ? I _think_ the single-atomic pathing was fine here but can't convince myself of it.
2. Is there a C implementation somewhere that we can reference?

Also I'm planning to add the middle bit of your PR message into the third commit's description so there's some context in the history, feel free to do this yourself if you prefer.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1234/5af7a65c0342d3f5ac1d250c7902aaabbfc2bd49..c0cdefbe886807089b9a38b3e21a80060caa0472)*

taiki-e:
> 1. This only shows up when multiple atomics are involved, right (hence seqcst) ? I _think_ the single-atomic pathing was fine here but can't convince myself of it.

 ARM is an architecture with a weak memory model, so the absence of a fence here affects SeqCst, AcqRel, and Acquire. 

Actually, the [code that encountered this bug that led to this investigation used Acquire load](https://github.com/taiki-e/portable-atomic/blob/5716e951b47494703a3926403c27a20add27fa9b/src/imp/fallback/seq_lock_wide.rs#L43-L64) (which, in pre-v6 Arm, this is implemented in CAS(0, 0)).

> 2. Is there a C implementation somewhere that we can reference?

It appears that libgcc has the same bug: https://github.com/gcc-mirror/gcc/blob/32657f29f918712ad7110bd54d8ebb3bf6b0a2a8/libgcc/config/arm/linux-atomic.c#L209

@Amanieu, who added these to compiler-builtins in https://github.com/rust-lang/compiler-builtins/pull/115, might know more about this.

Amanieu:
The implementation here should match libgcc, I would be more comfortable if this was first reported upstream to get a confirmation.

PR #1236, opened 2026-07-19
compiler-builtins: remap RUSTC_DEBUGINFO_MAP paths via -ffile-prefix-map

paradoxicalguy:
`libcompiler_builtins.rlib` leaks the local build path into `DW_AT_comp_dir` for C-compiled units (compiler-rt), even with rust-lang/rust's `rust.remap-debuginfo = true`. 
existing `-ffile-prefix-map={root}=.` only covers the compiler-rt source root, not `comp_dir`.

fix: iterate `RUSTC_DEBUGINFO_MAP` and apply each mapping via `-ffile-prefix-map`, mirroring `rustc_llvm`'s `build.rs`.

before: `DW_AT_comp_dir ("/rustc/08f62523b1c3262cc2f0288a3369b760ab4bda45/library/compiler-builtins/compiler-builtins")`
after: `DW_AT_comp_dir ("/rustc/08f62523b1c3262cc2f0288a3369b760ab4bda45")`

[see discussion](https://github.com/rust-lang/rust/pull/156570)

r? @Urgau 
cc @tgross35


compiler-builtins/build.rs
```diff
@@ -538,7 +538,22 @@ mod c {
...
+        
+        // `RUSTC_DEBUGINFO_MAP` is a bootstrap-specific env var (not a general
+        // Cargo/rustc mechanism) containing the `--remap-path-prefix` mappings
+        // bootstrap passes to rustc via its rustc shim. Bootstrap uses this
+        // instead of RUSTFLAGS because Cargo doesn't always forward RUSTFLAGS to
+        // build scripts and proc-macros (rust-lang/cargo#4423).
+        //
+        // Apply each mapping via -ffile-prefix-map, mirroring rustc_llvm's
+        // build.rs for the same env var.
+        if let Some(maps) = env::var_os("RUSTC_DEBUGINFO_MAP")
```
Mark-Simulacrum, line 550: I am not a fan of this living outside of rust-lang/rust. That usually leads to confusion and pain when we want to change the interface.

Can you say more about what information is missing in Cargo to allow reproducible artifacts from compiler-builtins? If root.display() above is not enough, what would be enough?

Somewhat relatedly, if rust-lang/rust's bootstrap feels this should be threaded down, why can't it use CFLAGS / CXXFLAGS to do so, rather than a custom env variable?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1236/dfe80e1d870b56f34a52bd8b3ce47479cdb95a2d..9d57d705fb12603f34989d22e60cae6d783476b5)*

paradoxicalguy: `root.display()` only covers the `compiler-rt` source root, it doesn't cover this crate's own directory (`CARGO_MANIFEST_DIR`), which is what `DW_AT_comp_dir` was leaking. 

on CFLAGS/CXXFLAGS, i don't know why bootstrap uses a custom env var instead of that. i'll check with @Urgau , who was [explaining](https://github.com/rust-lang/rust/pull/156570#discussion_r3605618567) the `RUSTFLAGS`-forwarding issue (rust-lang/cargo#4423) that led to `RUSTC_DEBUGINFO_MAP` existing in the first place. not sure if the same limitation applies to CFLAGS/CXXFLAGS.

happy to move this into rust-lang/rust directly if that's preferred. wasn't sure which repo was the right home for it given compiler-builtins is developed separately.

tgross35:
Sounds like this may be something to update elsewhere, https://github.com/rust-lang/rust/pull/156570#issuecomment-5019940595

paradoxicalguy:
closing this
fixed via [bootstrap](https://github.com/rust-lang/rust/pull/159673)

PR #1241, opened 2026-07-28
hexagon: use a single global_asm! block for source-order stability

### LINKED ISSUE #1240: compile compiler_builtins failed for target hexagon-unknown-linux-musl
rustc 1.99.0-nightly (09ee43b2d 2026-07-27)
target: hexagon-unknown-linux-musl
build with -Zbuild-std
```
error: unrecognized instruction
  ╰╴
note: instantiated into assembly here
  ╭▸ <inline asm>:3:1
  │
3 │ FUNCTION_BEGIN __hexagon_modsi3
  ╰╴━

error: unrecognized instruction
   ╰╴
note: instantiated into assembly here
   ╭▸ <inline asm>:41:1
   │
41 │ FUNCTION_END __hexagon_modsi3
   ╰╴━

error: unrecognized instruction
    ╰╴
note: instantiated into assembly here
    ╭▸ <inline asm>:888:1
    │
888 │ FUNCTION_BEGIN __hexagon_memcpy_likely_aligned_min32bytes_mult8bytes
    ╰╴━

error: unrecognized instruction
    ╰╴
note: instantiated into assembly here
    ╭▸ <inline asm>:918:1
     ╰╴━

error: unrecognized instruction
     ╰╴
note: instantiated into assembly here
     ╭▸ <inline asm>:1334:1
     │
1334 │ FUNCTION_BEGIN __hexagon_udivdi3
     ╰╴━

error: unrecognized instruction
     ╰╴
note: instantiated into assembly here
     ╭▸ <inline asm>:1378:1
     │
1378 │ FUNCTION_END __hexagon_udivdi3
     ╰╴━

error: unrecognized instruction
     ╰╴
note: instantiated into assembly here
     ╭▸ <inline asm>:1384:1
     │
1384 │ FUNCTION_BEGIN __hexagon_udivsi3
     ╰╴━

error: unrecognized instruction
     ╰╴
note: instantiated into assembly here
     ╭▸ <inline asm>:1414:1
     │
1414 │ FUNCTION_END __hexagon_udivsi3
     ╰╴━

error: unrecognized instruction
     ╰╴
note: instantiated into assembly here
     ╭▸ <inline asm>:1418:1
     │
1418 │ FUNCTION_BEGIN __hexagon_sqrtf
     ╰╴━

error: unrecognized instruction
     ╰╴
note: instantiated into assembly here
     ╭▸ <inline asm>:1462:1
     │
1462 │ FUNCTION_END __hexagon_sqrtf
     ╰╴━

error: could not compile `compiler_builtins` (lib) due to 32 previous errors
```

androm3da:
Since rust-lang/rust#145358 ("Sort mono items by symbol name", merged 2025-08-22), rustc sorts mono items within a codegen unit by symbol name rather than source order unless -Zcodegen-source-order is passed.

That breaks func_macro.s, which defines asm macros for use by "later" files.

Merge all includes into one global_asm! call. A single MonoItem cannot be reordered against itself, so the source order becomes the order the assembler sees. The file set and its order are unchanged, and all these items already shared one codegen unit, so object-file granularity is unaffected.

Closes: https://github.com/rust-lang/compiler-builtins/issues/1240

### DIFF
--- a/compiler-builtins/src/hexagon.rs
+++ b/compiler-builtins/src/hexagon.rs
@@ -44,69 +44,34 @@ intrinsics! {
     }
 }
 
-global_asm!(include_str!("hexagon/func_macro.s"), options(raw));
-
+// All hexagon assembly must live in ONE `global_asm!` invocation, and
+// `func_macro.s` must come first.
 global_asm!(
+    include_str!("hexagon/func_macro.s"),
     include_str!("hexagon/common_entry_exit_abi1.s"),
-    options(raw)
-);
-
-global_asm!(
     include_str!("hexagon/common_entry_exit_abi2.s"),
-    options(raw)
-);
-
-global_asm!(
     include_str!("hexagon/common_entry_exit_legacy.s"),
-    options(raw)
-);
-
-global_asm!(include_str!("hexagon/dfaddsub.s"), options(raw));
-
-global_asm!(include_str!("hexagon/dfdiv.s"), options(raw));
-
-global_asm!(include_str!("hexagon/dffma.s"), options(raw));
-
-global_asm!(include_str!("hexagon/dfminmax.s"), options(raw));
-
-global_asm!(include_str!("hexagon/dfmul.s"), options(raw));
-
-global_asm!(include_str!("hexagon/dfsqrt.s"), options(raw));
-
-global_asm!(include_str!("hexagon/divdi3.s"), options(raw));
-
-global_asm!(include_str!("hexagon/divsi3.s"), options(raw));
-
-global_asm!(include_str!("hexagon/fastmath2_dlib_asm.s"), options(raw));
-
-global_asm!(include_str!("hexagon/fastmath2_ldlib_asm.s"), options(raw));
-
-global_asm!(
+    include_str!("hexagon/dfaddsub.s"),
+    include_str!("hexagon/dfdiv.s"),
+    include_str!("hexagon/dffma.s"),
+    include_str!("hexagon/dfminmax.s"),
+    include_str!("hexagon/dfmul.s"),
+    include_str!("hexagon/dfsqrt.s"),
+    include_str!("hexagon/divdi3.s"),
+    include_str!("hexagon/divsi3.s"),
+    include_str!("hexagon/fastmath2_dlib_asm.s"),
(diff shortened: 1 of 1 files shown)


androm3da:
Fixes #1240

tgross35 (approved):
LGTM, thanks. Looks like that PR merged almost a year ago, any idea why there wasn't any problem until now?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1241/9c7d3e8d064e331144dbe3426a0bf5311399a857..75835e0208779bcb7c3f9135ca5352de426f0051)*

bjorn3:
Be aware that macros are technically not allowed in `global_asm!()`. They are not on the list of allowed directives at https://doc.rust-lang.org/reference/inline-assembly.html#directives-support and for every stateful directive (which includes macros) you have to undo their effect before exitting the `global_asm!()` block. So at the very least you did have to use `.push_section`/`.pop_section` rather than `.text` and undefine all macros at the end (is that last one even possible?).

androm3da:
> Be aware that macros are technically not allowed in `global_asm!()`. They are not on the list of allowed directives at https://doc.rust-lang.org/reference/inline-assembly.html#directives-support and for every stateful directive (which includes macros) you have to undo their effect before exitting the `global_asm!()` block. So at the very least you did have to use `.push_section`/`.pop_section` rather than `.text` and undefine all macros at the end (is that last one even possible?).

Okay - good point.  I liked having the asm here resemble the corresponding stuff in compiler-rt.  But I'm fine w/remediating this by fixing the asm to no longer depend on the `.macro` directive.

androm3da:
> LGTM, thanks. Looks like that PR merged almost a year ago, any idea why there wasn't any problem until now?

Sadly, we have not done a good enough job maintaining the hexagon targets. 😖

bjorn3:
You still want to avoid code duplication you could use a rust macro that you invoke inside the `global_asm!()`. Something like `global_asm!(function_begin!("foo"), "contents", function_end!("foo"), function_begin!("bar"), "contents2", function_end!("foo"))` or `global_asm!(function!("foo", "contents"), function!("bar", "contents2"))`. If you do it that way, you can probably split them into separate `global_asm!()` blocks again if you want.

PR #1242, opened 2026-07-28
cargo: update comment for using workspace-inherited dependency

xtqqczze:


xtqqczze:
Error from logs:

```sh
+ cargo build --manifest-path libm/Cargo.toml
error: failed to parse manifest at `/home/runner/work/compiler-builtins/compiler-builtins/libm/Cargo.toml`

Caused by:
  failed to find a workspace root
```

This is the problem:

https://github.com/rust-lang/compiler-builtins/blob/1c37df1ead4667d895dffc23900503507a9ad193/.github/workflows/main.yaml#L361-L363

But if we don't remove the workspace Cargo.toml we get the following error:

```log
error: failed to load manifest for workspace member `~/rust-lang-compiler-builtins/builtins-shim`

Caused by:
  failed to parse manifest at `~/rust-lang-compiler-builtins/builtins-shim/Cargo.toml`

Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2024` edition, and only supports `2015`, `2018`, and `2021` editions.
  ```
  
  So using workspace-inherited dependencies here depends on a MSRV of 1.85.

tgross35:
I think we could figure out some hack here, but it's not going to be worth it for a one line change

xtqqczze:
ok let's just update the comment

tgross35:
No need to update the comment, upgrading the edition is also tied to the MSRV

xtqqczze:
Right, I suppose `FIXME(msrv` is easier to grep for. I’ve also added the version to make it even clearer.

tgross35 (approved):

PR #1253, opened 2026-08-07
ci: Disable install-action fallback for nextest

Gelbpunkt:
This should make it fall back to a system-wide cargo-nextest install or running the tests without nextest, which is probably still faster than building nextest from source.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -140,8 +140,13 @@ jobs:
       run: ./ci/install-test-deps.sh "$JOB_TARGET" "$JOB_CHANNEL" "$RUN_IN_DOCKER"
 
     - uses: taiki-e/install-action@7f4eb899022d8fe70b20c4f3de697aa85c309026 # v2.85.11
+      continue-on-error: true
       with:
         tool: nextest@0.9.131
+        # On platforms without prebuilts, nextest can be installed system-wide
+        # or omitted. Building it probably takes longer than running tests
+        # without it
+        fallback: none
 
     - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
       with:
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -150,8 +150,13 @@ jobs:
         rustup target add "$JOB_TARGET"
 
     - uses: taiki-e/install-action@7a79fe8c3a13344501c80d99cae481c1c9085912 # v2.81.10
+      continue-on-error: true
       with:
         tool: nextest@0.9.131
+        # On platforms without prebuilts, nextest can be installed system-wide
+        # or omitted. Building it probably takes longer than running tests
+        # without it
+        fallback: none
 
     - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -150,8 +150,13 @@ jobs:
         rustup target add "$JOB_TARGET"
 
     - uses: taiki-e/install-action@7a79fe8c3a13344501c80d99cae481c1c9085912 # v2.81.10
+      continue-on-error: true
       with:
         tool: nextest@0.9.131
+        # On platforms without prebuilts, nextest can be installed system-wide
+        # or omitted. Building it probably takes longer than running tests
+        # without it
+        fallback: none
```
tgross35: I was thinking that it would be good to keep the binstall fallback (seems to be default https://github.com/taiki-e/install-action#supported-tools) because nextest could add other arches. Looking at it, seems like you can tell binstall to not do the install by setting an env, would you be able to see if that works?

```text
      --disable-strategies <DISABLE_STRATEGIES>
          Disable the strategies specified. If a strategy is specified in `--strategies` and `--disable-strategies`, then it will be removed [env: BINSTALL_DISABLE_STRATEGIES=] [possible values: crate-meta-data, quick-install, compile]
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1253/f9e221cb49441a653220465cd5184e01b5bfee0e..657b55a5b5c1f7c6b7e200a57035312bf61cf2b7)*

tgross35: Though then again, I hope it doesn't build+install binstall if that's not available.

@taiki-e do you have any advice for what we might want here? Basically to install directly if supported, install via binstall if we can download it, but never spend time building either binstall or nextest.

I think showing up as an error is fine to make it easy to see if some other arch stops getting the install for some reason. Would be nice if GH could show warnings instead of red everywhere...

taiki-e: Um, nextest/binstall/quickinstall don't provide prebuilt ppc64/ppc64le/s390x binaries.

I think the easiest way is to add `if: ` to exclude them as before https://github.com/rust-lang/compiler-builtins/pull/1019.

Or add an option something like `fallback: skip` to install-action.

Since nextest uses upload-rust-binary-action with cross/cargo-zigbuild, I think providing prebuilt binaries for s390x gnu and powerpc64le gnu/musl in nextest side should be easy, but I'm not sure about powerpc64 musl.
https://github.com/nextest-rs/nextest/blob/28725d46d6cfe57d505a4e500e3c99c81263848c/.github/workflows/release.yml#L132

tgross35: They don't have installers now, but one reason to avoid an `if` is so we just get it automatically if they start publishing a version. It would be nice if they added them though I don't know that it's worth it, given we're probably pretty much the only consumers.

In that case, I think the current `fallback: none` is fine?

tgross35:
@Gelbpunkt could you rebase this? I don't think we can do any better, so I'm happy to merge.

Gelbpunkt:
> @Gelbpunkt could you rebase this? I don't think we can do any better, so I'm happy to merge.

Done, although I'm still not 100% happy with this, but I guess it's still better than the status quo.

Gelbpunkt:
seems like `0001-compiler-rt-Disable-executable-stack-on-aeabi_u-read.patch` is no longer necessary and fails to apply in CI

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1253/849b5b55182109df3719373c0c7671d3e13a678e..0fd8a42d93244aab40c2fa4371ba93cfe4fcac3b)*

tgross35:
Pushed the Clippy fix here because I couldn't get it to merge separately, the IBM runners are acting flaky and timing out while building nextest... go figure

PR #1254, opened 2026-08-07
symbol-check: Consult rustc about powerpc64 ELF ABI

Gelbpunkt:
The current implementation assumed that powerpc64 glibc targets were all using ELFv1. This really isn't very portable and I don't want to hardcode a list of targets that are ELFv1, so we can instead just ask rustc about the value of target_abi for this specific target to make it work on all targets, even powerpc64-unknown-linux-gnuelfv2.

### DIFF
--- a/crates/symbol-check/tests/all.rs
+++ b/crates/symbol-check/tests/all.rs
@@ -1,7 +1,7 @@
-use std::env;
 use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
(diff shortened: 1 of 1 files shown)


tgross35 (approved):
Makes sense, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1254/f9e221cb49441a653220465cd5184e01b5bfee0e..f8afe612f3fafce479f7943bea0c553abf5dac17)*

PR #1261, opened 2026-08-08
Add integer to `f16` conversions

npmccallum:
## Summary

LLVM emits `__float{,un}sihf`, `__float{,un}dihf`, and
`__float{,un}tihf` when converting 32-, 64-, and 128-bit integers to
`f16`, but compiler-builtins does not currently provide these symbols.
Consequently, `int as f16` can fail to link on targets without another
provider, including musl and Apple targets. glibc targets generally work
only because libgcc supplies the symbols.

This implements the compiler-builtins side of rust-lang/rust#132614,
which is part of the `f16` tracking issue rust-lang/rust#116909.

## Provenance and credit

The conversion algorithms were originally written by Trevor Gross in
rust-lang/compiler-builtins#729.

Trevor's implementation commit, `df3c10a`, was based on the November
2024 tree and also contained test and ordering changes that conflict with
current `main`, so it could not be applied verbatim. This PR preserves
the conversion bodies and ordinary entry points unchanged while adapting
their placement and target configuration to the current tree. The
implementation commit retains Trevor as its primary Git author, with me
listed as co-author.

The test commit incorporates Trevor's original `f16` conversion cases
and system-library gating work, so he is also listed as co-author there.
The overflow-threshold validation follows Juho Kahala's explanation in
rust-lang/compiler-builtins#729, and credits him with a `Suggested-by` trailer.

## Changes

The PR is divided into three commits:

1. **Exercise the integer-to-`f16` link paths**

   Add all six signed and unsigned `i32`/`i64`/`i128` conversions to
   `builtins-test-intrinsics`.

   On targets without a fallback provider, this commit reproduces the
   missing-symbol failure before the implementation is added. The
   intrinsic test crate is excluded from the workspace, so the default
   workspace still builds at this commit.

2. **Implement the six conversion routines**

   Add:

   - `__floatunsihf` and `__floatsihf`
   - `__floatundihf` and `__floatdihf`
   - `__floatuntihf` and `__floattihf`

   The routines reuse the existing left-alignment and round-to-even
   machinery. Because the `f16` range is narrower than every supported
   source integer type, all six paths handle exponent saturation by
   returning the appropriately signed infinity.

3. **Exercise and validate the conversions**

   Register all six routines with the existing `i_to_f` tests and add
   the necessary system-library availability configuration.

## Rounding validation

The existing `i_to_f` check casts the result and its neighboring float
values back to integers. This independently checks rounding and also
exercises the platform `f -> i` casts, which have previously caught ABI
and platform-specific problems.

That check cannot directly cast infinity and its NaN neighbor back to an
integer: those casts saturate or produce zero, yielding a meaningless
error bracket.

For an infinite result, the revised test instead:

1. casts the two adjacent finite values toward zero back to integers;
2. derives the half-ULP overflow threshold from them;
3. verifies that the source integer reached that threshold; and
4. verifies that the infinity has the correct sign.

Equality is accepted as the round-to-even case: the hypothetical
unbounded power of two has an even significand and rounds outward before
the exponent saturates to infinity.

Finite results continue to use the original three-value neighbor
bracket. The native `as`-cast comparison is also retained.

## Platform handling

System-library comparisons are disabled through
`no_sys_f16_int_convert` where the complete routine set is unavailable:

- Apple targets
- Windows targets
- 32-bit x86

On x86_64 UEFI, the `i128`/`u128` entry points use the same split-`u64`
argument ABI now used by the existing `f32` and `f64` conversions.

### DIFF
--- a/builtins-test/tests/conv.rs
+++ b/builtins-test/tests/conv.rs
@@ -18,7 +18,7 @@ mod i_to_f {
                 #[test]
                 fn $fn() {
                     use compiler_builtins::float::conv::$fn;
-                    use compiler_builtins::support::Int;
+                    use compiler_builtins::support::{Int, MinInt};
 
                     fuzz(N, |x: $i_ty| {
                         let f0 = apfloat_fallback!(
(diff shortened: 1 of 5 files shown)


builtins-test/tests/conv.rs
```diff
@@ -35,42 +35,38 @@ mod i_to_f {
...
                                     x,
+                                    expected,
                                     f1.to_bits(),
-                                    y_minus_ulp,
-                                    y,
-                                    y_plus_ulp,
-                                    error_minus,
-                                    error,
-                                    error_plus,
                                 );
```
tgross35: @quaternic would you mind reviewing this portion?

tgross35 (commented):
Thank you for picking this up! I think the test bits are fine but Juho understood the issue better than me.

Totally fine to supersede my PR but mind picking df3c10a99e3c831dcc84326eb267c08cd621ad0c directly? I split off the conflict-y bit.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1261/7a3101d36fed3f6ade112de7e7100520a490b7e5..366b97ec5db2a3bb3990f4ee2a9187f29c468a21)*

npmccallum:
Thanks @tgross35, and thanks for splitting off `df3c10a` — I did try to take it directly. The snag is that it's the Nov-2024 version, so it conflicts against current `main` in `conv.rs`, and its test block uses the old `feature = "no-sys-f16-int-convert"` cfg (now `no_sys_f16_int_convert` via `build.rs`) plus the original bracket oracle — so taking it verbatim would either reintroduce the conflict or leave an intermediate commit that's red on `f16`, which I wanted to avoid for a clean history.

What I've done instead: kept the equivalent implementation but set **you as the author** of the conversions commit (I'm listed as co-author), so the primary credit is yours. It's the same six routines; the only deltas from `df3c10a` are dropping the unrelated reordering churn and adding `#[cfg_attr(target_os = "uefi", unadjusted_on_win64)]` on the `ti→hf` variants for parity with their `sf`/`df` siblings. The rounding oracle follows Juho's diagnosis — apfloat directly instead of the `f→i` bracket.

Happy to adjust the attribution however you'd prefer. Thanks again for the original work here.

builtins-test/tests/conv.rs
```diff
@@ -35,42 +35,38 @@ mod i_to_f {
...
-                                    && ((f0.to_bits() & 1) != 0))
-                            {
+                            type ApFloat = rustc_apfloat::ieee::$apfloat_ty;
+
+                            let expected = if <$i_ty>::SIGNED {
+                                ApFloat::from_i128(x.try_into().unwrap()).value
+                            } else {
+                                ApFloat::from_u128(x.try_into().unwrap()).value
+                            }
+                            .to_bits();
```
tgross35: So this is replacing the `as` cast with tests against apfloat?

We should keep testing against the `as` cast, those come in handy quite a bit for catching ABI or platform bugs that we wouldn't have otherwise. And these functions will get tested against arbitrary-precision math (MPFR) anyway.

The current checks just need to be updated to not produce that error.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1261/7f4c82e1ef14ad14945bf4a2794dfeda80b08226..b8d7db549acb8cc484767baead234753ba53c45b)*

npmccallum: Good call — reverted to the bracket check and kept the native `as`-cast comparison; agreed those catch ABI/platform issues that an apfloat/MPFR comparison would miss (we just hit exactly that class in the f16→i128 and msvc f128 paths).

The only change now is to skip the bracket when the result is infinite: an integer past the exponent range correctly converts to `inf` (only reachable for f16 here), and casting `inf` back to an integer saturates to `iN::MAX`, which is what produced the spurious failure. The infinite case is still covered by the native comparison below. Both the `as`-cast path and the apfloat fallback pass across the matrix.

tgross35: The current version is still unnecessarily skipping more tests that work currently. Can't the check be fixed rather than skipping things? Probably something like what I mentioned at https://github.com/rust-lang/compiler-builtins/pull/729#issuecomment-3889749178.

npmccallum: Thanks—fixed. Finite results keep the existing neighbor check; infinity now uses the two adjacent finite values to verify the overflow threshold and sign. The native `as` comparison remains. This follows Juho’s explanation in #729. Does this address your concern?

tgross35:
I've been working on getting a second set of eyes other than my own on the test changes, should have another look soon.

Could you please clarify what you used AI for? We want to account for what your level of understanding is. The PR description and the commit messages also appear machine-generated which is against our policy https://forge.rust-lang.org/policies/llm-usage.html#-banned, all communication must be handwritten.

I am also very much not thrilled that bb426993ac76a6a9edcac3f5819715950a3a2c8a appears to have amended a commit authored by me with a machine-generated message that I certainly did not write. Because of the continued confusion here, my plan is to pick the two other commits to https://github.com/rust-lang/compiler-builtins/pull/729 rather than the other way around, once approved and messages cleaned up.

PR #1270, opened 2026-08-10
Fix incorrect rounding for subnormals in widening FMA

### LINKED ISSUE #1262: Software FMA for f32 returns wrong result for subnormals on x86_64
libm's software FMA disagrees with hardware FMA result for rounding subnormals.

Run this to reproduce it on x86_64, and with libm depended on with `default-features=false` to get the software FMA rather than the hardware one:

```rust
fn main() {
    let a = std::hint::black_box(f32::from_bits(0x9700_0800));
    let b = std::hint::black_box(f32::from_bits(0x1cff_f001));
    let c = std::hint::black_box(f32::from_bits(0x0001_0002));

    let expected_bits = 0x0001_0001;
    let software_bits = libm::fmaf(a, b, c).to_bits();

    println!("a                 = {:#010x}", a.to_bits());
    println!("b                 = {:#010x}", b.to_bits());
    println!("c                 = {:#010x}", c.to_bits());
    println!("libm::fmaf        = {software_bits:#010x}");
    println!("correctly rounded = {expected_bits:#010x}");

    #[cfg(target_arch = "x86_64")]
    if std::is_x86_feature_detected!("fma") {
        // SAFETY: guarded by runtime FMA feature detection.
        let hardware_bits = unsafe { hardware_fmaf(a, b, c) }.to_bits();
        println!("x86 hardware FMA  = {hardware_bits:#010x}");
        assert_eq!(hardware_bits, expected_bits);
    }

    assert_eq!(software_bits, expected_bits);
}

#[cfg(target_arch = "x86_64")]
#[target_feature(enable = "fma")]
unsafe fn hardware_fmaf(a: f32, b: f32, c: f32) -> f32 {
    use core::arch::x86_64::{_mm_cvtss_f32, _mm_fmadd_ss, _mm_set_ss};

    _mm_cvtss_f32(_mm_fmadd_ss(_mm_set_ss(a), _mm_set_ss(b), _mm_set_ss(c)))
}
```

The output is:

```
a                 = 0x97000800
b                 = 0x1cfff001
c                 = 0x00010002
libm::fmaf        = 0x00010002
correctly rounded = 0x00010001
x86 hardware FMA  = 0x00010001

thread 'main' (365592) panicked at src/main.rs:23:5:
assertion `left == right` failed
  left: 65538
 right: 65537
```

This is happening on x86_64, so it has SSE2 and proper f64 math.

As far as I can tell, the root cause of the issue is this:

https://github.com/rust-lang/compiler-builtins/blob/7a3101d36fed3f6ade112de7e7100520a490b7e5/libm/src/math/generic/fma_wide.rs#L26-L28

This detection of points exactly halfway between representable ones only works for normal values, but not for subnormals.

This 2008 paper describes correct handling of subnormals, and their algorithm comes with a Coq proof: https://guillaume.melquiond.fr/doc/08-tc.pdf
A Rust implementation of this algorithm handles this case correctly for me in SIMD code without FMA (SSE4.2): https://github.com/linebender/fearless_simd/pull/323

Shnatsel:
Adopts the formally proven algorithm from the 2008 paper "Emulation of FMA and correctly-rounded sums:
proved algorithms using rounding to odd" by Sylvie Boldo and Guillaume Melquiond, https://guillaume.melquiond.fr/doc/08-tc.pdf

Fixes https://github.com/rust-lang/compiler-builtins/issues/1262

The common path is unchanged, only the handling of subnormals is altered.

I've added regression tests in-tree, and additionally validated the implementation via random testing on subnormals.

### DIFF
--- a/libm/src/math/generic/fma_wide.rs
+++ b/libm/src/math/generic/fma_wide.rs
@@ -5,7 +5,7 @@
 /* by Sylvie Boldo and Guillaume Melquiond, https://guillaume.melquiond.fr/doc/08-tc.pdf */
 
 use crate::support::{
-    CastFrom, CastInto, Float, FpResult, IntTy, MinInt, NarrowFloat, Round, Status, WideFloat,
+    CastFrom, CastInto, Float, FpResult, IntTy, MinInt, NarrowFloat, Round, WideFloat,
 };
 
 /// Fma implementation when a hardware-backed larger float type is available.
@@ -26,54 +26,17 @@ where
     let mut ui: B::Int = result.to_bits();
     let re = result.ex();
     let zb: B = z.widen();
-    let mut narrowed = result.narrow();
 
-    let prec_diff = B::SIG_BITS - F::SIG_BITS;
-    let excess_prec = ui & ((one << prec_diff) - one);
-    let halfway = one << (prec_diff - 1);
-    let abs_narrowed = narrowed.to_bits() & !F::SIGN_MASK;
-    let at_most_min_normal = abs_narrowed <= F::MIN_POSITIVE_NORMAL.to_bits();
-
-    // Common case: the larger precision is fine if...
-    //  - The result is NOT exactly halfway between two representable points
-    //  - The result does NOT touch the subnormal range
-    if (excess_prec != halfway && !at_most_min_normal)
-        // Or the result is NaN
-        || re == B::EXP_SAT
-        // Or the result is exact
-        || (result - xy == zb && result - zb == xy)
-        // Or the mode is something other than round to nearest
-        || round != Round::Nearest
-    {
-        let min_inexact_exp = (B::EXP_BIAS as i32 + F::EXP_MIN_SUBNORM) as u32;
-        let max_inexact_exp = (B::EXP_BIAS as i32 + F::EXP_MIN) as u32;
-
-        let mut status = Status::OK;
-
-        if (min_inexact_exp..max_inexact_exp).contains(&re) && status.inexact() {
-            // This branch is never hit; requires previous operations to set a status
-            status.set_inexact(false);
-
-            result = xy + z.widen();
-            narrowed = result.narrow();
-            if status.inexact() {
-                status.set_underflow(true);
-            } else {
-                status.set_inexact(true);
-            }
-        }
-
-        return FpResult {
-            val: narrowed,
-            status,
-        };
+    // The round-to-odd method from the paper is only needed for non-exceptional values
+    // and only when rounding to nearest
+    if re == B::EXP_SAT || round != Round::Nearest {
+        return FpResult::ok(result.narrow());
     }
 
-    // TwoSum recovers the exact residual of the widened addition. If the
-    // addition was inexact and its rounded significand is even, move it one
-    // ULP toward the residual to produce a round-to-odd intermediate. Rounding
-    // that intermediate to nearest in `F` gives the correctly rounded result:
-    // https://guillaume.melquiond.fr/doc/08-tc.pdf
+    // TwoSum recovers the exact residual of the widened addition. If the addition was inexact and
+    // its rounded significand is even, move it one ULP toward the residual to produce a round-to-odd
+    // intermediate. Theorem 3 proves that rounding this intermediate to nearest in `F` gives the
+    // correctly rounded result, including for subnormals and underflow.
     let virtual_z = result - xy;
     let residual = (xy - (result - virtual_z)) + (zb - virtual_z);
     let neg = ui >> (B::BITS - 1) != IntTy::<B>::ZERO;
--- a/libm/src/math/generic/fma_wide.rs
+++ b/libm/src/math/generic/fma_wide.rs
@@ -5,7 +5,7 @@
 /* by Sylvie Boldo and Guillaume Melquiond, https://guillaume.melquiond.fr/doc/08-tc.pdf */
 
 use crate::support::{
-    CastFrom, CastInto, Float, FpResult, IntTy, MinInt, NarrowFloat, Round, WideFloat,
+    CastFrom, CastInto, Float, FpResult, IntTy, MinInt, NarrowFloat, Round, Status, WideFloat,
 };
 
 /// Fma implementation when a hardware-backed larger float type is available.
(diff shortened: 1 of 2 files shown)


Shnatsel:
The added unit test involving f128 on 32-bit ARM failed due to ABI bugs: `my_f64 as f128` as well as `.narrow()` produce garbage values by reading from the wrong reigster. I've worked around that by disabling the f128 unit test on ARM, but 32-bit ARM probably shouldn't declare `f128_enabled` at all.

The instruction count on benchmarks also regresses, but that's expected given that we're doing more work for to ensure correctness. The bulk of it is under an `if` that only fires for subnormals, which are [slow even in hardware](https://gitlab.in2p3.fr/CTA-LAPP/COURS/GRAY_SCOTT_REVOLUTIONS/GrayScottRevolution/-/wikis/uploads/4-ComputingPrecision/Subnormal.pdf) so it shouldn't be a big deal.

tgross35:
Thank you for the issue and the fix, this looks good at first pass but I need to do some number crunching.

To confirm: https://github.com/rust-lang/compiler-builtins/issues/1262 mentions x86-64 specifically, but this is present on all arches without hardware fmaf right?

> The added unit test involving f128 on 32-bit ARM failed due to ABI bugs: `my_f64 as f128` as well as `.narrow()` produce garbage values by reading from the wrong reigster. I've worked around that by disabling the f128 unit test on ARM, but 32-bit ARM probably shouldn't declare `f128_enabled` at all.

Happen to have a repro? I'm not aware of any ABI bugs on this platform, everything we know of is tracked at https://github.com/rust-lang/rust/blob/ef20314466010b8b9259ec5f86230c530ca08661/compiler/rustc_codegen_llvm/src/llvm_util.rs#L392-L412.

Shnatsel:
Thank you! I'd appreciate a thorough review. I fit the algorithm from the paper into the existing structure with raising floating-point flags as best I could, but I'm not 100% confident I got it right.

> To confirm: https://github.com/rust-lang/compiler-builtins/issues/1262 mentions x86-64 specifically, but this is present on all arches without hardware fmaf right?

Yes. I've double-checked and 32-bit ARM on qemu fails the f32 rounding test too. I mentioned x86_64 to avoid confusion around non-conformant i586 (x87) floats.

> Happen to have a repro?

Yes, check out commit 8b351a6 in this branch. The logs from CI runs show the failures clearly. It's not just rounding failures, the result is just garbage.

tgross35:
I meant would you be able to reproduce the ABI issue with a smaller snip of code? I wrote a quick test and annotated what I think is going on at https://rust.godbolt.org/z/9PjYETqf6. It's a little weird that Rust and C have different return ABIs, but everything seems otherwise correct there.

Shnatsel:
Oh! Yes, this fails in qemu for me on 32-bit ARM, both targets that failed in CI, while passing on x86_64:

```rust
#![feature(f128)]

fn main() {
    let x = std::hint::black_box(1.0_f64);
    let y = std::hint::black_box(x as f128);
    let bits = y.to_bits();

    assert_eq!(
        bits,
        0x3fff_0000_0000_0000_0000_0000_0000_0000,
        "got {bits:#034x}",
    );
}
```

tgross35:
Whoops, I missed that it's the hf targets that fail, but the codegen still looks fine https://rust.godbolt.org/z/hPd196xbE. Digging...

Shnatsel:
I don't know enough ARM assembly to help here, but here's what my friendly neighbourhood LLM thinks of this:


<details><summary>LLM analysis</summary>
<p>

I ran it with:

```bash
rustc +nightly reproducers/arm-f128-hardfloat-abi.rs \
  --target armv7-unknown-linux-gnueabihf \
  -C linker=arm-linux-gnueabihf-gcc \
  -o /tmp/arm-f128-abi

qemu-arm -L /usr/arm-linux-gnueabihf /tmp/arm-f128-abi
```

It passes on x86_64 but fails on ARM hard-float. The incorrect bits can vary because the callee reads the wrong registers.

To expose the problematic call:

```bash
rustc +nightly reproducers/arm-f128-hardfloat-abi.rs \
  --target armv7-unknown-linux-gnueabihf \
  -C linker=arm-linux-gnueabihf-gcc \
  -O --emit=asm \
  -o /tmp/arm-f128-abi.s

rg -C 8 __extenddftf2 /tmp/arm-f128-abi.s
```

The assembly shows rustc placing the `f64` in `d0`:

```asm
vldr d0, [...]
bl   __extenddftf2
```

But compiler-builtins defines `__extenddftf2` as `extern "aapcs"`, which reads the argument from `r0/r1`.

Verified with:

```text
rustc 1.99.0-nightly (84b36a78a 2026-08-06)
LLVM 23.1.0
```

</p>
</details> 

Please take it with a grain of salt as always.

tgross35:
I think that's pretty accurate, wrote up https://github.com/rust-lang/compiler-builtins/issues/1271 to follow up more

Shnatsel:
I've spent some time minifying the diff to the algorithm because my previous changes messed up the exception flag handling, e.g. on these values:

```rust
a = f32::from_bits(0x5f78_0000);
b = f32::from_bits(0x5f84_2108);
c = f32::from_bits(0x8000_0001);
```

In the first iteration of this PR that raised the overflow flag because of the early narrowing.

I believe the state as of e254558 deviates from the original algorithm as little as possible: it hoists one variable out of a branch (which gets const-folded anyway), adds a check for subnormals and adds one additional check required for correctness.

It took me a while to fully grasp the algorithm and realize this simple fix is possible :sweat_smile:

libm/src/math/generic/fma_wide.rs
```diff
@@ -38,11 +43,10 @@ where
         || round != Round::Nearest
     {
         let min_inexact_exp = (B::EXP_BIAS as i32 + F::EXP_MIN_SUBNORM) as u32;
-        let max_inexact_exp = (B::EXP_BIAS as i32 + F::EXP_MIN) as u32;
 
         let mut status = Status::OK;
```
RalfJung, line 47: Rust doesn't support the float status register and we make no guarantees about which values one sees there. Why are we spending cycles to figure out the status...? Seems like this will slow down everyone just to set some bits that nobody is permitted to look at anyway.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1270/f662be3408ab669689f308423c022fa4cdd5af5b..1a89b833ea695be2039d2ae28a2efe467ff3ab44)*

Shnatsel: If all the flag logic can be deleted, that's great news!

Shnatsel:
Since I opened this PR, I've also reported the issue to upstream musl and there's now a [proposed patch](https://www.openwall.com/lists/musl/2026/08/20/1) significantly rewriting this algorithm.

@dzaima has verified the new algorithm proposed for musl using an SMT solver and found it to be correct: https://www.openwall.com/lists/musl/2026/09/03/2

PR #1281, opened 2026-08-14
Fix `__hexagon_fmadf5` overflow under directed rounding

SuperChamp234:
The overflow path in __hexagon_fmadf5 computed the round-mode predicate (inf vs. largest-finite) into p0, but then overwrote p0 with `dfcmp.eq(ATMP,ATMP)` before consuming it. Since ATMP is +inf, `dfcmp.eq(ATMP,ATMP)` is always true, so CTMP was unconditionally set to infinity and the rounding-mode decision was discarded.

As a result, a finite-magnitude overflow returned +/-inf in every rounding mode.

### DIFF
--- a/compiler-builtins/src/hexagon/dffma.s
+++ b/compiler-builtins/src/hexagon/dffma.s
@@ -338,10 +338,8 @@ __hexagon_fmadf5:
  {
   p0 = !cmp.eq(r2,#1)
   p0 = !cmp.eq(r3,#2)
- }
- {
-  p0 = dfcmp.eq(r9:8,r9:8)
   if (p0.new) r11:10 = r9:8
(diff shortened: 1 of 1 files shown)


tgross35:
@androm3da could you review this?

androm3da:
LGTM: corresponds to https://github.com/llvm/llvm-project/pull/207373

androm3da (approved):

tgross35 (approved):
Thanks both!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1281/89d20d9352892609d726ebb981fb53cadb381d1a..a246e36414ea0cfb0eae3d3f0c3a457fb8b681cf)*

PR #1288, opened 2026-08-17
implement `Complex` mul and div libcalls

folkertdev:


builtins-test/tests/complex.rs
folkertdev, line 1: This is based on 

- https://github.com/llvm/llvm-project/blob/main/compiler-rt/test/builtins/Unit/mulsc3_test.c
- https://github.com/llvm/llvm-project/blob/main/compiler-rt/test/builtins/Unit/divsc3_test.c

and friends. Is this sufficient and sort of right in terms of approach? I can try and clean it up a bit more if so.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1288/77ed6aa63508fde50f8202fc380d70fe82ed3444..6a3bea0156bf79f328a04ed2083cdbccb6d1dd12)*

builtins-test/tests/complex.rs
```diff
@@ -0,0 +1,517 @@
...
+            const NEG_INFINITY: $f = <$f>::NEG_INFINITY;
+            const NAN: $f = <$f>::NAN;
+            const SNAN: $f = <$f>::SNAN;
+
+            #[allow(overflowing_literals)]
+            let (small, big) = if size_of::<$f>() == 2 {
+                (1.0e-2, 1.0e2)
+            } else {
+                (1.0e-6, 1.0e6)
+            };
```
folkertdev, line 179: dealing with `1e6` being too large for `f16`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1288/77ed6aa63508fde50f8202fc380d70fe82ed3444..6a3bea0156bf79f328a04ed2083cdbccb6d1dd12)*

builtins-test/tests/complex.rs
folkertdev: I tried using a fuzzer but that runs into false positives small non-zero numbers (of the shape `0.00000000000000000001`)

compiler-builtins/README.md
```diff
@@ -223,6 +232,12 @@ by Rust.
 - ~~i386/floatundixf.S~~
 - ~~x86_64/floatdixf.c~~
 - ~~x86_64/floatundixf.S~~
+- ~~powixf2.c~~
```
folkertdev, line 235: this one has nothing to do with complex numbers I think? so I've added it to the x87 list here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1288/77ed6aa63508fde50f8202fc380d70fe82ed3444..6a3bea0156bf79f328a04ed2083cdbccb6d1dd12)*

builtins-test/tests/complex.rs
```diff
@@ -0,0 +1,518 @@
+#![cfg_attr(f16_enabled, feature(f16))]
+#![cfg_attr(f128_enabled, feature(f128))]
+#![feature(complex_numbers)]
```
folkertdev, line 3: where is `f128_enabled` actually set? Should `complex_numbers` be guarded similarly?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1288/77ed6aa63508fde50f8202fc380d70fe82ed3444..6a3bea0156bf79f328a04ed2083cdbccb6d1dd12)*

compiler-builtins/src/float/complex/div.rs
folkertdev, line 1: based on https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/builtins/divsc3.c and friends

compiler-builtins/src/float/complex/mul.rs
folkertdev, line 1: based on https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/builtins/divsc3.c and friends

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1288/77ed6aa63508fde50f8202fc380d70fe82ed3444..6a3bea0156bf79f328a04ed2083cdbccb6d1dd12)*

compiler-builtins/src/float/complex/mul.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    Complex::new(F::INFINITY * (a * c - b * d), F::INFINITY * (a * d + b * c))
+}
+
+intrinsics! {
+    #[cfg(f16_enabled)]
+    pub extern "C" fn __mulhc3(a: f16, b: f16, c: f16, d: f16) -> core::num::Complex<f16> {
+        mul(a, b, c, d)
+    }
```
folkertdev: The name `__mulhc3` occurs in the LLVM source (and some other places online) but I don't think it is actually generated or called? So, maybe we can/should drop the implementations for `f16`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1288/77ed6aa63508fde50f8202fc380d70fe82ed3444..6a3bea0156bf79f328a04ed2083cdbccb6d1dd12)*

beetrees: Clang can definitely generate `__mulhc3` ([example](https://godbolt.org/z/fMWYs4cGG)). LLVM itself doesn't generate complex number libcalls at all AFAIK (it's entirely handled by Clang), so the only thing that matters here is what we call from `core`.

builtins-test/tests/complex.rs
beetrees: It's set in [libm/configure.rs](https://github.com/rust-lang/compiler-builtins/blob/9c424c6dc0e8da0e02f04158f4f14dd1c358f1b4/libm/configure.rs#L145), and is currently equivalent to `cfg(target_has_reliable_f128)`. `complex_numbers` shouldn't need it unless it has its own platform-specific issues.

tgross35: For c-b it's just `reliable_f128` https://github.com/rust-lang/compiler-builtins/blob/9c424c6dc0e8da0e02f04158f4f14dd1c358f1b4/libm/configure.rs#L129-L165. No need for something similar here

compiler-builtins/src/float/complex/mul.rs
folkertdev: Well it seems most consistent to use this approach for all 4 float types. `f16` is a bit annoying to test due to its smaller range but we'll manage.

compiler-builtins/src/float/complex/div.rs
beetrees: Just in case you were interested, LLVM's algorithms are basically 1:1 transliterations of the example algorithms given in the [C standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf#subsection.14.5.2).

builtins-test/tests/complex.rs
folkertdev: Then is CI using some older nightly version? the hash from `rust-version` should have `core::num::Complex`, and it works locally for me.

folkertdev: Ah, it is `BENCHMARK_RUSTC: nightly-2026-08-06`

why is that out of sync?

tgross35: Out of sync with what? It's pinned so tests show changes in libm, not the compiler

builtins-test/tests/complex.rs
tgross35: It would be best to test against MPC with our usual test generators, but I can take care of this since the test code is somewhat complex. Had something in the works at https://github.com/rust-lang/compiler-builtins/commit/38fab001cd37379b48fb99c894c37128572d1170 but never got it over the line.

compiler-builtins/src/float/complex/mul.rs
tgross35: Yeah please include `__mulhc3`. Like beetrees said I don't think LLVM has any complex intrinsics, it's basically library plus frontend.

In general I'd prefer to always cover all four float types when possible, there's always some exceptions to deal with.

folkertdev: Sounds good. We do run into an issue on i586 because currently the ABI code uses a vector register for passing such values (which requires SSE2). I suspect just bitpacking into an `f32` will also work and circumvent this problem

builtins-test/tests/complex.rs
tgross35: > I tried using a fuzzer but that runs into false positives small non-zero numbers (of the shape `0.00000000000000000001`)

Also these probably aren't false positives, this algorithm is imprecise. I think @quaternic had an idea for how we can better define the bounds.

compiler-builtins/src/float/complex/mul.rs
tgross35: It's fine to just disable the test on `cfg(x86_no_sse2)` until that's fixed. Looks like they are just passed and returned indirectly on i586? https://c.godbolt.org/z/bhaeYMvjc

Tbh we should treat it as an indirect struct of i32/i64 at function boundaries rather than float/double to avoid NaN quieting...

beetrees: A thought: if we're just calling `__{mul,div}[hsdt]c3` from `core`, is there any reason not to just implement them in `core` directly and skip the indirection to `compiler-builtins`? That way we wouldn't have to worry about matching the builtin ABI that Clang/GCC use.

folkertdev: I had the same thought, with the slow path in a `#[cold]` function. The only downside I think is that on platforms where libc does provide the function you'd have a slight binary size increase. 

I'm open to it, it does seem simpler.

beetrees: Rust `std`/`core` already contain a lot of `libc` functionality (e.g. stdio, float parsing/formatting), plus LTO will be able to remove it if it's unused, so I reckon putting it in `core` is the best option. Otherwise we'll have to have different export names for soft-float etc., which will be a lot of implementation complexity both in `compiler-builtins` and in `core`.

folkertdev: right, and actually for like Miri that is much easier too. I'll play around with it a bit tomorrow, I also need to look into the specialization (because integers etc. should just use the standard formula)

tgross35: My preference is usually to have float routines here because they're easier to test, but the Miri/CTFE argument makes sense. It could be fine to have the canonical definitions live here so they get tested, and sync them to `core` as needed.

That said, figuring out what to do about `ilogb` and `scalbn` is nonobvious so I think landing them here is still the best way to start. If there are ABI concerns then I'd just name it `__rust_mulhc3` and similar like we do for other intrinsics that we own.

> Otherwise we'll have to have different export names for soft-float etc., which will be a lot of implementation complexity both in `compiler-builtins` and in `core`.

Only the issues we're already hitting with f16 on i586 and f128 on ppc, no?

folkertdev: I'm hitting `scalbn` not being available in core (for division), should that just be added in general, or is there a reason it's missing?

Overall I need the `Float` trait in core to be a bit more powerful, so far it looks reasonable though.

beetrees: > My preference is usually to have float routines here because they're easier to test, but the Miri/CTFE argument makes sense. It could be fine to have the canonical definitions live here so they get tested, and sync them to core as needed.

If we want to have the definitions in `compiler-builtins` for access to its testing infrastructure, we could have them in a file here and then `#[path = "../../compiler-builtins/path/to/file.rs"] mod complex_math;` somewhere in `core` (similar to how `stdarch` is included in `core`). That would allow the definitions to be directly tested in `compiler-builtins` while also allowing Miri, CTFE and MIR/LLVM optimisations to work. Alternatively, the definitions could be in `core` but still tested in `compiler-builtins` - that would just require a subtree sync/nightly update for any improvements to reach `compiler-builtins` from `core`, and would obviously mean that PR CI wouldn't run the `compiler-builtins` extra tests (although the regular complex tests would still be run by `rust-lang/rust` CI of course).

folkertdev: We already appear to be using the `#[path = "..."]` approach with `core_simd` and `core_arch`, at least for `stdarch` i don't remember that ever having caused trouble really. We use `crate::` instead of `core::`  for it to work nicely with bootstrap I think but otherwise it's fine.

folkertdev: Though when I try this naively it doesn't just work, and probably needs some `cfg` magic to enable/disable the right things depending on the context. I'm not sure how invasive that'll be.

```
error[E0433]: cannot find `math` in the crate root
 --> /home/folkertdev/rust/compiler-builtins/compiler-builtins/src/float/complex/div.rs:3:12
  |
3 | use crate::math::libm_math::generic::{fmax, ilogb, scalbn};
  |            ^^^^ unresolved import
  |
help: a similar path exists
  |
3 | use crate::f32::math::libm_math::generic::{fmax, ilogb, scalbn};
  |            +++++

error[E0432]: unresolved import `crate::support`
 --> /home/folkertdev/rust/compiler-builtins/compiler-builtins/src/float/complex/div.rs:4:12
  |
4 | use crate::support::{CastInto, Float};
  |            ^^^^^^^ could not find `support` in the crate root

error[E0433]: cannot find module or crate `libm` in this scope
 --> library/core/src/num/complex.rs:3:5
  |
3 | use libm::math::generic;
  |     ^^^^ use of unresolved module or unlinked crate `libm`
  |
  = help: if you wanted to use a crate named `libm`, use `cargo add libm` to add it to your `Cargo.toml`

error[E0432]: unresolved import `crate::support`
 --> /home/folkertdev/rust/compiler-builtins/compiler-builtins/src/float/complex/mul.rs:3:12
  |
3 | use crate::support::Float;
  |            ^^^^^^^ could not find `support` in the crate root

error: cannot find macro `intrinsics` in this scope
  --> /home/folkertdev/rust/compiler-builtins/compiler-builtins/src/float/complex/div.rs:56:1
   |
56 | intrinsics! {
   | ^^^^^^^^^^

error: cannot find macro `intrinsics` in this scope
  --> /home/folkertdev/rust/compiler-builtins/compiler-builtins/src/float/complex/mul.rs:62:1
   |
62 | intrinsics! {
   | ^^^^^^^^^^
```

tgross35: > I'm hitting `scalbn` not being available in core (for division), should that just be added in general, or is there a reason it's missing?

We don't have any API that needs it currently, though IIRC there was discussion about making this `<<`. Feel free to propose that if you're interested, it would need to be hooked anyway since `scalbn` isn't const (APFloat does have it).

Considering it some more, I think my proposal would be to merge these basically as-is except with a `__rust_*` name, and then duplicate the algorithm in APFloat for CTFE/Miri - should be reasonably easy, and you could test against the runtime impl. That's still algorithm duplication which I don't love, but it means the rust-lang/rust can use the simplest implementation while we can poke around for optimizations in this repo. Also we don't need to deal with figuring out imports (like above) or how hitting `const` restrictions (assuming that's something to eventually care about).

Also I suspect @RalfJung would want to add some variance to the outputs in Miri anyway, considering we can't guarantee correctly rounded output here.

> Overall I need the `Float` trait in core to be a bit more powerful, so far it looks reasonable though.

For future reference, the traits at https://github.com/rust-lang/rust/blob/e71c0f1e3395b10a8c331317be1a5c107bdf7b2e/library/core/src/num/imp/traits.rs are based on the traits here and it should be fine to bring more methods over as needed.

RalfJung: >  it would need to be hooked anyway since scalbn isn't const (APFloat does have it).

Does or does not?
The next paragraph sounds like you mean "does not", but scalbn [exists](https://docs.rs/rustc_apfloat/latest/rustc_apfloat/trait.Float.html#method.scalbn) in apfloat and Miri uses it.
Miri also does not apply any rounding error since I would assume this operation is precise. It does not even touch the mantissa after all.

tgross35: No that is what I meant, I was saying that APFloat having `scalbn` means we could some form of it const. The rest is about the complex multiplication and division here, the only relation to `scalbn` is that it's used as part of division.

What I meant is that these multiplication and division algorithms don't actually provide correctly rounded results for all inputs. I don't think we can easily do better (these are the same algorithms used by the C libraries I've seen) but I also don't think we'd want to guarantee that the result is imprecise forever, because something better could come along.

So my thought was that it may be beneficial for Miri to inject randomness to complex multiplication and division, so the result lies between the results of these functions and the correctly rounded result. Of course, first we would need to figure out how to do that...

RalfJung: Oh we won't just implement complex mul/div as normal Rust functions in the standard library?

tgross35: That's part of what this thread is about :) there are tradeoffs but I think we should replicate Clang/GCC, at least for the time being.

folkertdev:
I've updated the names, I'm not exactly sure why though? At least currently, the symbols do implement the actual C signature and behavior.

builtins-test/tests/complex.rs
tgross35: I've since bumped the version

compiler-builtins/src/float/complex/div.rs
```diff
@@ -0,0 +1,76 @@
+use core::num::Complex;
+
+use crate::math::libm_math::generic::{fmax, ilogb, scalbn};
+use crate::support::{CastInto, Float};
+
+/// Returns the quotient of `(a + ib)` and `(c + id)`.
+///
+/// This implementation uses the standard formula, but has special behavior when the output
+/// of that formula has both a real and imaginary component that are NaN.
+fn div<F: Float>(mut a: F, mut b: F, mut c: F, mut d: F) -> Complex<F>
```
tgross35: Can you name this `cdiv` or `complex_div`? Bit easier to distinguish from the real version. (Ditto for mul)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1288/6474adc72054bd73fda446b5b628876c9ac4188d..819f9df08ca191ae2bef87449bbdd455800438c8)*

compiler-builtins/src/float/complex/div.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    pub extern "C" fn __rust_divdc3(a: f64, b: f64, c: f64, d: f64) -> core::num::Complex<f64> {
+        div(a, b, c, d)
+    }
+
+    // On PowerPC `__divtc3` is for double-double `long double` rather than IEEE `f128`.
+    #[ppc_name = __rust_divkc3]
+    #[cfg(f128_enabled)]
+    pub extern "C" fn __rust_divtc3(a: f128, b: f128, c: f128, d: f128) -> core::num::Complex<f128> {
```
tgross35: I think we might as well skip the alias if we're controlling the names. Also the comment doesn't hurt but also isn't really necessary, the t/k difference is documented by `ppc_name` in the macro.

Same for mul

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1288/6474adc72054bd73fda446b5b628876c9ac4188d..819f9df08ca191ae2bef87449bbdd455800438c8)*

compiler-builtins/src/float/complex/div.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    z
+}
+
+intrinsics! {
+    #[cfg(all(f16_enabled, not(x86_no_sse2)))]
+    pub extern "C" fn __rust_divhc3(a: f16, b: f16, c: f16, d: f16) -> core::num::Complex<f16> {
+        div(a, b, c, d)
+    }
```
tgross35, line 60: Is there actually a crash when trying to build on i586? If not then it's fine to just remove the attribute and only gate the tests. If so, could use a comment.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1288/6474adc72054bd73fda446b5b628876c9ac4188d..819f9df08ca191ae2bef87449bbdd455800438c8)*

compiler-builtins/src/float/complex/div.rs
```diff
@@ -0,0 +1,76 @@
...
+where
+    u32: CastInto<F::Int>,
+{
+    let max = fmax(c.abs(), d.abs());
+    let mut ilogbw = 0;
+    if max.is_finite() && max != F::ZERO {
+        ilogbw = ilogb(max);
+        c = scalbn(c, -ilogbw);
+        d = scalbn(d, -ilogbw);
+    }
```
tgross35: The original casts the result of `logb` to a `double` then back to `int`. That's just useless right? All possible values should roundtrip fine for all types.

builtins-test/tests/complex.rs
```diff
@@ -0,0 +1,518 @@
...
+                    if classify(actual) != Class::Zero {
+                        return true;
+                    }
+                }
+                Class::Infinite | Class::NaN | Class::NonZeroAndNaN => {
+                    if classify(actual) != Class::NaN {
+                        return true;
+                    }
+                }
+            },
```
tgross35: Nit: flattening makes this easier to read than the C-style patterns

```rust
        let expected = match (classify(p), classify(q)) {
            (Class::Zero, Class::Zero | Class::NonZero) => Class::Zero,
            (Class::Zero, Class::Infinite | Class::NaN | Class::NonZeroAndNaN) => Class::NaN,
            // ...
        };
        classify(actual) == expected
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1288/6474adc72054bd73fda446b5b628876c9ac4188d..819f9df08ca191ae2bef87449bbdd455800438c8)*

compiler-builtins/src/float/complex/div.rs
folkertdev: it errors

```
  error: this function definition uses SIMD vector type `Complex<f16>` which (with the chosen ABI) requires the `sse` target feature, which is not enabled
     --> builtins-shim/../compiler-builtins/src/macros.rs:424:9
      |
  424 |           pub $(unsafe $($empty)?)? extern $abi fn $name( $($argname: $ty),* ) $(-> $ret)? {
      |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function defined here
```

Because we use a vector type to get the correct ABI.

RalfJung: That error was written assuming that it'd be obvious that there's a vector type. When there are vectors hiding in `core::num::Complex<f16> ` we may have to make this error message more clear...

beetrees: cc https://github.com/rust-lang/rust/issues/131819

tgross35 (approved):
LGTM, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1288/b2a59c9550afc1fabdc744b15a220c67ec166a9c..dfac6d7a30a4504ca193c140529ff7c2b179b17d)*

tgross35:
> I've updated the names, I'm not exactly sure why though? At least currently, the symbols do implement the actual C signature and behavior.

https://github.com/rust-lang/compiler-builtins/pull/1288#discussion_r3798674339 mentioned an incompatibility on i586, which I took to mean we can't call the C versions of these functions on this platform. Is that not accurate?

Once that and any other ABI mismatches are resolved, or if it's not a problem at all, it owuld be good to use the C names and save some code size. Or we could skip tests on platforms where there are ABI bugs but I figure that's more of a headache.

folkertdev:
> which I took to mean we can't call the C versions of these functions on this platform. Is that not accurate?

Only the `f16` variants, the rest works fine. It's the well-known "f16 is passed in a register that does not exist in the baseline" issue. 

I'm not totally sure how to handle that. Normally a vector register is used to pass it as `<2 x half>`, I guess we can treat it as a normal `{half, half}` struct when SSE is not enabled?

beetrees:
PowerPC also has a similar issue with `Complex<f128>`, which is passed in vector registers.

RalfJung:
> I'm not totally sure how to handle that.

We can't do anything about this in Rust I think, it needs to be fixed in LLVM first -- when lowing an operation to a libcall, LLVM needs to consider the target features of the current function and pick a libcall that can actually be performed using the available target features.

beetrees:
For complex numbers specifically, the libcalls are inserted by the frontend (or called directly from the library): LLVM itself has no concept of complex numbers.

RalfJung:
Ah. Yeah then it is up to us to insert calls that can actually work on the current ABI.

PR #1292, opened 2026-08-18
Remove `#[aapcs_on_arm]`

### LINKED ISSUE #1271: Libcall calling convention mismatch on 32-bit arm hardfloat
Discovered by @Shnatsel at https://github.com/rust-lang/compiler-builtins/pull/1270.

On arm32 we use `aapcs_on_arm` for all float intrinsics so f16/f32/f64 get passed in GPRs. However, it looks like on hardfloat targets LLVM is expecting them to be passed in float registers - at least for `__extenddftf2`. This hasn't been noticed before because the libcalls aren't typically needed on armhf, but they are with f128.

I'm actually not sure why we need to use `extern "aapcs"` at all: If aapcs is the default on softfloat targets and hardfloat targets want floats in fpregs, could we just delete `aapcs_on_arm` and use `extern "C"`?

Extremely long example showing all our libcalls https://rust.godbolt.org/z/rqx8v88bq

beetrees:
Fixes rust-lang/compiler-builtins#1271 by removing `#[aapcs_on_arm]`: `compiler-rt` only does the equivalent on ARM soft-float targets where the `"C"` ABI is already AAPCS.

### DIFF
--- a/compiler-builtins/src/macros.rs
+++ b/compiler-builtins/src/macros.rs
@@ -38,12 +38,9 @@
 ///
 /// A quick overview of attributes supported right now are:
 ///
+// FIXME: Add missing attributes.
 /// * `maybe_use_optimized_c_shim` - indicates that the Rust implementation is
 ///   ignored if an optimized C version was compiled.
-/// * `aapcs_on_arm` - forces the ABI of the function to be `"aapcs"` on ARM and
-///   the specified ABI everywhere else.
-/// * `unadjusted_on_win64` - like `aapcs_on_arm` this switches to the
-///   `"unadjusted"` abi on Win64 and the specified abi elsewhere.
 /// * `arm_aeabi_alias` - handles the "aliasing" of various intrinsics on ARM
 ///   their otherwise typical names to other prefixed ones.
 /// * `ppc_name` - changes the name of the symbol on PowerPC platforms without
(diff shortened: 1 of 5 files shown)


compiler-builtins/src/macros.rs
```diff
@@ -40,10 +40,8 @@
 ///
 /// * `maybe_use_optimized_c_shim` - indicates that the Rust implementation is
 ///   ignored if an optimized C version was compiled.
-/// * `aapcs_on_arm` - forces the ABI of the function to be `"aapcs"` on ARM and
-///   the specified ABI everywhere else.
-/// * `unadjusted_on_win64` - like `aapcs_on_arm` this switches to the
-///   `"unadjusted"` abi on Win64 and the specified abi elsewhere.
+/// * `unadjusted_on_win64` - this switches to the `"unadjusted"` abi on Win64
+///   and the specified abi elsewhere.
```
bjorn3: I forgot to remove this comment when removing `unadjusted_on_win64`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1292/d2a99bc4449b85388657e9a5d88cf2f496240e78..1c8227c64ff4ee08bae584a8eff838d7c1fb07ea)*

beetrees: I've removed it. I noticed several existing attributes aren't mentioned: I've left a `FIXME` for that.

tgross35 (approved):
LGTM but needs a rebase

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1292/d2a99bc4449b85388657e9a5d88cf2f496240e78..9deb8452f7c87e6572e4fd7526d9a4d33cee2e82)*

tgross35:
(Pushed to add the description to the commit message since this repo uses rebase)

PR #1293, opened 2026-08-18
Rename `#[ppc_alias]` to `#[ppc_name]`

beetrees:
`#[ppc_alias]` suggests this is just adding an additional export name (which would be incorrect), whereas it is actually renaming the function on PPC (which is correct).


tgross35 (approved):
I've gotten confused by this too, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1293/d2a99bc4449b85388657e9a5d88cf2f496240e78..52f8b4b94d4a8126371bdb7ab3e4c519b850a0b9)*

PR #1296, opened 2026-08-20
libm: add unit tests for the relaxed cbrtf64 and hypotf64

rootkiller6788:
The `relaxed` module (renamed from `approx` in rust-lang/compiler-builtins#1277) exposes `cbrtf64` and `hypotf64` as public API, but nothing exercises them: the module is not part of the MPFR/musl comparison suite. This adds in-crate unit tests so regressions in these implementations are caught by the normal `cargo test -p libm` run that CI executes on every target.

Coverage added:
- **Special values**: signed zeros, ±inf, NaN propagation, subnormal handling, and max-finite (no spurious overflow/underflow to inf/zero).
- **Exact cases**: perfect cubes (`8 -> 2`, `27 -> 3`, `-8 -> -2`, `0.125 -> 0.5`) and Pythagorean triples (`(3,4) -> 5`, `(5,12) -> 13`).
- **Full-domain sweeps**: compare against the extensively-tested main `cbrt`/`hypot` across every exponent (including the `hypot` scaling boundaries around 2^510 and 2^-450) and assert the relaxed results stay within a few ULPs.

Measured worst-case deviation from the main implementations on x86_64 is 0 ULPs for `cbrtf64` and 1 ULP for `hypotf64`; the sweep tolerance of 8 ULPs gives headroom for platforms with x87 double rounding while still catching real regressions.

All libm unit tests pass (116 total) in debug, release, and `--no-default-features` configurations, with `-Dwarnings`.

### DIFF
--- a/libm/src/math/relaxed/cbrtf64.rs
+++ b/libm/src/math/relaxed/cbrtf64.rs
@@ -111,3 +111,95 @@ pub fn cbrtf64(x: f64) -> f64 {
     t = t + t * r; /* error <= 0.5 + 0.5/3 + epsilon */
     t
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// The distance between `a` and `b` measured in ULPs, i.e. the number of
+    /// distinct representable `f64` values in between (inclusive of `±0`).
+    fn ulps(a: f64, b: f64) -> u64 {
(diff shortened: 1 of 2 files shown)


oli-obk:
The account has been banned for spamming and will not be responding further to this PR

PR #1300, opened 2026-08-25
support hardware sqrt for f32 on arm32 targets

silverstillisntgold:
Following up on [this](https://github.com/rust-lang/rust/issues/137578#issuecomment-5405505524) discussion.

Currently this uses a bit of a janky trick to work around a rustc bug. Since we can't use sreg to operate on input `x` in-place, we instead cast it to an integer, load it into an integer register, then move it to a floating point register where we can use sqrt. A little annoying, but 3 instructions is still far better than a whole software sqrt routine lol. Hopefully the [fix](https://github.com/rust-lang/rust/pull/159988) for the bug will land in the next couple of weeks and will be in stable 1.100.0. The original bug report can be found [here](https://github.com/rust-lang/rust/issues/159976).

I tried to mimic the structure of previous implementations, but let me know if anything should be changed. Was also going to do the f64 `sqrt` variant for chips like the M7, M55, M85, etc., but I'm not sure there's currently a reliable (and stable) way to detect both HF ABI and fp64 support.

### DIFF
--- a/libm/configure.rs
+++ b/libm/configure.rs
@@ -19,6 +19,7 @@ pub struct Config {
     pub target_triple: String,
     pub target_triple_split: Vec<String>,
     pub target_arch: String,
+    pub target_abi: Option<String>,
     pub target_env: String,
     pub target_families: Vec<String>,
     pub target_os: String,
@@ -63,6 +64,8 @@ impl Config {
             opt_level: env::var("OPT_LEVEL").unwrap(),
             cargo_features,
             target_arch: env::var("CARGO_CFG_TARGET_ARCH").unwrap(),
+            // FIXME(msrv): defined since 1.78.
+            target_abi: env::var("CARGO_CFG_TARGET_ABI").ok(),
             target_env: env::var("CARGO_CFG_TARGET_ENV").unwrap(),
             target_families,
             target_os: env::var("CARGO_CFG_TARGET_OS").unwrap(),
@@ -114,6 +117,9 @@ pub fn emit(cfg: &Config) {
     // We have to cfg our code accordingly.
     let thumb_1 = split[0] == "thumbv6m" || split[0] == "thumbv8m.base";
 
+    // FIXME(msrv): `cfg(target_abi)` is usable since 1.78.
+    let target_abi_eabihf = cfg.target_abi.as_deref() == Some("eabihf");
+
     // Shorthand to detect i586 targets
     let x86_no_sse2 = cfg.target_arch == "x86" && !cfg.target_features.iter().any(|f| f == "sse2");
 
@@ -123,7 +129,10 @@ pub fn emit(cfg: &Config) {
     // Arch shorthand config is used in most crates.
     set_cfg("thumb", thumb);
     set_cfg("thumb_1", thumb_1);
+    set_cfg("target_abi_eabihf", target_abi_eabihf);
     set_cfg("x86_no_sse2", x86_no_sse2);
+    // FIXME(arm_target_feature): the exact target features are not yet stable.
+    println!("cargo:rustc-check-cfg=cfg(target_feature, values(\"vfp2sp\", \"vfp2\"))");
 
     match cfg.library {
         Library::CompilerBuiltins => {
--- a/libm/src/math/arch/arm/mod.rs
+++ b/libm/src/math/arch/arm/mod.rs
@@ -1,0 +1,8 @@
+//! Architecture-specific support for Arm32/Thumb with hard float ABI.
+
+mod sqrt;
+
+#[cfg(target_feature = "vfp2")]
+pub use sqrt::sqrt;
+#[cfg(target_feature = "vfp2sp")]
+pub use sqrt::sqrtf;
--- a/libm/src/math/arch/arm/sqrt.rs
+++ b/libm/src/math/arch/arm/sqrt.rs
@@ -1,0 +1,29 @@
+#[cfg(target_feature = "vfp2sp")]
+pub fn sqrtf(mut x: f32) -> f32 {
+    // SAFETY: vsqrt.f32 is available with `vfp2sp`.
+    unsafe {
+        core::arch::asm!(
+            "vsqrt.f32 {x}, {x}",
+            x = inout(sreg) x,
+            options(nostack, nomem, pure),
+        );
+    }
+    x
+}
+
+#[cfg(target_feature = "vfp2")]
+pub fn sqrt(mut x: f64) -> f64 {
+    // SAFETY: vsqrt.f64 is available with `vfp2`.
+    // Can't just use plain old `dreg` because `vfp2` only provides a base
+    // level of FPU support with a non-specified amount of fp64 registers.
+    // It's fine to use `dreg_low16` here, but use `dreg_low8` instead to
+    // keep the requirements as minimal as possible.
+    unsafe {
+        core::arch::asm!(
+            "vsqrt.f64 {x}, {x}",
+            x = inout(dreg_low8) x,
+            options(nostack, nomem, pure),
+        );
+    }
+    x
+}
--- a/libm/src/math/arch/mod.rs
+++ b/libm/src/math/arch/mod.rs
@@ -41,6 +41,14 @@ cfg_select_nofmt! {
             sqrtf16,
         };
     }
+    all(target_arch = "arm", target_abi_eabihf) => {
+        mod arm;
+
+        #[cfg(target_feature = "vfp2sp")]
+        pub use arm::sqrtf;
+        #[cfg(target_feature = "vfp2")]
+        pub use arm::sqrt;
+    }
     any(target_arch = "loongarch32", target_arch = "loongarch64") => {
         mod loongarch;
 
--- a/libm/src/math/sqrt.rs
+++ b/libm/src/math/sqrt.rs
@@ -21,6 +21,7 @@ pub fn sqrtf(x: f32) -> f32 {
         name: sqrtf,
         use_arch: any(
             all(target_arch = "aarch64", target_feature = "neon"),
+            all(target_arch = "arm", target_abi_eabihf, target_feature = "vfp2sp"),
             all(target_arch = "wasm32", intrinsics_enabled),
             all(any(target_arch = "loongarch32", target_arch = "loongarch64"), target_feature = "f"),
             target_feature = "sse2"
@@ -38,6 +39,7 @@ pub fn sqrt(x: f64) -> f64 {
         name: sqrt,
         use_arch: any(
             all(target_arch = "aarch64", target_feature = "neon"),
+            all(target_arch = "arm", target_abi_eabihf, target_feature = "vfp2"),
             all(target_arch = "wasm32", intrinsics_enabled),
             all(any(target_arch = "loongarch32", target_arch = "loongarch64"), target_feature = "d"),
             target_feature = "sse2"
--- a/etc/function-definitions.json
+++ b/etc/function-definitions.json
@@ -985,6 +985,7 @@
     "sqrt": {
         "sources": [
             "libm/src/math/arch/aarch64/sqrt.rs",
+            "libm/src/math/arch/arm/sqrt.rs",
             "libm/src/math/arch/loongarch/sqrt.rs",
             "libm/src/math/arch/wasm32/sqrt.rs",
             "libm/src/math/arch/x86/sqrt.rs",
@@ -996,6 +997,7 @@
     "sqrtf": {
         "sources": [
             "libm/src/math/arch/aarch64/sqrt.rs",
+            "libm/src/math/arch/arm/sqrt.rs",
             "libm/src/math/arch/loongarch/sqrt.rs",
             "libm/src/math/arch/wasm32/sqrt.rs",
             "libm/src/math/arch/x86/sqrt.rs",


silverstillisntgold:
seems like I murdered CI

silverstillisntgold:
Seems like the only failure is on the libm MSRV check due to the use of `target_abi` for ensuring hardware FP support. Looks like this [feature](https://github.com/rust-lang/rust/issues/80970) was stabilized in 1.78.0 (released on May 2, 2024). Unfortunate. Guess that puts this PR on the backburner until a 0.3 release alongside an MSRV bump?

tgross35:
The ABI thing can be worked around with the build script: update https://github.com/rust-lang/compiler-builtins/blob/b2a59c9550afc1fabdc744b15a220c67ec166a9c/libm/configure.rs#L99-L126 with a `target_abi_eabihf` and `target_abi_eabi`, after picking up `CARGO_CFG_TARGET_ABI`.

It looks like you could also check for the `fpregs` without `soft-float`? I'm not sure whether that's more accurate. @taiki-e you may have some thoughts here.

```text
$ diff -up <(rustc --print cfg --target armv7-unknown-linux-gnueabi) <(rustc --print cfg --target armv7-unknown-linux-gnueabihf)
--- /dev/fd/63  2026-08-25 21:04:43
+++ /dev/fd/62  2026-08-25 21:04:43
@@ -3,7 +3,7 @@ relocation_model="pic"
 overflow_checks
 panic="unwind"
 relocation_model="pic"
-target_abi="eabi"
+target_abi="eabihf"
 target_arch="arm"
 target_endian="little"
 target_env="gnu"
@@ -12,7 +12,6 @@ target_feature="fpregs"
 target_feature="d32"
 target_feature="dsp"
 target_feature="fpregs"
-target_feature="soft-float"
 target_feature="thumb2"
 target_feature="v5te"
 target_feature="v6"
```

Is `sqrt` available in the `fpregs` baseline?

libm/src/math/arch/arm/sqrt.rs
```diff
@@ -0,0 +1,21 @@
+use core::arch::asm;
+
+#[inline]
+pub fn sqrtf(x: f32) -> f32 {
+    // Needed to work around a bug in rustc.
+    let mut bits = x.to_bits();
+    // TODO: When Rust 1.100.0 is released, this can instead use sreg
+    // to avoid moving in and out of the s0 register.
```
tgross35: Use `FIXME` rather than `TODO` for anything intended to be merged (that's a conventions for most rust-lang repos).

Also please add a tiny bit more info about what's actually going on to the comment, e.g. "FIXME(rust-lang/rust#1234) rust incorrectly forbids vreg access with fpregs but not vfp2. Load via GPRs to avoid this."

libm/src/math/arch/arm/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! Architecture-specific support for arm32 with hard float ABI.
```
tgross35: Call this module `arm32hf` to make that obvious, in case we eventually need a soft float version.

silverstillisntgold:
Hadn't even considered utilizing the build script, that might also make it easy to implement f64 `sqrt`. I was initially avoiding using target-features since that approach falls apart when building using the thumb targets for embedded.

```
rustc --print cfg --target thumbv8m.main-none-eabihf
debug_assertions
panic="abort"
target_abi="eabihf"
target_arch="arm"
target_endian="little"
target_env=""
target_has_atomic="16"
target_has_atomic="32"
target_has_atomic="8"
target_has_atomic="ptr"
target_has_atomic_primitive_alignment="16"
target_has_atomic_primitive_alignment="32"
target_has_atomic_primitive_alignment="8"
target_has_atomic_primitive_alignment="ptr"
target_os="none"
target_pointer_width="32"
target_vendor="unknown"
```

The output is identical for thumbv7em-none-eabihf and both non-hf variants.

tgross35:
Is`vsqrt.f64` available with the `vfp2` target feature? Seems to be the case based on https://rust.godbolt.org/z/5hdce3EGK. If so, it should be fine to add a f64 version gated on that, similar to what is done with `sqrtf16` on aarch64.

(Please confirm what `vfp2` actually enables)

libm/src/math/arch/mod.rs
```diff
@@ -41,6 +41,11 @@ cfg_select_nofmt! {
             sqrtf16,
         };
     }
+    all(target_arch = "arm", target_abi = "eabihf") => {
```
tgross35: Change to `target_abi_eabihf` mentioned above

tgross35: Does `target_abi = "eabihf"` imply `fregs` and is `fregs` always sufficient for f32 sqrt? I think yes for both of those, but would like confirmation.

tgross35 (changes requested):
This plus the change to use the build script

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1300/b2a59c9550afc1fabdc744b15a220c67ec166a9c..020bec7b7b424dd20b8f7471dbdf88840631a8f9)*

tgross35:
One other note is that you may want to consider doing something similar for `fma` once this lands, since that's used by many other operations.

libm/src/math/arch/mod.rs
beetrees: AFAICT, `vsqrt.f32` requires the `vfp2sp` LLVM target feature (implied by the `vfp2` Rust target feature).

tgross35: Hm - LLVM seems to emit it at least for `thumbv7em-none-eabihf`, which doesn't list that target feature https://rust.godbolt.org/z/13nWEzY53

beetrees: `vfp2sp` isn't a Rust target feature, so it won't get a `cfg(target_feature = "...")`.

tgross35: Ah, okay, so that feature is implied by `vfp2` but also is enabled by default on some Arm targets? That's kind of hard to approximate, any reason not to just add that target feature?

beetrees: Not that I'm aware of.

taiki-e:
> It looks like you could also check for the `fpregs` without `soft-float`? I'm not sure whether that's more accurate. @taiki-e you may have some thoughts here.

> ```diff
>  target_feature="fpregs"
> -target_feature="soft-float"
>  target_feature="thumb2"
> ```

Hmm. It's odd that fpregs,vfp2,vfp3,d32 are enabled for soft-float targets by default.

Given that they are enabled for armv7-unknown-linux-gnueabi and armv7a-none-eabi but not for arm-unknown-linux-gnueabi, armv7r-none-eabi, and thumbv7em-none-eabi, I guess LLVM infer that fregs are available for `v7` + `aclass`.

libm/src/math/arch/mod.rs
tgross35: Wrote up https://github.com/rust-lang/rust/issues/161826.

In the meantime @silverstillisntgold could you change this gate to only `all(target_arch = "arm")` then gate the individual function `sqrtf` on `any(target_feature = "vfp2", target_feature = "vfp2sp")`? This means we will only get the hard float ops on non-thumb targets until `vfp2sp2` is exposed, but that's fine.

(You'll need to update https://github.com/rust-lang/compiler-builtins/blob/e407190291f9b2e83c10babdeafca248c3a862a0/libm/Cargo.toml#L47 to account for the unknown `target_feature`, and possibly add the same thing to `builtins-shim/Cargo.toml`)

tgross35: Correction, keep this gated on `all(target_arch = "arm", target_abi = "eabihf")`. We shouldn't clobber the fp state if using the softfloat ABI.

silverstillisntgold:
Spent some time looking into this yesterday and it's a weird situation. The `eabihf` ending specifically indicates that the target can pass fp32 (and maybe f64) values via registers, which means the target should have fp registers, which implies it should have an FPU (https://doc.rust-lang.org/rustc/platform-support/arm-none-eabi.html#instruction-sets). This is why the `-fpregs` flag is suggested as the means to disable the FPU on targets which support it by default. And LLVM seems to assume FPU's include a `sqrt`, so using `all(target_arch = "arm", target_abi = "eabihf")` should indeed be sufficient for detecting that a sqrt instruction supporting at least f32 is available.

The whole vfp* family is a chain of features each of which is a superset of those beneath it, and which are most often implied rather than being directly enabled. [Thumbv8m](https://github.com/rust-lang/rust/blob/main/compiler/rustc_target/src/spec/targets/thumbv8m_main_none_eabihf.rs) uses `+fp-armv8d16sp` as a baseline, and you'll notice it ends with `sp`, which is f32. This implies `vfp5sp`, and all single precision version before it. Most of this seems to happen on the [LLVM side](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/ARM/ARMProcessors.td), but some of it can be seen [here](https://github.com/rust-lang/rust/blob/main/compiler/rustc_target/src/target_features.rs#L160).

You can see that something like [Armv7a](https://github.com/rust-lang/rust/blob/main/compiler/rustc_target/src/spec/targets/armv7a_none_eabihf.rs) has the `+vfp3d16` feature (no `sp` suffix). So it supports f64 (which implies f32) by default.

Doing a feature test for `vfp2` is definitely the most convenient way to handle enabling a f64 variant, but arm features are currently [unstable](https://github.com/rust-lang/rust/issues/150246). Right now doing the f32 variant should be easy, but for f64 the only stable way I can think of is to match on specific `target_cpu` values (or even specific target triples) which are assumed by LLVM to have f64 support (M7, M55, M85, etc.). This approach would probably be super ugly.

I can finish up reworking the sqrtf feature gating, but I think a proper f64 sqrt implementation should probably be it's own PR. Might even be best to block it on the stabilization of arm features. Targets which have f64 tend to be more powerful than those with only f32 (f32 only are mostly embedded targets), so it could be argued that they benefit less from the hardware acceleration and can hold off just a bit longer with the current software implementation.

silverstillisntgold:
> > It looks like you could also check for the `fpregs` without `soft-float`? I'm not sure whether that's more accurate. @taiki-e you may have some thoughts here.
> 
> > ```diff
> >  target_feature="fpregs"
> > -target_feature="soft-float"
> >  target_feature="thumb2"
> > ```
> 
> Hmm. It's odd that fpregs,vfp2,vfp3,d32 are enabled for soft-float targets by default.
> 
> Given that they are enabled for armv7-unknown-linux-gnueabi and armv7a-none-eabi but not for arm-unknown-linux-gnueabi, armv7r-none-eabi, and thumbv7em-none-eabi, I guess LLVM infer that fregs are available for `v7` + `aclass`.

My understanding is that the eabihf suffix only impacts the use of FP registers for passing values per [this explanation](https://doc.rust-lang.org/rustc/platform-support/arm-none-eabi.html#instruction-sets). It seems to be for literally nothing more than to constrain the ABI. You can still use fp ops within your code and if you have features enabled that let you use hardware fp ops then LLVM should still emit them.

The [Cortex-A7](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/ARM/ARMProcessors.td#L188) lists `FeatureVFP4`, so it can use f64/f32 ops, just not in a way that shows up in the ABI.

tgross35:
> Spent some time looking into this yesterday and it's a weird situation. The `eabihf` ending specifically indicates that the target can pass fp32 (and maybe f64) values via registers, which means the target should have fp registers, which implies it should have an FPU (https://doc.rust-lang.org/rustc/platform-support/arm-none-eabi.html#instruction-sets). This is why the `-fpregs` flag is suggested as the means to disable the FPU on targets which support it by default.

My understanding is that `eabihf` implies two things:

1. The `fpregs` feature
2. We can actually use the FPU and clobber its state

So gating the module on `all(target_arch = "arm", target_abi = "eabihf")`, as it is now, makes sense. 

> And LLVM seems to assume FPU's include a `sqrt`, so using `all(target_arch = "arm", target_abi = "eabihf")` should indeed be sufficient for detecting that a sqrt instruction supporting at least f32 is available.

The problem is that `fpregs` doesn't imply `vfp2` or `vfp2sp` on the Rust side https://github.com/rust-lang/rust/blob/d9dd0703ba332dbdd19ae6c9bf7ab98ab7c933c7/compiler/rustc_target/src/target_features.rs#L180 so we don't actually reject `-Ctarget-feature=-vfp2,-vfp2sp2`. LLVM currently crashes but the combination of vreg ABI with soft float math is theoretically possible, or we could get a new weird target that uses vregs but can't do `vsqrt.f32` for some reason.

I assume that will be clarified as part of @adamgemmell's refactor (mentioned at https://github.com/rust-lang/rust/issues/161826) but for the time being, I'd prefer to gate the individual function on `any(target_feature = "vfp2", target_feature = "vfp2sp")` to be on the safe side. It is unfortunate that this will only work with nightly, but we really can't do better until https://github.com/rust-lang/rust/issues/150246 is stabilized. Or update `target_features.rs` to indicate that `fpregs` implies `vfp2sp` - but I don't know that's actually something we'd want to do.

> I can finish up reworking the sqrtf feature gating, but I think a proper f64 sqrt implementation should probably be it's own PR. Might even be best to block it on the stabilization of arm features.

Up to you! I'm happy to take both together since they're pretty similar, with a `#[cfg(any(target_feature = "vfp2", target_feature = "vfp2sp"))]` gate for `sqrtf` and `#[cfg(target_feature = "vfp2")]` for `sqrt`, but it's also fine to split.

silverstillisntgold:
If the only safe option is to use flags only available on nightly then it's probably best to delay this until `arm_target_feature` stabilizes. Since if a user is already on nightly they may as well use `core_float_math` or `core_intrinsics`, where LLVM already correctly handles any possible lowering and they can avoid pulling in another dependency.

tgross35:
I think that's orthogonal, if you use `libm` and build with nightly you'll get these implementations without needing to do anything special or use `#![feature]` (i.e. your crate doesn't effectively become nightly-only). And `arm_target_feature` could very well stabilize first if Adam is working on it.

Even once `core_float_math` stabilizes, there will still be a number of crates that must continue to use `libm` to meet the MSRV.

tgross35:
`target_feature = “vfp2sp2”` will be supported in the upcoming nightly, so that simplifies one of the issues.

silverstillisntgold:
Seems like `CARGO_CFG_TARGET_ABI` isn't emitted by cargo in the MSRV version check 😢

libm/configure.rs
```diff
@@ -63,6 +64,7 @@ impl Config {
             opt_level: env::var("OPT_LEVEL").unwrap(),
             cargo_features,
             target_arch: env::var("CARGO_CFG_TARGET_ARCH").unwrap(),
+            target_abi: env::var("CARGO_CFG_TARGET_ABI").unwrap(),
```
tgross35:
```suggestion
            // FIXME(msrv): defined since 1.78
            target_abi: env::var("CARGO_CFG_TARGET_ABI").ok(),
```

and make the `target_abi` optional

libm/configure.rs
```diff
@@ -114,6 +116,14 @@ pub fn emit(cfg: &Config) {
     // We have to cfg our code accordingly.
     let thumb_1 = split[0] == "thumbv6m" || split[0] == "thumbv8m.base";
 
+    // Detect Arm32/Thumb targets which have an FPU.
+    // Ensure the architecture is correct, then that rustc is using the HF abi,
+    // then that the target supports the required feature.
+    let arm_f32_hf =
+        cfg.target_arch == "arm" && cfg.target_abi == "eabihf" && cfg.has_target_feature("vfp2sp");
+    // Support for f64 implies support for f32.
+    let arm_f64_hf = arm_f32_hf && cfg.has_target_feature("vfp2");
```
tgross35:
```suggestion
    // FIXME(msrv): `cfg(target_abi)` is usable since 1.78
    let target_abi_eabihf = cfg.target_abi.as_deref() == Some("eabihf");
```

Just an alias for the thing we can't do at MSRV is sufficient. The `x86_no_sse2` alias exists because we use it a lot and the workaround is clunky, and there's really no alternative to the `thumb` aliases.

libm/src/math/sqrt.rs
```diff
@@ -21,6 +21,7 @@ pub fn sqrtf(x: f32) -> f32 {
         name: sqrtf,
         use_arch: any(
             all(target_arch = "aarch64", target_feature = "neon"),
+            arm_f32_hf,
```
tgross35: For the above,

```suggestion
            all(target_arch = "arm", target_abi_eabihf, target_feature = "vfp2sp")
```

libm/src/math/sqrt.rs
```diff
@@ -38,6 +39,7 @@ pub fn sqrt(x: f64) -> f64 {
         name: sqrt,
         use_arch: any(
             all(target_arch = "aarch64", target_feature = "neon"),
+            arm_f64_hf,
```
tgross35:
```suggestion
            all(target_arch = "arm", target_abi_eabihf, target_feature = "vfp2")
```

Similarly

libm/src/math/arch/mod.rs
```diff
@@ -41,6 +41,13 @@ cfg_select_nofmt! {
             sqrtf16,
         };
     }
+    arm_f32_hf => {
+        mod arm;
+
+        pub use arm::sqrtf;
+        #[cfg(arm_f64_hf)]
+        pub use arm::sqrt;
+    }
```
tgross35:
```suggestion
    all(target_arch = "arm", target_abi_eabihf) => {
        mod arm;
        
        #[cfg(target_feature = "vfp2sp")]
        pub use arm::sqrtf;
        #[cfg(target_feature = "vfp2")]
        pub use arm::sqrt;
    }
```

Since we could theoretically wind up with this enabled on targets without vfp2sp

libm/src/math/arch/arm/sqrt.rs
```diff
@@ -0,0 +1,28 @@
+use core::arch::asm;
+
+pub fn sqrtf(mut x: f32) -> f32 {
```
tgross35, line 2:
```suggestion
#[cfg(target_feature = "vfp2sp")]
pub fn sqrtf(mut x: f32) -> f32 {
```

Ditto

libm/src/math/arch/arm/sqrt.rs
```diff
@@ -0,0 +1,28 @@
...
+            "vsqrt.f32 {x}, {x}",
+            x = inout(sreg) x,
+            options(nostack, nomem, pure),
+        );
+    }
+    x
+}
+
+#[cfg(arm_f64_hf)]
+pub fn sqrt(mut x: f64) -> f64 {
```
tgross35:
```suggestion
#[cfg(target_feature = "vfp2")]
pub fn sqrt(mut x: f64) -> f64 {
```

Alias change

libm/src/math/arch/arm/mod.rs
```diff
@@ -0,0 +1,7 @@
+//! Architecture-specific support for Arm32/Thumb with hard float ABI.
+
+mod sqrt;
+
+#[cfg(arm_f64_hf)]
+pub use sqrt::sqrt;
+pub use sqrt::sqrtf;
```
tgross35:
```suggestion
#[cfg(target_feature = "vfp2")]
pub use sqrt::sqrt;
#[cfg(target_feature = "vfp2sp")]
pub use sqrt::sqrtf;
```

libm/src/math/arch/arm/sqrt.rs
```diff
@@ -0,0 +1,28 @@
...
+    // SAFETY: vsqrt.f64 is available for this backend.
```
tgross35:
```suggestion
    // SAFETY: vsqrt.f64 is available with `vfp2`.
```

libm/src/math/arch/arm/sqrt.rs
```diff
@@ -0,0 +1,28 @@
+use core::arch::asm;
+
+pub fn sqrtf(mut x: f32) -> f32 {
+    // SAFETY: vsqrt.f32 is available for this backend.
```
tgross35:
```suggestion
    // SAFETY: vsqrt.f32 is available with `vfp2sp`.
```

libm/src/math/arch/arm/sqrt.rs
```diff
@@ -0,0 +1,28 @@
...
+    // SAFETY: vsqrt.f64 is available for this backend.
+    // Need to use `dreg_low16` because `dreg` implies `d32`,
+    // which isn't required to support an f64 FPU.
+    unsafe {
+        asm!(
+            "vsqrt.f64 {x}, {x}",
+            x = inout(dreg_low16) x,
+            options(nostack, nomem, pure),
+        );
+    }
```
tgross35: Perhaps `dreg_low8` in case we get other features?

Not sure if there's hardware with f64 support but only 8 registers, but these functions are pretty much always standalone so shouldn't hurt.

tgross35 (changes requested):
A handful of very small changes but with them, I think this should be all set. Could you please squash as well?

(the first commit message is outdated too, since `sreg` now works)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1300/b2a59c9550afc1fabdc744b15a220c67ec166a9c..a8e9ab81425a0c1e7164a0f0c9cc940f10344f99)*

libm/configure.rs
```diff
@@ -123,6 +129,7 @@ pub fn emit(cfg: &Config) {
     // Arch shorthand config is used in most crates.
     set_cfg("thumb", thumb);
     set_cfg("thumb_1", thumb_1);
+    set_cfg("target_abi_eabihf", target_abi_eabihf);
     set_cfg("x86_no_sse2", x86_no_sse2);
```
tgross35, line 133:
```suggestion
    set_cfg("x86_no_sse2", x86_no_sse2);
    // FIXME(arm_target_feature): the exact target features are not yet stable.
    println!("cargo::rustc-check-cfg=cfg(target_feature, values("vfp2sp", "vfp2"))");
```

Guess there's one more thing, then LGTM assuming the rest of CI passes

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1300/0cda4eeeffb618b92a22697808981154656ff762..14f60ee60cf1c803d34f69d58061b6a79acc24b6)*

silverstillisntgold:
Remaining failures are due to ubuntu-26.04 benches using `rustc 1.100.0-nightly (e71c0f1e3 2026-08-18)`, which doesn't have the `vfp2sp` feature.

libm/configure.rs
```diff
@@ -123,7 +129,10 @@ pub fn emit(cfg: &Config) {
...
+    // FIXME(arm_target_feature): the exact target features are not yet stable.
+    println!("`cargo::rustc-check-cfg=cfg(target_feature, values(\"vfp2sp\", \"vfp2\"))`");
```
tgross35:
```suggestion
    // FIXME(arm_target_feature): the exact target features are not yet stable.
    println!("cargo:rustc-check-cfg=cfg(target_feature, values(\"vfp2sp\", \"vfp2\"))");
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1300/d680c4fba52a24a83c1893ce59e99cfd3b77fd71..b697783eb75db8fdca97c38a48c40a7c03b220f6)*

tgross35:
> Remaining failures are due to ubuntu-26.04 benches using `rustc 1.100.0-nightly (e71c0f1e3 2026-08-18)`, which doesn't have the `vfp2sp` feature.

Ah yeah I guess it's only old nightlies that don't work. Not the biggest deal as long as we don't break stable, but should be an easy fix once we get the syntax right.

libm/configure.rs
silverstillisntgold: CI failed on MSRV without the backticks

silverstillisntgold:
> The problem is that `fpregs` doesn't imply `vfp2` or `vfp2sp` on the Rust side https://github.com/rust-lang/rust/blob/d9dd0703ba332dbdd19ae6c9bf7ab98ab7c933c7/compiler/rustc_target/src/target_features.rs#L180 so we don't actually reject `-Ctarget-feature=-vfp2,-vfp2sp2`. LLVM currently crashes but the combination of vreg ABI with soft float math is theoretically possible, or we could get a new weird target that uses vregs but can't do `vsqrt.f32` for some reason.
> 
> I assume that will be clarified as part of @adamgemmell's refactor (mentioned at [rust-lang/rust#161826](https://github.com/rust-lang/rust/issues/161826)) but for the time being, I'd prefer to gate the individual function on `any(target_feature = "vfp2", target_feature = "vfp2sp")` to be on the safe side. It is unfortunate that this will only work with nightly, but we really can't do better until [rust-lang/rust#150246](https://github.com/rust-lang/rust/issues/150246) is stabilized. Or update `target_features.rs` to indicate that `fpregs` implies `vfp2sp` - but I don't know that's actually something we'd want to do.

Before this is merged I'd like to again try to make the case for treating `eabihf` as enough of a requirement for `sqrtf`. Yes, someone could pass `-Ctarget-feature=-vfp2,-vfp2sp2`, but those features are unstable and I doubt many people even know of their existence. I checked all 28 `eabihf` targets in stable's [spec/targets](https://github.com/rust-lang/rust/tree/1.98.0/compiler/rustc_target/src/spec/targets) and all of them support at least `vfp2sp`, either explicitly via their `feature` field or through their backend `llvm_target` representation. Most are `vfp3`, with the thumb targets mostly being `vfp4sp` with reduced register counts. Using `-Ctarget-cpu` is stable and can only add features to the target baseline, so no risk of specific CPU targets causing issues.

I'd argue that users explicitly opting to use the hard float ABI, then using unstable features to disable floating point operations just don't exist. Even though it's theoretically possible, it's a contradicting combination that's entirely impractical. And I can't foresee a world where a new target is added which somehow manages to reasonably combine these two opposing behaviors.

libm/src/math/arch/arm/mod.rs
silverstillisntgold: Left it just `arm` for now since that maintains the same naming scheme as the other target backends. In the future if arm specific soft implementations are found to be useful then the `arm` module can have internal `hf` and `sf` modules containing their respective implementations.

tgross35:
My concern is still that the feature will be stable in the near future (hopefully!), and https://github.com/rust-lang/rust/blob/70222712809cd5cc1718ed8995914a1cbacb6b92/compiler/rustc_target/src/target_features.rs#L181 does not say `fpregs` implies `vfp2sp`. It's a one line change to make rustc promise that and I'd be happy to reflect it here once that happens, but until then it will be a valid, if niche, combination (once stable).

We can keep discussing because maybe there's an alternative (detect whether or not `arm_target_feature` is stable?), but I'd rather get this merged in its current state since it's so close. We can always relax things in the future.

silverstillisntgold:
> My concern is still that the feature will be stable in the near future (hopefully!), and https://github.com/rust-lang/rust/blob/70222712809cd5cc1718ed8995914a1cbacb6b92/compiler/rustc_target/src/target_features.rs#L181 does not say `fpregs` implies `vfp2sp`. It's a one line change to make rustc promise that and I'd be happy to reflect it here once that happens, but until then it will be a valid, if niche, combination (once stable).
> 
> We can keep discussing because maybe there's an alternative (detect whether or not `arm_target_feature` is stable?), but I'd rather get this merged in its current state since it's so close. We can always relax things in the future.

I don't think that `fpregs` will ever imply `vfp2sp` because [`vfp2sp` now implies `fpregs`](https://github.com/rust-lang/rust/blob/70222712809cd5cc1718ed8995914a1cbacb6b92/compiler/rustc_target/src/target_features.rs#L211). I'm not sure how rustc handles circular dependencies in target features but there don't seem to be any others in `target_features.rs` so I'm guessing it's not functionality that's supported.

My argument is that even though the `fpregs` feature itself doesn't enable fp operations, use of an `eabihf` target is the user explicitly declaring they have a CPU/MCU which is known to have fp registers and an FPU (otherwise they'd use the equivalent soft-float `eabi` target), and so there's no reason to support them disabling fp ops provided by the FPU they've just knowingly opted in to (even though it's technically possible). And since all currently available (and likely any future released) arm/thumb targets which provide `eabihf` targets declare at minimum support for f32 regs/FPU, it's entirely reasonable to infer that all arm/thumb `eabihf` targets can safely call `sqrtf`. The f64 `sqrt` would still be gated on `vfp2` either way since it's not a baseline across all arm/thumb targets.

If you'd still rather have both gated behind the features then I totally understand, and it's probably fine to merge for now, assuming it all looks good to you.

adamgemmell:
Use of the `vfp*` features is what signifies the presence of an FPU, not use of a hf target. What does gating this on `fpregs` achieve, given that you say disabling `vfp2sp` won't happen in practice? It's also worth pointing out that removing a target_feature from a function in the future is not a breaking change, but adding it back is.

There's alternative implementations to a circular feature dependency - we could tie them together or fold them together under the same feature name, which have slightly different options available for splitting them post-stabilisation.

> LLVM currently crashes but the combination of vreg ABI with soft float math is theoretically possible, or we could get a new weird target that uses vregs but can't do vsqrt.f32 for some reason.

Reading this suggests to me that we shouldn't expose `fpregs` at all (or leave it unstable) until that area can be worked out properly. That would bring it inline with aarch64 (and I think x86_64?) targets where the FPU and fp regs features are one.

I noticed the fact that FPU features are enabled on some a32 softfloat targets too and I think that's a mistake. I wouldn't use 32 bit arm target specs as an authoritative source on how something should work 😄

silverstillisntgold:
> Use of the `vfp*` features is what signifies the presence of an FPU, not use of a hf target. What does gating this on `fpregs` achieve, given that you say disabling `vfp2sp` won't happen in practice? It's also worth pointing out that removing a target_feature from a function in the future is not a breaking change, but adding it back is.

I'm advocating for gating the `sqrtf` implementation on the presence of the `eabihf` target ABI, not on the `fpregs` flag. Compiling for the `hf` ABI implies the user knows their target has some level of fp support, and all relevant targets support an f32 FPU as a baseline (they have **at least** `vfp2sp` enabled). From rustc's [platform support](https://doc.rust-lang.org/rustc/platform-support/arm-none-eabi.html#instruction-sets):

"Rust targets ending in eabihf use the so-called hard-float ABI: functions which take f32 or f64 as arguments will have them passed via FPU registers. These targets **therefore require the availability of an FPU** and will assume some baseline level of floating-point support is available (which can vary depending on the target)."

> There's alternative implementations to a circular feature dependency - we could tie them together or fold them together under the same feature name, which have slightly different options available for splitting them post-stabilisation.
> 
> > LLVM currently crashes but the combination of vreg ABI with soft float math is theoretically possible, or we could get a new weird target that uses vregs but can't do vsqrt.f32 for some reason.
> 
> Reading this suggests to me that we shouldn't expose `fpregs` at all (or leave it unstable) until that area can be worked out properly. That would bring it inline with aarch64 (and I think x86_64?) targets where the FPU and fp regs features are one.

What's currently being proposed wouldn't require direct access to `fpregs`.

> I noticed the fact that FPU features are enabled on some a32 softfloat targets too and I think that's a mistake. I wouldn't use 32 bit arm target specs as an authoritative source on how something should work 😄

Even though it's weird this may very well be intentional, from the same rustc platform support doc linked above:

"Rust targets ending with eabi use the so-called soft-float ABI: functions which take f32 or f64 as arguments will have those values packed into integer registers. This means that an FPU is not required from an ABI perspective, but within a function floating-point instructions may still be used if the code is compiled with a target-cpu or target-feature option that enables FPU support."

adamgemmell:
Still agree with Trevor that `vfp2sp` is required: https://rust.godbolt.org/z/jfjx79adz (this error originates from LLVM). The feature can always be removed in the future

> Even though it's weird this may very well be intentional, from the same rustc platform support doc linked above:
> 
> "Rust targets ending with eabi use the so-called soft-float ABI: functions which take f32 or f64 as arguments will have those values packed into integer registers. This means that an FPU is not required from an ABI perspective, but within a function floating-point instructions may still be used if the code is compiled with a target-cpu or target-feature option that enables FPU support."

Interesting, thanks for the link

silverstillisntgold:
I also agree that `vfp2sp` is definitely a hard requirement. But because `vfp2sp` is enabled on all hf targets, I believe it's fine if the compilation argument combination you shared fails to compile. If a user wants to use a specific ABI without `eabihf` they should use the provided [`sf` equivalent](https://rust.godbolt.org/z/ojdTKf4rc) (which also fails in this example), instead of choosing a target which provides a set of features then immediately disabling those features.

libm/configure.rs
tgross35: Well the backticks make it do nothing :) meaning we still get the warnings on versions that support `rustc-check-cfg` but don't know the `target_feature` values, such as stable (not checked in CI, I should add that) or the nightly used for the benchmarks. They just don't show up in MSRV checks because it predates check-cfg.

Note the single colon in the suggestion, I don't think it was tested with that variant

libm/src/math/arch/arm/sqrt.rs
```diff
@@ -0,0 +1,31 @@
...
+
+#[cfg(target_feature = "vfp2")]
+pub fn sqrt(mut x: f64) -> f64 {
+    // SAFETY: vsqrt.f64 is available with `vfp2`.
+    // Can't just use plain old `dreg` because `vfp2` only provides a base
+    // level of FPU support with a non-specified amount of fp64 registers.
+    // It's fine to use `dreg_low16` here, but use `dreg_low8` instead to
+    // keep the requirements as low as possible.
+    unsafe {
+        asm!(
```
tgross35: Just realized we'll get a warning here

```suggestion
#[cfg(target_feature = "vfp2sp")]
pub fn sqrtf(mut x: f32) -> f32 {
    // SAFETY: vsqrt.f32 is available with `vfp2sp`.
    unsafe {
        core::arch::asm!(
            "vsqrt.f32 {x}, {x}",
            x = inout(sreg) x,
            options(nostack, nomem, pure),
        );
    }
    x
}

#[cfg(target_feature = "vfp2")]
pub fn sqrt(mut x: f64) -> f64 {
    // SAFETY: vsqrt.f64 is available with `vfp2`.
    // Can't just use plain old `dreg` because `vfp2` only provides a base
    // level of FPU support with a non-specified amount of fp64 registers.
    // It's fine to use `dreg_low16` here, but use `dreg_low8` instead to
    // keep the requirements as low as possible.
    unsafe {
        core::arch::asm!(
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1300/d680c4fba52a24a83c1893ce59e99cfd3b77fd71..b697783eb75db8fdca97c38a48c40a7c03b220f6)*

tgross35:
@adamgemmell how close do you think `arm_target_feature` is? In theory we could stabilize using `vfp2sp` and `vfp2` in `cfg` but forbid toggling them, since they don't seem like the problematic part of https://github.com/rust-lang/rust/pull/149512#discussion_r2584941260, which would make this issue go away.

> If you'd still rather have both gated behind the features then I totally understand, and it's probably fine to merge for now, assuming it all looks good to you.

Yeah, I'd prefer to get this merged and we can figure something out in a followup. Just needs the two open things resolved to take care of warnings.

I guess one option is to have a list of current eabihf targets and manually set `target_feature = "vfp2sp"` or `target_feature = "vfp2"` if (1) the target is in that list, (2) there are no `target_feature`s, meaning it's running with a version prior to `arm_target_feature` being stabilized. (1) would prevent my concern about miscompiling with a future target that is eabihf but doesn't have the right instructions, if such a thing were ever to exist, and (2) would resolve the concern about users toggling `vfp2sp` off since there's no way to do that stably.

tgross35 (approved):
Thanks for sticking this with, looks great to me!

If you're interested in following up to get this working on stable, I think a triple->features map as I mentioned in the previous comment would work.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1300/f662be3408ab669689f308423c022fa4cdd5af5b..e3c7088178000c6e4fefe4e500630a55beb2ea2c)*

adamgemmell:
I'm scheduled to work on them next quarter, and I'd like to gather up and run the pros/cons by people before stabilising any features, in particular FPU features. Once set we've moved to stabilise the features relatively quickly in the past (a few months?). Is the motivation mainly unblocking using these functions on nightly?

tgross35:
Oh right sorry, you did say that on the issue. For these specific features, having them stable would mean mean we get to use these and future asm ops without nightly—I was just weighing whether a workaround is worth it for now. Or, if other features are blocked on naming for the time being, whether there might be a subset we could stabilize sooner.

adamgemmell:
The combinatorial explosion of fpu features might be a contention point, for example `vfp2` can be modelled as `vfp2sp` + `fp64` so we might not want to expose that. That might have knock-on effects for naming of the minimum feature levels like `vfp2sp`. For those reasons I'd prefer not to stabilise either for now.

silverstillisntgold:
@tgross35 @adamgemmell Sorry to bring this back from the dead. Was researching the required feature levels for ARM32/Thumb fma and RISC-V sqrt/fma and found that version 1.97 stabilized [CfgStableToggleUnstable](https://github.com/rust-lang/rust/pull/155962), which is specifically for unstable features which can be used in `#[cfg(x)]` feature gating. The RISC-V team then used this to [stabilize using d, e, and f extensions](https://github.com/rust-lang/rust/pull/156188) in cfg blocks in version 1.98 without fully stabilizing the features.

Is this something worth pursuing for the ARM features, ahead of full stabilization of `arm_target_feature`?

silverstillisntgold:
> @tgross35 @adamgemmell Sorry to bring this back from the dead. Was researching the required feature levels for ARM32/Thumb fma and RISC-V sqrt/fma and found that version 1.97 stabilized [CfgStableToggleUnstable](https://github.com/rust-lang/rust/pull/155962), which is specifically for unstable features which can be used in `#[cfg(x)]` feature gating. The RISC-V team then used this to [stabilize using d, e, and f extensions](https://github.com/rust-lang/rust/pull/156188) in cfg blocks in version 1.98 without fully stabilizing the features.
> 
> Is this something worth pursuing for the ARM features, ahead of full stabilization of `arm_target_feature`?

The pull to stabilize d, e, and f was actually reverted, but only because of concerns around the `e` feature being a negative feature. The [follow-up](https://github.com/rust-lang/rust/pull/161385) for stabilizing just `d` and `f` looks to be on track to make it into version 1.100, albeit using `Stable` instead of `CfgStableToggleUnstable`. Still, now that the functionality is in the compiler this might be a reasonable use for it.

tgross35:
I think that, based on what Adam was saying, there's a chance that the names might not actually stay `vfp2sp` and `vfp2` - in which case we wouldn't want to stabilize them in any form yet. Though that's a good option once the names are figured out, if we're not yet sure what should be toggleable.

PR #1303, opened 2026-08-26
use `extern "C"` and  C names for complex div and mul of `f32` and `f64`

folkertdev:
The `f16` and `f128` versions have abi issues, so continue to use `extern "Rust"` with a `__rust` prefix.

### DIFF
--- a/builtins-test/tests/complex.rs
+++ b/builtins-test/tests/complex.rs
@@ -382,45 +382,33 @@ mod complex {
         };
     }
 
-    #[cfg(all(f16_enabled, not(x86_no_sse2)))]
+    #[cfg(f16_enabled)]
     complex_mul! {
         f16, __rust_mulhc3, 1.0e-3;
     }
 
-    #[cfg(all(f16_enabled, not(x86_no_sse2)))]
+    #[cfg(f16_enabled)]
     complex_div! {
         f16, __rust_divhc3, 1.0e-3;
     }
 
     complex_mul! {
-        f32, __rust_mulsc3, 1.0e-6;
-        f64, __rust_muldc3, 1.0e-9;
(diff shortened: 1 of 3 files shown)


tgross35:
For the status update, this is pretty much waiting on the resolution of https://github.com/rust-lang/rust/pull/161814

folkertdev:
Is it? the solution here works either way right? We can remove the `extern "Rust"` variants again when we do merge that PR you linked but otherwise this should be fine?

builtins-test/tests/complex.rs
```diff
@@ -382,45 +382,33 @@ mod complex {
...
+    }
 
-            complex_div! {
-                f128, __rust_divtc3, 1.0e-12;
-            }
-        }
+    #[cfg(f128_enabled)]
+    complex_div! {
+        f128, __rust_divkc3, 1.0e-12;
     }
```
tgross35, line 413: These still have the PPC names

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1303/dd502243dc0d5f99aa50b6bb3fad53ec9a4df0ce..0e71f338df8c76729fbadb13cb7cddd7bcc3c2f2)*

folkertdev: ah, and locally they don't run when you cross-compile to powerpc (because no vsx). anyhow, fixed

tgross35 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1303/dd502243dc0d5f99aa50b6bb3fad53ec9a4df0ce..d0a1fb28be59064268716de9faf34b95f307e43f)*

PR #1304, opened 2026-08-30
c-b: Drop the removed `abi_unadjusted` feature gate

HuzaifaAbdulRehman:
`abi_unadjusted` was removed in rust-lang/rust#161398, so every job on current nightly stops at E0557 in `lib.rs:6`. Main is red for the same reason.

Nothing uses `extern "unadjusted"`, so the gate goes rather than being renamed.

Not built locally, no toolchain here. I used an LLM assistant while looking into this.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -114,9 +114,10 @@ jobs:
           os: windows-2025-vs2026
         - target: x86_64-pc-windows-msvc
           os: windows-2025-vs2026
-        - target: i686-pc-windows-gnu
-          os: windows-2025-vs2026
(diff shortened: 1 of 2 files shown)


tgross35 (approved):
Thanks! I added a commit disabling the i686-pc-windows-gnu job since it's broken, https://github.com/rust-lang/compiler-builtins/issues/1306

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1304/5c5f07851b1878013ac81e129b0517feaaf8661d..2187ae312f4346dba4151f7e1c4bc9fdf65f3756)*

PR #1317, opened 2026-09-07
Add `aeabi_u(read|write)(4|8)` intrinsics.

### LINKED ISSUE #162259: Undefined symbol errors on `__aeabi_uread*` and   `__aeabi_uwrite*` functions with opt-level = "z" and `build-std`
### Code
On more recent Rust versions (after update to LLVM 23, I assume), I no longer can build some bare-metal projects.
There are errors such as 
```
 = note: rust-lld: error: undefined symbol: __aeabi_uread4
          >>> referenced by mod.rs:1758 (src\ptr\mod.rs:1758)
```
Also mentioning `__aeabi_uread8`, `__aeabi_uwrite4`, `__aeabi_uread8`.
This only happens with option `opt-level = "z"` AND `build-std` enabled.

.cargo/cargo.toml:
```toml
[target.'cfg(all(target_arch = "arm", target_os = "none"))']
# replace STM32G0B1RETx with your chip as listed in `probe-rs chip list`
runner = "probe-rs run --chip STM32G0B1RETx"

[build]
target = "thumbv6m-none-eabi"

[env]
DEFMT_LOG = "trace"

[unstable]
build-std = ["core"]
```

Problem can be easily reproduced on embassy project examples 
e.g. (https://github.com/embassy-rs/embassy/blob/main/examples/stm32g0/src/bin/blinky.rs)
This example's target is `thumbv6m-none-eabi`.

### Version it worked on

It most recently worked on: Rust 1.95

### Version with regression

`rustc --version --verbose`:
```
rustc 1.100.0-nightly (2e2b193f8 2026-09-02)
```

Dirbaio:
This is a 1:1 port of `aeabi_u(read|write)(4|8)` from the compiler-rt sources.

Fixes https://github.com/rust-lang/rust/issues/162259

One note: the bug happened because build.rs [blanket excludes](https://github.com/rust-lang/compiler-builtins/blob/451f8f4bb295b2f1f374973582b4ddf48e62b1a4/compiler-builtins/build.rs#L472-L483) all .S files for thumbv6m and thumbv8m.base.
THis feels risky to me becaues nothing prevents another new intrinsic to get accidentally forgotten in the future. Compiler-rt uses an explicit include-list for these targets, but if we did that it would be equally risky (we could forget to add a file as well). We could change it to an exclude list but maybe this increases the maintenance burden. I've checked the only affected intrinsics are the 4 this PR adds (all others are already ported), so I've chosen to not touch that code for now.

Another note: Big endian is not covered by any test AFAICT. 

LLM disclosure: I've used Claude to port the ASM and write the tests, then reviewed and fixed minor stuff manually.


compiler-builtins/src/arm.rs
```diff
@@ -277,4 +277,269 @@ intrinsics! {
...
+            "ldrb r2, [r0, #1]",
+            "lsls r2, r2, #16",
+            "ldrb r0, [r0]",
+            "lsls r0, r0, #24",
+            "adds r0, r0, r2",
+            "adds r0, r0, r1",
+            "bx lr",
+        );
+        #[cfg(all(
+            not(all(target_feature = "thumb-mode", not(target_feature = "thumb2"))),
```
thejpster: `any(target_feature = "thumb2", not(target_feature = "thumb-mode")),` might be easier to read here?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1317/451f8f4bb295b2f1f374973582b4ddf48e62b1a4..f8fdb2588720159ad67e9adcb3aa6c0cf5b021de)*

compiler-builtins/src/arm.rs
```diff
@@ -277,4 +277,269 @@ intrinsics! {
         // SAFETY: memclr preconditions apply, less strict alignment.
         unsafe { __aeabi_memset4(dst, n, 0) };
     }
+
+    // Read a `u32` from a possibly unaligned address.
+    //
```
thejpster, line 286: Shouldn't these have `///`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1317/451f8f4bb295b2f1f374973582b4ddf48e62b1a4..f8fdb2588720159ad67e9adcb3aa6c0cf5b021de)*

compiler-builtins/src/arm.rs
```diff
@@ -277,4 +277,269 @@ intrinsics! {
...
+    // # Safety
+    //
+    // `address` must be valid for writing eight bytes.
+    #[unsafe(naked)]
+    #[cfg(not(any(target_vendor = "apple", target_env = "msvc")))]
+    pub unsafe extern "aapcs" fn __aeabi_uwrite8(value: u64, address: *mut u8) -> u64 {
+        #[cfg(target_endian = "little")]
+        core::arch::naked_asm!(
+            "strb r0, [r2, #0]",
```
thejpster: Some notes on which function params have gone into which registers might be useful for anyone unfortunate enough to look at this in the future.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1317/451f8f4bb295b2f1f374973582b4ddf48e62b1a4..f8fdb2588720159ad67e9adcb3aa6c0cf5b021de)*

builtins-test-intrinsics/src/main.rs
```diff
@@ -389,6 +390,31 @@ mod intrinsics {
...
+
+    #[optimize(size)]
+    pub fn aeabi_uwrite4(x: u32, p: *mut u8) {
+        unsafe { p.cast::<u32>().write_unaligned(x) }
+    }
+
+    #[optimize(size)]
+    pub fn aeabi_uwrite8(x: u64, p: *mut u8) {
+        unsafe { p.cast::<u64>().write_unaligned(x) }
+    }
```
tgross35, line 432: These should be `unsafe`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1317/23f834808a3faa381c4859c6017793e7ff75898d..472f1218296beff265e1acc6ed421a62f15c42bd)*

compiler-builtins/src/arm.rs
thejpster: Has this been tested on ARMv4T and ARMv5TE (our only two Thumb-but-not-Thumb2 targets, as far as I am aware)?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1317/451f8f4bb295b2f1f374973582b4ddf48e62b1a4..f8fdb2588720159ad67e9adcb3aa6c0cf5b021de)*

builtins-test-intrinsics/src/main.rs
```diff
@@ -500,6 +526,16 @@ fn run() {
...
+    bb(aeabi_uread4(bb([0u8; 8].as_ptr().wrapping_add(1))));
+    bb(aeabi_uread8(bb([0u8; 16].as_ptr().wrapping_add(1))));
+    bb(aeabi_uwrite4(
+        bb(2),
+        bb([0u8; 8].as_mut_ptr().wrapping_add(1)),
+    ));
+    bb(aeabi_uwrite8(
+        bb(2),
+        bb([0u8; 16].as_mut_ptr().wrapping_add(1)),
+    ));
```
tgross35: Are these meant to be unaligned reads? You should make them `[0u32; ...]`/ `[0u64; ...]` and do `byte_add(1)`, because I think the arrays could be placed at any alignment

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1317/23f834808a3faa381c4859c6017793e7ff75898d..472f1218296beff265e1acc6ed421a62f15c42bd)*

builtins-test/tests/mem.rs
```diff
@@ -315,3 +315,64 @@ fn wide_string_length() {
...
+mod aeabi_unaligned {
+    use compiler_builtins::arm::__aeabi_uread4::__aeabi_uread4;
+    use compiler_builtins::arm::__aeabi_uread8::__aeabi_uread8;
+    use compiler_builtins::arm::__aeabi_uwrite4::__aeabi_uwrite4;
+    use compiler_builtins::arm::__aeabi_uwrite8::__aeabi_uwrite8;
+
+    /// A buffer with distinct bytes so that any mixup in byte order or offset is detected.
+    fn pattern() -> [u8; 16] {
+        core::array::from_fn(|i| (i as u8).wrapping_mul(0x37).wrapping_add(0x11))
+    }
```
tgross35: May as well just inline

```suggestion
    /// A buffer with distinct bytes so that any mixup in byte order or offset is detected.
    const PATTERN: [u8; 16] = [
        0x11, 0x48, 0x7f, 0xb6, 0xed, 0x24, 0x5b, 0x92, 0xc9, 0x0, 0x37, 0x6e, 0xa5, 0xdc, 0x13,
        0x4a,
    ];
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1317/23f834808a3faa381c4859c6017793e7ff75898d..472f1218296beff265e1acc6ed421a62f15c42bd)*

builtins-test/tests/mem.rs
```diff
@@ -315,3 +315,64 @@ fn wide_string_length() {
...
+
+    #[test]
+    fn uread() {
+        let buf = pattern();
+        for offset in 0..8 {
+            let b = &buf[offset..];
+            let expected4 = u32::from_ne_bytes([b[0], b[1], b[2], b[3]]);
+            let expected8 = u64::from_ne_bytes([b[0], b[1], b[2], b[3], b[4], b[5], b[6], b[7]]);
```
tgross35: Use `.first_chunk::<4>().unwrap()`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1317/23f834808a3faa381c4859c6017793e7ff75898d..472f1218296beff265e1acc6ed421a62f15c42bd)*

tgross35:
References:

* https://github.com/llvm/llvm-project/blob/f9a8dec9496fea7c55f591a66a64dbcdc085fc2d/compiler-rt/lib/builtins/arm/aeabi_uread4.S
* https://github.com/llvm/llvm-project/blob/f9a8dec9496fea7c55f591a66a64dbcdc085fc2d/compiler-rt/lib/builtins/arm/aeabi_uread8.S
* https://github.com/llvm/llvm-project/blob/f9a8dec9496fea7c55f591a66a64dbcdc085fc2d/compiler-rt/lib/builtins/arm/aeabi_uwrite4.S
* https://github.com/llvm/llvm-project/blob/f9a8dec9496fea7c55f591a66a64dbcdc085fc2d/compiler-rt/lib/builtins/arm/aeabi_uwrite8.S

compiler-builtins/src/arm.rs
```diff
@@ -277,4 +277,269 @@ intrinsics! {
...
+    ///
+    /// `address` must be valid for reading four bytes.
+    #[unsafe(naked)]
+    #[cfg(not(any(target_vendor = "apple", target_env = "msvc")))]
+    pub unsafe extern "aapcs" fn __aeabi_uread4(address: *const u8) -> u32 {
+        #[cfg(all(
+            target_feature = "thumb-mode",
+            not(target_feature = "thumb2"),
+            target_endian = "little"
+        ))]
```
tgross35: Note that we have a cfg for thumb that isn't present in rustc, https://github.com/rust-lang/compiler-builtins/blob/23f834808a3faa381c4859c6017793e7ff75898d/libm/configure.rs#L112-L131

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1317/23f834808a3faa381c4859c6017793e7ff75898d..472f1218296beff265e1acc6ed421a62f15c42bd)*

tgross35: Also can you `cfg_select!`? To make sure all cases are covered

compiler-builtins/src/arm.rs
Dirbaio: the `intrinsics!` macro breaks if you do that.

tgross35: I think you can put `#[...]` attributes on top of doc comments and it works because of how the macro parses things. But I wouldn't worry about it either way

builtins-test-intrinsics/src/main.rs
Dirbaio: I don't think it matters? the purpose of this test is just to check compiler-builtins does provide the symbols, calling them is enough for that even if the address is not unaligned. The other test does check functionality for all possible alignments.

tgross35: That's fair, but what's the purpose of the `wrapping_add` then?

compiler-builtins/src/arm.rs
Dirbaio: `thumb_1` should also be set for thumbv4t and thumbv5te. should I fix that?

builtins-test/tests/mem.rs
Dirbaio: done

builtins-test/tests/mem.rs
Dirbaio: done

builtins-test-intrinsics/src/main.rs
Dirbaio: yep it's useless. Removed!

builtins-test-intrinsics/src/main.rs
Dirbaio: done

compiler-builtins/src/arm.rs
Dirbaio: simplified even more by using the `thumb_1` cfg

compiler-builtins/src/arm.rs
Dirbaio: it also fails for the thumbv6k test since it's named "thumbv6" and not "thumbv6m"

compiler-builtins/src/arm.rs
thejpster: Apparently `thumbv6m-none-eabi` and `thumbv8m.base-none-eabi` count as Thumb-but-not-Thumb2 as well.

compiler-builtins/src/arm.rs
tgross35: Yes please!

If you have any ideas for how to improve this, it would be welcome. Maybe setting if a hardcoded list of targets match or if `cfg.has_target_feature("thumb-mode") && !cfg.has_target_feature("thumb2")`?

(We can't just use target_feature because `arm_target_feature` is unstable. Doesn't matter for c-b but does for libm)

Dirbaio: Done!

- fixed `thumb_1` to include all thumb-but-not-thumb2 target names. Shouldn't break anything else because there was literally no uses.
- Changed cfgs to use `thumb_1`
- Changed to use `cfg_select!`

compiler-builtins/src/arm.rs
Dirbaio: I've tested by compiling a thumbv6m binary and running it on a MCU. 

CI runs tests under QEMU for ARM mode only, no Thumb1 nor Thumb2. We could add them:

- Thumb2 is easy, there's `thumbv7neon-unknown-linux-gnueabihf`
- Thumb1 needs ugly hacks, there's no thumb1 Linux target. Something like `arm-unknown-linux-gnueabi` with `-Ctarget-feature=+thumb-mode --cfg thumb_1`

I've run those locally and they pass. I don't think I should add these in this PR, since it needs backporting it's best to keep diff minimal?

compiler-builtins/src/arm.rs
Dirbaio: is that rea''y necessary? it's just the standard ARM ABI where u32/ptrs occupy 1 reg and u64 occupies 2

compiler-builtins/src/arm.rs
tgross35: Feel free to add the CI here, just put it in a separate commit. Or a separate PR - makes no difference since I rebase merge.

tgross35: ^ the neon target that we can just run, that is. The other one should be separate.

I think ideally we’d have some kind of test runner that executes no-std targets in CI, that would be rather helpful. I’m not very familiar with this space but luckily you’re embedded experts :)

Dirbaio: done! I'm not too happy about the thumb1 hack though, if you have a better idea i'm all ears.

Dirbaio: ah just read your 2nd message. I've pushed to test only Thumb2. The interesting one to test is Thumb1 though which uses separate asm. Thumb2 and ARM mode share the same ASM code.

compiler-builtins/src/arm.rs
```diff
@@ -277,4 +277,261 @@ intrinsics! {
         // SAFETY: memclr preconditions apply, less strict alignment.
         unsafe { __aeabi_memset4(dst, n, 0) };
     }
+
+    // Read a `u32` from a possibly unaligned address.
```
tgross35, line 285: Could you add a link to https://github.com/ARM-software/abi-aa/blob/main/rtabi32/rtabi32.rst#533unaligned-memory-access above this section?

tgross35 (commented):
I renamed `thumb_1` to `thumb1_only` since (IMO) the name was a bit misleading 4907d5dc74d0d2f64f249384108989cbb42fd87e, and added a commented-out assertion that the target features line up with the config we're setting 81d14035f1659202feb650af86ebfbf2a733faf7. Sorry for the conflicts, could you rebase on top of that and uncomment that assertion?

Other than that, one comment request and LGTM, unless @thejpster has anything. I verified the instructions match up with LLVM's version.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1317/23f834808a3faa381c4859c6017793e7ff75898d..edfbe56525afffa32a721bf48c843a2f4740da49)*

compiler-builtins/src/arm.rs
Dirbaio: done!

thejpster (approved):

compiler-builtins/src/arm.rs
thejpster: I can live without it

tgross35 (approved):
Looks great, thanks!

Updated the commit messages to add tags and links (this repo uses rebase merging)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1317/81d14035f1659202feb650af86ebfbf2a733faf7..78e29597c5c11d2ca4fafd87fed594746392630f)*

PR #1320, opened 2026-09-16
ci: build armv7-gnueabihf with canonical runner

ubiratansoares:
Canonical has made more runners available for us, in particular these `armv7-linux-gnueabihf` ones. 

Creating this PR to give them a try and validate the integration. Ignore this PR for now.


tgross35:
I'm excited for this one!

.github/workflows/main.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
         - target: arm-unknown-linux-gnueabihf
           os: ubuntu-26.04
         - target: armv7-unknown-linux-gnueabihf
-          os: ubuntu-26.04
+          os: self-hosted-linux-armhf-noble-large-rust
```
tgross35: I know this is a WIP but we should probably keep the existing job too

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1320/9e648612293141da499883115a8300d3258939e2..0f23bdd5bd2b1e16e96af1ed29c09b914e61df14)*

ubiratansoares: Good point, I'll add it back :P

tgross35:
What makes these armv7 runners? https://github.com/rust-lang/compiler-builtins/actions/runs/35132215572/job/104915778763?pr=1320#step:3:21 and https://github.com/rust-lang/compiler-builtins/actions/runs/35132215572/job/104915778763?pr=1320#step:6:168 looks like it's an aarch64 host

ubiratansoares:
> What makes these armv7 runners?

I was told it and I took it for granted, but let me report this back to Canonical just in case 

screenshot-2026-09-16 at 20 20 35

.github/workflows/main.yaml
ubiratansoares: Done [here](https://github.com/rust-lang/compiler-builtins/pull/1320/commits/f2f4107f93a003e0c52f2d87279e73229f690e26)

ubiratansoares:
So, according to what I got from Canonical

> the runner itself is a genuine 32-bit armhf build. the ci log ( uname -a  =  aarch64 ) is just the host kernel reporting its own native architecture as these VMs run an arm64 kernel with AArch32/EL0 support enabled, which lets 32-bit armhf binaries execute natively

In this case, I think we can keep the arm64 host toolchain, cross-link with the armhf gcc, and run the test binaries natively in AArch32 mode, ie with no qemu. That would require adding a few additional setup steps via `ci/install-test-deps.sh` if I understand the ci harness correctly.

@tgross35  does that make sense to you?

tgross35:
Makes sense, I’m just surprised that it seems like arm32 isn’t reported anywhere and that rustup decided on an aarch64 host.

ubiratansoares:
@tgross35 I added the cross-compiling shenanigans, and also squashed out some test commits. Let me know if this looks good to you!

PR #1321, opened 2026-09-17
Rewrite AArch64 outlined atomic functions with asm

pmur:
These functions should follow aapcs, so a strictly asm version shouldn't be needed. This resolves some of the ugliness of reading HAVE_LSE_ATOMICS and enabling bti on various targets.

When compiled with opt-level=1 or higher, these produce nearly identical asm on linux with minor changes to register choices loading HAVE_LSE_ATOMICS.

~~Some minor macro changes are made to ensure existing lse tests continue to build and run.~~

~~TODO: collate and post asm differences. They should be minor.~~

~~TODO: figure out why tests now need `--features unmangled-names` to build now.~~

~~Questions: Do the test suites also run with opt-level=1? This should work at 0 (it did for my testing).~~


compiler-builtins/src/aarch64_outline_atomics.rs
```diff
@@ -243,29 +172,40 @@ macro_rules! compare_and_swap {
...
-                    // UXT    s(tmp0), s(0)
-                    concat!(uxt!($bytes), " ", reg!($bytes, 16), ", ", reg!($bytes, 0)),
-                    "0:",
-                    // LDXR   s(0), [x2]
-                    concat!(ldxr!($ordering, $bytes), " ", reg!($bytes, 0), ", [x2]"),
-                    // cmp    s(0), s(tmp0)
-                    concat!("cmp ", reg!($bytes, 0), ", ", reg!($bytes, 16)),
-                    "bne    1f",
-                    // STXR   w(tmp1), s(1), [x2]
-                    concat!(stxr!($ordering, $bytes), " w17, ", reg!($bytes, 1), ", [x2]"),
```
tgross35, line 263: The comments like these appear lost but would be good to preserve, the macro soup is kind of messy. (Wish we had templated `concat!`.)

compiler-builtins/src/aarch64_outline_atomics.rs
```diff
@@ -243,29 +172,40 @@ macro_rules! compare_and_swap {
...
-                    "cbnz   w17, 0b",
-                    "1:",
-                    "ret",
-                    have_lse = sym crate::aarch64_outline_atomics::HAVE_LSE_ATOMICS,
+                let mut expected = expected;
```
tgross35, line 178: The param can just be made `mut expected: int_ty!($bytes)` (applies a few places)

compiler-builtins/src/aarch64_outline_atomics.rs
```diff
@@ -308,24 +271,35 @@ macro_rules! swap {
...
+                            concat!("mov ", reg!($bytes, 16), ", ", reg!($bytes, 0)),
+                            "1:",
+                            concat!(ldxr!($ordering, $bytes), " ", reg!($bytes, 0), ", [x1]"),
+                            concat!(stxr!($ordering, $bytes), " w17, ", reg!($bytes, 16), ", [x1]"),
+                            "cbnz w17, 1b",
+                            inlateout("x0") left,
+                            in("x1") right_ptr,
+                            options(nostack),
+                            out("x16") _,
+                            out("w17") _,
```
tgross35: Style nit: put `options` at the end

compiler-builtins/src/aarch64_outline_atomics.rs
```diff
@@ -276,28 +216,51 @@ macro_rules! compare_and_swap_u128 {
...
-                    // STXP   w(tmp2), x2, x3, [x4]
-                    concat!(stxp!($ordering), " w15, x2, x3, [x4]"),
-                    "cbnz   w15, 0b",
-                    "1:",
-                    "ret",
-                    have_lse = sym crate::aarch64_outline_atomics::HAVE_LSE_ATOMICS,
+                let mut expected_lo = (expected & 0xFFFFFFFFFFFFFFFF) as u64;
+                let mut expected_hi = (expected >> 64) as u64;
+                let desired_lo = (desired & 0xFFFFFFFFFFFFFFFF) as u64;
+                let desired_hi = (desired >> 64) as u64;
```
tgross35: Nit: `as u64` is sufficient for the lo bits without masking first, bit easier to see that the masks aren't significant

tgross35 (commented):
> When compiled with opt-level=1 or higher, these produce nearly identical asm on linux with minor changes to register choices loading HAVE_LSE_ATOMICS.
> 
> TODO: figure out why tests now need `--features unmangled-names` to build now.

`naked` functions are special and always get a public module https://github.com/pmur/compiler-builtins/blob/6ad91ea6dbae4f2d0ce36989e0a2874c43d8f025/compiler-builtins/src/macros.rs#L378-L401. I think you could fix tests by removing one `::$name` from this line https://github.com/rust-lang/compiler-builtins/blob/9e648612293141da499883115a8300d3258939e2/builtins-test/tests/lse.rs#L135 (please rewrap it if you do, don't know how I missed that overflow).

> Questions: Do the test suites also run with opt-level=1? This should work at 0 (it did for my testing).

We run with a few options, unoptimized and opt-level=3 should both be covered. Any reason for opt-level=1 specifically? https://github.com/rust-lang/compiler-builtins/blob/9e648612293141da499883115a8300d3258939e2/ci/run.sh#L67-L73

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/compiler-builtins/1321/9e648612293141da499883115a8300d3258939e2..6ad91ea6dbae4f2d0ce36989e0a2874c43d8f025)*

pmur:
> We run with a few options, unoptimized and opt-level=3 should both be covered. Any reason for opt-level=1 specifically?

No, only that it is where the codegen becomes comparable to the previous implementation. Thanks for the incredible fast code review :open_mouth: .

compiler-builtins/src/aarch64_outline_atomics.rs
pmur: That currently runs afoul of `intrinsics!`. I am not sure how (or if) I can teach it how to accept a mut arg.

pmur:
Attaching a [diff](https://github.com/user-attachments/files/32386835/diff.txt) of the naked asm vs native asm implementation with optimization enabled.

It's ~1300 lines like:
```diff
-   0:  90000010        adrp    x16, 0 <compiler_builtins::aarch64_outline_atomics::HAVE_LSE_ATOMICS>
-   4:  39400210        ldrb    w16, [x16]
-   8:  34000070        cbz     w16, 14 <__aarch64_cas16_acq_rel+0x14>
+   0:  90000008        adrp    x8, 0 <compiler_builtins::aarch64_outline_atomics::HAVE_LSE_ATOMICS>
+   4:  39400108        ldrb    w8, [x8]
+   8:  34000068        cbz     w8, 14 <__aarch64_cas16_acq_rel+0x14>
    c:  4860fc82        caspal  x0, x1, x2, x3, [x4]
   10:  d65f03c0        ret
   14:  aa0003f0        mov     x16, x0
``` 

Notes about methodology:
```
# Extract functions into separate dirs and files
aarch64-linux-gnu-objdump -dC rlib > rlib.txt
gawk 'BEGIN { ON=0 } match($0, /<(__aarch64.*)>:/, name) { ON = 1; NAME=name[1] ".txt"; print NAME } /^$/ { ON = 0 } ON == 1 {print $0 >> NAME }' rlib.txt
# Diff old and new
for i in *.txt; do diff -ub $i ../old/$i; done > ../diff.txt 
```

Edit: Fix diff old/new.

PR #1324, opened 2026-09-22
Fix `i_to_f` test

beetrees:
When casting from `fN` to `iX`, `fN::{,NEG_}INFINITY` always saturates to to `[iu]X::{MAX,MIN}`, which can sometimes be further away from the original integer that was casted to `fN` than the next closest value to zero, `fN::{MAX,MIN}`. This PR fixes the test by manually calculating the integer value that is 1 ULP above/below `fN::{MAX,MIN}`, therefore allowing the test to pass in this case.

cc rust-lang/compiler-builtins#729

r? @tgross35


beetrees:
Test failures appear unrelated

